Issue #72, opened 2015-02-16
Poor contrast in HTML file between success/failure in overview


erasche:
I'm red-green colour blind, and I cannot tell any difference between the success and failure messages in the left hand overview. `:P`

It would be useful to do some combination of the following:

- increase colour intensity
- visually distinguish pass/fail (e.g. bold for failures)
- textually indicate pass/fail for those with poorer vision than me or using screen readers (e.g. "Pass" or "Fail" in the `(Test #1)` sections)

Galaxy is pretty horrifically disability unfriendly, but since planemo produces easily modified HTML reports, it'd be nice if they were a bit more accommodating.

PR #74, opened 2015-02-16
Fixes for #72

erasche:
IMO this is much easier to read and should address my colour blindness/vision impairment complaints. 

image


jmchilton:
Looks great, thanks!

PR #75, opened 2015-02-16
Attempt to parse json returned for toolshed errors

### LINKED ISSUE #67: Strange json output for pushing to an unchanged repository
```
esr@cpt:~/Projects/galaxy/tools-iuc/tools/macs2⟫ planemo shed_upload  --shed_target testtoolshed
{
    "content_alert": "The file \"/var/opt/galaxy/toolshed_data/test_toolshed/database/community_files/000/repo_749/dir2html.py\" contains HTML content.\n", 
    "err_msg": "No changes to repository."
}
```

It's bright red and the `content_alert`...

erasche:
Instead of erroring with some ugly json, print just the more useful error message from upstream.

I'm not fantastically happy about all the exit codes from this function being -1, it's pretty unhelpful. That said no matter what they are if it isn't a perfectly successful exit from this function, recursive uploads will reduce this to an error exit no matter what, so not really sure what to do about that.

This fixes #67 (writing again because apparently the first one doesn't xref properly)

### DIFF
--- a/planemo/commands/cmd_shed_upload.py
+++ b/planemo/commands/cmd_shed_upload.py
@@ -8,6 +8,7 @@ from planemo import shed
 from planemo.io import error
 from planemo.io import info
 from planemo.io import shell
+import json
 
 import fnmatch
 import os
@@ -117,7 +118,12 @@ def __handle_upload(ctx, path, **kwds):
     try:
         tsi.repositories.update_repository(repo_id, tar_path, **update_kwds)
     except Exception as e:
-        error("Could not update %s" % path)
-        error(e.read())
+        exception_content = e.read()
+        try:
+            upstream_error = json.loads(exception_content)
+            error(upstream_error['err_msg'])
+        except:
+            error("Could not update %s" % path)
+            error(exception_content)
         return -1
     info("Repository %s updated successfully." % path)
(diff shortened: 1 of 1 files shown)


erasche:
Will reduce complexity

erasche:
Okay, this is a fairly ugly way of reducing complexity which I really don't agree with, but it will now pass flake8. Please feel free to make recommendations.

planemo/commands/cmd_shed_upload.py
```diff
@@ -102,8 +103,33 @@ def __handle_upload(ctx, path, **kwds):
...
+        return -1
+
+    try:
+        tsi.repositories.update_repository(repo_id, tar_path, **update_kwds)
+    except Exception as e:
+        exception_content = e.read()
+        try:
+            upstream_error = json.loads(exception_content)
+            error(upstream_error['err_msg'])
+        except:
```
jmchilton: Bare `except`s are an anti-pattern - https://docs.python.org/2/howto/doanddont.html#except - can we modify this to be `except Exception`. I would be happy to make the change post-merge.

erasche: Indeed they are, sorry about that. Updated.

erasche: I've switched to reading the exception, hope that meets your requirements. Not a huge improvement, but it (was) done elsewhere in that file previously, so it seemed like it'd be ok? I haven't had time to debug exactly what was causing the failure to read the error message in the second bare exception that was patched, for future debugging, it happened in `package_vienna_*` in the tools_iuc repo.

jmchilton:
Awesome - this looks good to me. Thanks!

PR #98, opened 2015-03-08
planemo tool_init should wrap $input_file in double quotes

peterjc:
I was reading the writing a tool docs for v0.6.0 https://planemo.readthedocs.org/en/latest/writing.html and noticed the command string template did not double quote the filenames:

```
    <command><![CDATA[
        seqtk seq -a $input1 > $output1
    ]]></command>
```

This *untested* change attempts to do that, and ought to give:

```
    <command><![CDATA[
        seqtk seq -a "$input1" > "$output1"
    ]]></command>
```


jmchilton:
Awesome, thanks for the fix.

PR #101, opened 2015-03-17
Add support for automatic repo creation

erasche:
Should this be on by default? That's my only real question. If planemo can't find the target repository, it'll automatically create one.

Depends on bioblend > https://github.com/galaxyproject/bioblend/pull/124

PR #102, opened 2015-03-17
Toolshed Repo creation command

erasche:
Depends on bioblend > https://github.com/galaxyproject/bioblend/pull/124

xref #101


jmchilton:
Crap - I merged the commits with a real ``merge`` so the PR would close when I pushed but then I rebased things out of vanity and the merge information got lost. Sorry about that - the commits are nonetheless in there - I hope that is okay.

erasche:
No worries! Not a problem.

PR #103, opened 2015-03-19
Rename shed_repo_create to just shed_create.

jmchilton:
Repo breaks pattern or other shed commands - shed_diff is a repo diff but is not called shed_repo_diff. Same applies to shed_upload and shed_download.

Ping @erasche  - any objections to me making this modification to your command?


martenson:
:+1:

erasche:
:+1:

PR #113, opened 2015-04-06
Add code of conduct for the project.

jmchilton:
Inspired (entirely stolen) from khmer (https://github.com/ged-lab/khmer/blob/master/CODE_OF_CONDUCT.rst). What I like about the khmer code of conduct that is missing from a lot of code of conduct documents (including the http://contributor-covenant.org/ template khmer's is based upon) is specific recourse actions are listed in the document itself.

### DIFF
--- a/CODE_OF_CONDUCT.rst
+++ b/CODE_OF_CONDUCT.rst
@@ -1,0 +1,34 @@
+Contributor Code of Conduct
+===========================
+
+As contributors and maintainers of this project, we pledge to respect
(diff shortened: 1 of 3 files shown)


Ry4an:
driveby: :+1: this sort of thing makes projects stronger.

davebx:
If I grew an infinite number of bodies, each with an infinite number of arms, I still wouldn't have enough :+1: to do this justice.

bgruening:
:+1:

Issue #125, opened 2015-04-14
Normalize Issues


erasche:
1. Removes XML declaration (That's the `<?xml version="1.0" standalone="yes" encoding="UTF-8" ?>` string, or at the minimum `<?xml version="1.0"?>` that appears at the top of most XML files.)
2. Removes CDATA from `<command/>` blocks even when there's a `>` character there.

jmchilton:
The real issue here is that it is advertised I think :) - the output is not something you should replace your tools with.

erasche:
Well, then time to remove it before more people open similar issues? (Or hide it from the command menu?)

jmchilton:
Is #126 sufficient?

erasche:
+1

Issue #110, opened 2015-03-27
normalize lacks in place modification & destroys line wrapping


mr-c:
Many projects have a requirement for line wrapping at 79/80 characters whenever possible.

Personally I find that it improves readability.

Would be great if `planemo normalize` didn't undo that.

bgruening:
:+1: I usually put the `help` text into a new line.

PR #126, opened 2015-04-14
Docfix: Update normalize help text to be scary.

jmchilton:
Reflect the fact that I don't think the output should actually be meant to be a human readable tool (nothing but problems - see #125, #110, galaxyproject/tools-devteam#75).

@bgruening - you are the last hold out - do you mind if I merge this and release a 0.7.1 that warns people not to use ``normalize``?


davebx:
I :+1: this with trepidation, apprehension, unease, perturbation, anxiety, and fear.

PR #129, opened 2015-04-16
* Edited NO_TEST_DATA_MESSAGE to refer to --test_data, not --test-data

pvanheus:
0.3.0 (2015-02-13) changed --test-data to --test_data, but the NO_TEST_DATA_MESSAGE was referring to --test-data.

The code still looks for a default directory called test-data though. Perhaps this should be changed to test_data or it should look for both test-data and test_data?

### DIFF
--- a/planemo/galaxy_config.py
+++ b/planemo/galaxy_config.py
@@ -16,7 +16,7 @@ from planemo.io import shell
 
 NO_TEST_DATA_MESSAGE = (
     "planemo couldn't find a target test-data directory, you should likely "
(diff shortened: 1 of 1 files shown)


jmchilton:
Thanks for the fix, Galaxy and the tool shed call the directory ``test-data`` but all the planemo command-line parameters use ``_``. I think this is okay since one thing is a command line and one is a file and those can have differing conventions, though if I could do it over I might have gone with ``-`` for Planemo's command-line parameters.

PR #130, opened 2015-04-17
Use XSD to lint tools and repositories.

jmchilton:
- Implement validation abstraction that will use either lxml (Python lib) or xmllint (command line app) dependending on what is available - with test cases.
 - Add a ``--xsd`` flag to the ``lint`` command (for tools) that lints against the experimental XSD from https://github.com/JeanFred/Galaxy-XSD.
 - Implement ``repository_dependencies.xsd`` to describe Tool Shed ``repository_dependencies.xml`` files (fairly complete).
 - Implement ``tool_dependencies.xsd`` to describe Tool Shed ``tool_dependencies.xml`` files.
   - Validates attributes and elements down to the ``action`` elements and then largely gives up (sticking ``any`` and ``anyAttribute`` tags on that element).
   - Registers everything in tools-devteam as valid, and detects one invalid XML file in tools-iuc.
 - Implement new ``shed_lint`` command that:
   - Validates discovered ``tool_dependencies.xml`` against schema
   - Validates discovered ``repository_dependencies.xml`` against schema
   - Bare minimum to lint ``.shed.yml`` files.
   - Optionally also lints tools in repositories with the ``--tools`` argument. Like ``shed_upload`` uses presence of ``.shed.yml`` to detect repositories, this should be generalized.
   - Can recursively lint many repositories at one time with ``-r``.
 - Refactoring of existing stuff to support this and make room for XSD validation of tool XML files and generalizing applying actions over many repositories and many tools.


planemo/shed.py
```diff
@@ -192,6 +194,24 @@ def build_tarball(tool_path):
     return temp_path
 
 
+def walk_repositories(path):
```
erasche: oh, nice, will reuse this in recursive shed uploads.

martenson:
This is looking great. Very useful.

planemo/shed_lint.py
```diff
@@ -0,0 +1,87 @@
...
+    if not os.path.exists(shed_yaml):
+        lint_ctx.info("No .shed.yml file found, skipping.")
+        return
+    try:
+        shed_contents = yaml.load(open(shed_yaml, "r"))
+    except Exception as e:
+        lint_ctx.warn("Failed to parse .shed.yml file [%s]" % str(e))
+
+    warned = False
+    for required_key in ["owner", "name"]:
```
erasche: any reason for not using the expanded key set? You might add `description`, as that's the only other thing that's strictly required for a create-repo to succeed.

planemo/xml/xsd/tool/citations.xsd
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
erasche: not using submodules?

erasche: I know how much fun <s>copy and pasting</s> scripted import of files from other projects is ;)

planemo/shed.py
jmchilton: That is where I got it, it has already be refactored out of ``shed_upload``.

erasche:
+1 from me, can't wait to add this to the jenkins testing bot! :D

planemo/shed.py
erasche: Oh, excellent, thanks!

fre. 17. apr. 2015 kl. 09.51 skrev John Chilton <notifications@github.com>:

planemo/shed_lint.py
jmchilton: It is required for creation, but not for vanilla ``upload`` - even these two fields are not strictly required... hmm.... I am not against requiring everything in `.shed.yml` but this was a conservative first attempt.

planemo/xml/xsd/tool/citations.xsd
jmchilton: Hmm... maybe... https://github.com/pypa/pip/pull/577 ... yeah I should do that.

jmchilton: Done - thanks for the suggestion.

PR #133, opened 2015-04-22
Improved .shed.yml linting and type default.

jmchilton:
No .shed.yml fields are required anymore, but when present more are validated more carefully.

 - Validate ``owner`` string (if specified) matches validation enforced by Tool Shed.
 - Validate ``name`` string (if specified) matches validation enforced by Tool Shed.
 - Validate ``type`` (if specified) is valid.
 - Validate ``name`` and ``type`` conform to conventions (``package_.*`` for deps and ``suite_.*`` for suites).

Use new validation to more confidently select a default type based on name during creation - instead of always falling back on ``unrestricted`` use ``tool_dependency_definition`` for ``package_`` repositories and ``repository_suite_definition`` for ``suite_`` repositories.

### DIFF
--- a/planemo/shed.py
+++ b/planemo/shed.py
@@ -128,7 +128,7 @@ def create_repository(ctx, tsi, path, **kwds):
 
     description = repo_config.get("description", None)
     long_description = repo_config.get("long_description", None)
-    type = repo_config.get("type", "unrestricted")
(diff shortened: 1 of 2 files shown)


jmchilton:
Rebased to fix linting problems.

PR #140, opened 2015-04-24
More linting configurablity.

jmchilton:
Add ``--skip`` argument to ``lint`` and ``shed_lint`` to pass in modules to skip while linting - I think the common ones are going to be ``citations`` and ``xml_order`` - so that would be ``--skip 'citations,xml_order'`` for instance. (TODO: Probably worth still catching unknown top-level tags even if xml_order is disabled - so those should be split out I guess).

A default value for skip can be specified on a per-user basis by specifying a list of items in ``~/.planemo.yml`` under a ``lint_skip`` key.

This commit also improves the display of linting so the short name is what is display - e.g. user will now see "Applying linter citations" and not "Appling linter lint_citations".

### DIFF
--- a/planemo/commands/cmd_lint.py
+++ b/planemo/commands/cmd_lint.py
@@ -13,12 +13,13 @@ from planemo.tool_lint import lint_tools_on_path
 @options.optional_tools_arg()
 @options.report_level_option()
 @options.fail_level_option()
+@options.skip_option()
 @options.lint_xsd_option()
 @pass_context
 def cli(ctx, path, **kwds):
     """Check specified tool(s) for common errors and adherence to best
     practices.
     """
-    lint_args = build_lint_args(**kwds)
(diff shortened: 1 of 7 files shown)


erasche:
> A default value for skip can be specified on a per-user basis by specifying a list of items in ~/.planemo.yml under a lint_skip key.

you know, in all of those awesome changes, not one of them is rST and in the docs/ folder

:P, <3

jmchilton:
Yes, I will tidy things up when I work on ``HISTORY.rst`` for a 0.8.0 release.

erasche:
Okay, just giving ya a hard time. Have a good weekend

fre. 24. apr. 2015 kl. 15.06 skrev John Chilton <notifications@github.com>:

nsoranzo:
Thanks a lot!

PR #152, opened 2015-04-29
Warn on empty citations block.

jmchilton:
See https://github.com/galaxyproject/tools-iuc/commit/9107450a9a95db3dd7d99999d6cc2f745f759771#commitcomment-10962483.

I think there needs to be a way to tell planemo that it is fine that tool doesn't use citations - e.g.

```
<!-- NOLINT:CITATIONS - common unix utility doesn't need citation. -->
<!-- NOLINT:CITATIONS - I wrote it and I don't want people to know -->
```

so if people would instead to prefer to wait for that to be availabe before this gets merged that would be acceptable.

### DIFF
--- a/planemo_ext/galaxy/tools/linters/citations.py
+++ b/planemo_ext/galaxy/tools/linters/citations.py
@@ -24,4 +24,5 @@ def lint_citations(tool_xml, lint_ctx):
 
     if valid_citations > 0:
         lint_ctx.valid("Found %d likely valid citations.", valid_citations)
-
+    else:
+        lint_ctx.warn("No valid citation elements found in <citations> block.")


peterjc:
How about something explicit, like:

```
<citations>
    <citation type="none">I wrote this tool and I don't care about it being cited</citation>
</citations>
```

erasche:
+1 to @peterjc's.

jxtx:
I think the original behavior, an empty citation block, is a simpler explicit way to state that there are no citations.

erasche:
@jxtx hmmm. Maybe so, but @peterjc's option would let us make that information available, lest one day we reach critical mass for cited tools and people wonder why specific tools aren't cited.

@jmchilton If @jxtx's comment is true, then empty test block should be downgraded to an INFO, while missing test block should remain a WARN.

jmchilton:
Yeah - I am fine with @jxtx solution but then planemo should walk all the children and make sure each is a ``citation`` to ensure sure there are no ``<catition>`` or ``<citition>`` elements, etc...

peterjc:
I'm OK with @jxtx's proposal.

jmchilton:
@jmchilton that is already being done - you wrote the code you should have known that.

Alright - a lot of discussion to decide that the status quo is fine and @erasche can continue cheating :).

erasche:
Cheers ;)

PR #154, opened 2015-04-30
Fix yaml for tool sheds in .planemo.yml template

peterjc:


jmchilton:
Awesome- thanks for the fix.

Issue #159, opened 2015-04-30
File order in tar-ball from shed_upload


peterjc:
I know the order does not matter by the time the tar-ball is pushed to the Tool Shed, but for local checks it would be nice if the ``.shed.yml`` ordering given in the ``include`` directives was preserved.

e.g.

```
$ cat ~/repositories/pico_galaxy/tools/venn_list/.shed.yml
name: venn_list
owner: peterjc
homepage_url: https://github.com/peterjc/pico_galaxy/tools/venn_list
remote_repository_url: https://github.com/peterjc/pico_galaxy/tools/venn_list
description: Draw Venn Diagram (PDF) from lists, FASTA files, etc
long_description: |
  Draw Venn Diagrams for 1, 2 or 3 sets of identifiers as a PDF file.

  Can parse FASTA, FASTQ, SFF or tabular files (taking column one) for identifiers.
  This can be combined with the extensive tabular file filtering and manipulation
  tools within Galaxy to prepare the input data.

  Uses the R/Bioconductor package limma to draw the PDF file, called from Python with
  rpy.

  Uses Biopython to parse SFF files.
categories:
- Graphics
- Sequence Analysis
- Visualization
type: unrestricted
include:
- README.rst
- venn_list.py
- venn_list.xml
- tool_dependencies.xml
```

This gave the following:

```
$ planemo shed_upload --tar_only  ~/repositories/pico_galaxy/tools/venn_list 
cp /tmp/tmpWm9B9U shed_upload.tar.gz
$ tar -xvf /tmp/tmpWm9B9U
venn_list.py
tool_dependencies.xml
README.rst
venn_list.xml
```

peterjc:
Reading over ``planemo/shed.py`` I think the code constructs a temporary folder containing all the files and folders to be turned into the tar-ball? In this case, the order is at the mercy of the file-system and my request is non-trivial.

Plan B would be sorting by name - sorting the top level is easy but does not alter the order of files in sub-folders, which is down to how ``tarfile``'s recursive add works:

```
$ git diff
diff --git a/planemo/shed.py b/planemo/shed.py
index 6189176..8bd9a2e 100644
--- a/planemo/shed.py
+++ b/planemo/shed.py
@@ -401,7 +401,9 @@ def build_tarball(realized_path, **kwds):
     try:
         tar = tarfile.open(temp_path, "w:gz", dereference=True)
         try:
-            for name in os.listdir(realized_path):
+            # File system order essentially random, so at least sort
[...]

PR #164, opened 2015-04-30
Sort planemo shed_upload tar-ball by filename

peterjc:
This would resolve issue #159 by the simple expedient of building a list of all the files in the staging directory, and using the sorted list to build the tar-ball.

Note this does not explicitly handle directories, thus any empty directory would be missing from the tar-ball. Since the Tool Shed uses hg internally this does not matter (same for git too).

e.g. silly test case showing the desired sorting:

```
$ planemo shed_upload --tar_only  ~/repositories/pico_galaxy/tools/venn_list && tar -tzf shed_upload.tar.gz
cp /tmp/tmpYAjHCW shed_upload.tar.gz
aaa
test-data/magic.pdf
test-data/rhodopsin_proteins.fasta
test-data/venn_list.tabular
test-dataAAA
tools/venn_list/README.rst
tools/venn_list/tool_dependencies.xml
tools/venn_list/venn_list.py
tools/venn_list/venn_list.xml
toolsAAA
zzz
```

### DIFF
--- a/planemo/shed.py
+++ b/planemo/shed.py
@@ -397,18 +397,21 @@ def build_tarball(realized_path, **kwds):
     # It should be pushed up a level into the thing that is uploading tar
     # balls to iterate over them - but placing it here for now because
     # it address some bugs.
+
+    # Simplest solution to sorting the files is to use a list,
+    files = []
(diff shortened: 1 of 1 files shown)


jmchilton:
Awesome - thanks a bunch!

Issue #141, opened 2015-04-25
Improved shed_diff.


jmchilton:
shed_diff should...

- [x] Build actual tar balls and diff those (so ignored files like ``.shed.yml`` are not included).
- [x] Have a default enabled mode like that does an actual XML diff of TS artifacts and filters TS generated attribute changes - disable with ``--raw``.

After these changes ``shed_diff`` will be ready for realized repositories and will be able to predict if a ``shed_upload`` will be needed. After that, ``shed-upload`` should diff before uploading.

This would completely alleviate my lingering concerns on https://github.com/galaxyproject/tools-iuc/issues/50.

jmchilton:
Fixed with 40a1f57 and #167.

PR #167, opened 2015-05-01
Smart shed_diff.

jmchilton:
- Fix ``shed_diff`` so its exit code actually is non-zero if there are differences.
 - By default - don't report differences for tool shed populates attributes (``toolshed``, ``changeset_revision``). This can be disabled with ``--raw``.
 - Tests for ensure ``--raw`` and smart differencing are working and exit codes are fixed.
 - Update shed test webapp to inject the tool shed populated fields into respective arguments.

Implements #141.

### DIFF
--- a/planemo/commands/cmd_shed_diff.py
+++ b/planemo/commands/cmd_shed_diff.py
@@ -29,6 +29,12 @@ from planemo import shed
          " this to testtoolshed.",
     default=None,
 )
+@click.option(
+    "--raw",
+    is_flag=True,
+    help="Do not attempt smart diff of XML to filter out attributes "
+         "populated by the Tool Shed.",
+)
 @options.recursive_shed_option()
 @pass_context
 def cli(ctx, path, **kwds):
@@ -53,7 +59,7 @@ def cli(ctx, path, **kwds):
     def diff(realized_repository):
         working = tempfile.mkdtemp(prefix="tool_shed_diff_")
         try:
-            shed.diff_in(ctx, working, realized_repository, **kwds)
+            return shed.diff_in(ctx, working, realized_repository, **kwds)
         finally:
             shutil.rmtree(working)
 
--- a/planemo/shed/diff.py
+++ b/planemo/shed/diff.py
@@ -1,0 +1,51 @@
+from __future__ import print_function
+
+import os
+import sys
+from xml.etree import ElementTree
+
+from planemo.xml import diff
+
+
+def diff_and_remove(working, label_a, label_b, f):
+    a_deps = os.path.join(working, label_a, "tool_dependencies.xml")
+    b_deps = os.path.join(working, label_b, "tool_dependencies.xml")
+    a_repos = os.path.join(working, label_a, "repository_dependencies.xml")
+    b_repos = os.path.join(working, label_b, "repository_dependencies.xml")
+
+    deps_diff = 0
+    if os.path.exists(a_deps) and os.path.exists(b_deps):
+        deps_diff = _shed_diff(a_deps, b_deps, f)
+        os.remove(a_deps)
+        os.remove(b_deps)
+
+    repos_diff = 0
+    if os.path.exists(a_repos) and os.path.exists(b_repos):
+        repos_diff = _shed_diff(a_repos, b_repos, f)
+        os.remove(a_repos)
+        os.remove(b_repos)
+
+    return deps_diff and repos_diff
+
+
+def _shed_diff(file_a, file_b, f=sys.stdout):
+    xml_a = ElementTree.parse(file_a).getroot()
+    xml_b = ElementTree.parse(file_b).getroot()
+    _strip_shed_attributes(xml_a)
+    _strip_shed_attributes(xml_b)
+    return diff.diff(xml_a, xml_b, reporter=f.write)
+
+
+def _strip_shed_attributes(xml_element):
+    if xml_element.tag == "repository":
+        _remove_attribs(xml_element)
+    children = xml_element.getchildren()
+    if len(children) > 0:
+        for child in children:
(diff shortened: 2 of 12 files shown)


erasche:
> Fix shed_diff so its exit code actually is non-zero if there are differences.

+1

erasche:
> Update shed test webapp to inject the tool shed populated fields into respective arguments.

can you just `subprocess.check_command(['sh', 'run_shed.sh', '--daemon'])`?

planemo/commands/cmd_shed_diff.py
```diff
@@ -29,6 +29,12 @@
          " this to testtoolshed.",
     default=None,
 )
+@click.option(
+    "--raw",
```
erasche: unused?

planemo/commands/cmd_shed_diff.py
```diff
@@ -53,7 +59,7 @@ def cli(ctx, path, **kwds):
     def diff(realized_repository):
         working = tempfile.mkdtemp(prefix="tool_shed_diff_")
         try:
-            shed.diff_in(ctx, working, realized_repository, **kwds)
+            return shed.diff_in(ctx, working, realized_repository, **kwds)
```
erasche: only return in one place, returning (I'm assuming) an error code, or None. Maybe better to explicitly return None/-1?  Sorry, it's late, tipsy code review is probably not the best ;)

planemo/shed/diff.py
```diff
@@ -0,0 +1,51 @@
...
+
+    return deps_diff and repos_diff
+
+
+def _shed_diff(file_a, file_b, f=sys.stdout):
+    xml_a = ElementTree.parse(file_a).getroot()
+    xml_b = ElementTree.parse(file_b).getroot()
+    _strip_shed_attributes(xml_a)
+    _strip_shed_attributes(xml_b)
+    return diff.diff(xml_a, xml_b, reporter=f.write)
```
erasche: +1

planemo/xml/diff.py
```diff
@@ -0,0 +1,56 @@
...
+            reporter('x2 has an attribute x1 is missing: %s'
+                     % name)
+            return False
+    if not text_compare(x1.text, x2.text):
+        reporter('text: %r != %r' % (x1.text, x2.text))
+        return False
+    if not text_compare(x1.tail, x2.tail):
+        reporter('tail: %r != %r' % (x1.tail, x2.tail))
+        return False
+    return _compare_children(x1, x2, reporter)
```
erasche: yep, used this too. Good choice.

tests/repository_dependencies_shed.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
+<repositories description="A suite of Galaxy tools designed to work with version 1.2 of the SAMtools package.">
+    <repository changeset_revision="cf875cbe2df4" name="data_manager_sam_fasta_index_builder" owner="devteam" toolshed="https://toolshed.g2.bx.psu.edu" />
```
erasche: maybe test some without hardcoded `changeset_revision`? Not strictly necessary, just being picky since I'm a bottle and a half in :P

erasche:
+1

jmchilton:
> can you just subprocess.check_command(['sh', 'run_shed.sh', '--daemon'])?

That would really increase the runtime of those tests - it takes well less than a second to start the server and run those tests. I also like mocking out the tool shed interface because it really crystalizes exactly the API planemo is consuming.

There is one functional test that actually exercises a real tool shed (https://github.com/galaxyproject/planemo/blob/master/tests/test_shed.py) and certainly it would be cool if that would automatically start a tool shed and stuff but I have not gotten around to it yet. It is a little bit harder than just ``run_tests.sh`` - since it should configure an isolated environment for the test.

planemo/commands/cmd_shed_diff.py
jmchilton: It is the last statement in the diff - it is wrapped in a finally not exception handling so this ``return`` is the only place to return from.

planemo/commands/cmd_shed_diff.py
jmchilton: It is used inside of ``planemo.shed``.

tests/repository_dependencies_shed.xml
jmchilton: Well I was a bottle and half in when I wrote this so seems appropriate :). This is meant to emulate the shed version - the plain version is in tests/repository_dependencies.xml - I use this to test that ``shed_diff`` filters the required attributes to make sure these files are actually different.

Thanks for the detailed review!

planemo/commands/cmd_shed_diff.py
erasche: Whoops. Missed that. Sorry!

erasche:
That all sounds fair. Definitely harkens back to your point about "reason for travis and a reason for buildbot". 

If you'd like long running tests, the IUC jenkins bot is underutilised! :)

jmchilton:
Thanks for the offer - it isn't so much about CI though. I definitely want that tool shed test running on Travis - another minute is not my concern. I think my other comment was about Galaxy and that minute would drawfs in comparision to the hours it would take to run Galaxy's functional tests. What I love is ``PLANEMO_SKIP_GALAXY_TESTS=1 nosetests`` is taking literally a couple seconds on my personal laptop. A hour matters in CI, a minute doesn't IMO - but a minute does matter during development. I love refactoring quickly and with impunity...

erasche:
Gotcha, sounds good. Didn't realise that shed tests ran that quickly. Made a offhand assumption that it'd be 15-30m to really test all the shed related functionality and corner cases. (Especially if testing against shed@master and shed@dev)

PR #171, opened 2015-05-01
Update setup.py after planemo/shed.py --> planemo/shed/__init__.py

peterjc:
Fixes side effect of a15b3bf9a5707acdfd9cdf2da1401b700aef2710


jmchilton:
Doh - thanks for the fix!

Issue #170, opened 2015-05-01
Collect and Use Git Revision


jmchilton:
Use commit ``git rev-parse HEAD`` + ``if `git diff --quiet` then '-dirty' else ''`` to capture git information at ``shed_upload``. At first use this information in the commit message to HG, but longer term we should collect the data on the tool shed in a more semantically useful way.

See also #169.

peterjc:
Are any of the Galaxy community using ``hg`` for tool development? If so, adding something similar for mercurial hashes would be nice.

Also, thinking out loud, I'd like a safety feature to stop me from trying to upload uncommitted changes to the ToolShed. i.e. rather than adding "dirty" message (which is useful), should we prompt the user for confirmation? Or abort unless a ``--force`` flag is used? (following the ``git`` command line conventions, maybe not the best idea).

jmchilton:
I would happily merge a PR to add ``hg`` support - but I don't intend to spend time on it personally. I think tool development is moving to Github and it would seem like wasted effort. I think we want to *actively* encourage Github specifically and short of that git more generally.

I like the spirit of the second idea a lot. For convenience and backward compatibility - my preference would be to invert the default (disallow ``-dirty`` commits with a flag - like ``--no_dirty`` - though that particular flag seems poorly named) and document how to modify the default with ``~/.planemo.yml``.

PR #174, opened 2015-05-01
Allow the shed_ operations to consume git URLs instead of paths.

### LINKED ISSUE #169: Shed Operations on Github URLs
Update shed realization process to take in git URLs in addiiton to file paths. This will allow each of ``shed_create``, ``shed_upload``, ``shed_diff``, ``shed_download``, and ``shed_lint`` to work on repositories sitting in Github.

jmchilton:
create, upload, download, diff, and lint projects directly from Github - closes #169.

Inputs starting with ``git+`` (e.g. git+/path/to/repo.git) with will have the ``git+`` stripped and passed directly to the ``git`` command-line. Alternatively URLs starting with `git:` will be passed as is to the ``git`` command-line. This seems to roughly correspond to ``pip``.

### DIFF
--- a/planemo/galaxy_config.py
+++ b/planemo/galaxy_config.py
@@ -15,6 +15,7 @@ import click
 from planemo import galaxy_run
 from planemo.io import warn
 from planemo.io import shell
+from planemo import git
 
 NO_TEST_DATA_MESSAGE = (
     "planemo couldn't find a target test-data directory, you should likely "
@@ -331,7 +332,7 @@ def _install_galaxy_via_git(ctx, config_directory, kwds):
     _ensure_galaxy_repository_available(ctx)
     workspace = ctx.workspace
     gx_repo = os.path.join(workspace, "gx_repo")
-    command = "git clone %s galaxy-dev" % (gx_repo)
+    command = git.command_clone(ctx, gx_repo, "galaxy-dev")
     _install_with_command(config_directory, command)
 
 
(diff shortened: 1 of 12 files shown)


planemo/galaxy_config.py
```diff
@@ -354,7 +355,8 @@ def _ensure_galaxy_repository_available(ctx):
         shell("git --git-dir %s fetch >/dev/null 2>&1" % gx_repo)
     else:
         remote_repo = "https://github.com/galaxyproject/galaxy"
-        shell("git clone --bare %s %s" % (remote_repo, gx_repo))
+        command = git.command_clone(ctx, remote_repo, gx_repo, bare=True)
+        shell(command)
```
bgruening: Assuming we are only targeting github/bitbucket it would be faster to pull the latest tarball. With all the test data over time a git pull can become a bottle-neck.

planemo/io.py
```diff
@@ -58,3 +61,12 @@ def untar_to(url, path, tar_args):
         shell("%s | %s" % (download_cmd, untar_cmd))
     else:
         shell("%s > '%s'" % (download_cmd, path))
+
+
+@contextlib.contextmanager
+def temp_directory():
```
bgruening: Oh, this is beautiful!

planemo/galaxy_config.py
jmchilton: So this line is pulling Galaxy down for ``test``/``serve`` when used with ``--install_galaxy`` - which used to use a tar ball by default  but now does a clone so that it can do incremental updates on each call very quickly.

It sounds like you mean for the actual shed stuff though - which could be optimized somewhat. I don't think we should just download the tars though because I would like to use the git metadata during upload (#170) and because this will need to support submodules at some point. I do think we could do ``clone`` with ``--depth=1`` probably to speed things up though paired with creating a directory to cache repositories between calls (mirroring how Galaxy is quickly cloned).

bgruening: Fair enough, thanks for explaining.

PR #175, opened 2015-05-02
Fix `clone --bare` mode

bgruening:


jmchilton:
Ugh - sorry about that - thanks for the fix!

Issue #180, opened 2015-05-05
Nesting include settings in .shed.yml (source as list)?


peterjc:
Consider a moderately complex ``include`` setting like this which is quite repetitive:

```
include:
- README.rst
- venn_list.py
- venn_list.xml
- tool_dependencies.xml
- source: ../../test-data/magic.pdf
  strip_components: 2
- source: ../../test-data/venn_list.tabular
  strip_components: 2
- source: ../../test-data/rhodopsin_proteins.fasta
  strip_components: 2
```

It seems that currently ``include`` is a flat list of elements (defaulting to ``source`` entries if not a mini-dict), but allowing each ``source`` entry to be a list would allow this kind of bundling:

```
include:
- README.rst
- venn_list.py
- venn_list.xml
- tool_dependencies.xml
- strip_components: 2
  source:
  - ../../test-data/magic.pdf
  - ../../test-data/venn_list.tabular
  - ../../test-data/rhodopsin_proteins.fasta
```

e.g. Something like this hack:

```
$ git diff
diff --git a/planemo/shed/__init__.py b/planemo/shed/__init__.py
index 8bacfc8..e108bc6 100644
--- a/planemo/shed/__init__.py
+++ b/planemo/shed/__init__.py
@@ -882,7 +882,10 @@ class RealizedFile(object):
     def realized_files_for(path, include_info):
         if not isinstance(include_info, dict):
             include_info = {"source": include_info}
-        source = include_info.get("source")
+        source_list = include_info.get("source")
+        if not isinstance(source_list, list):
+            # A single filename or pattern
+            source_list = [source_list]
         destination = include_info.get("destination", None)
         strip_components = include_info.get("strip_components", 0)
         if destination is None:
@@ -890,14 +893,16 @@ class RealizedFile(object):
             destination_specified = False
         else:
             destination_specified = True
-        abs_source = os.path.join(path, source)
-        dest_is_file = destination_specified and os.path.isfile(abs_source)
         realized_files = []
-        for globbed_file in _glob(path, source):
-            src = os.path.relpath(globbed_file, path)
-            realized_files.append(
-                RealizedFile(path, src, destination, dest_is_file, strip_components)
-            )
+        for source in source_list:
+            abs_source = os.path.join(path, source)
+            dest_is_file = destination_specified and os.path.isfile(abs_source)
+            for globbed_file in _glob(path, source):
+                src = os.path.relpath(globbed_file, path)
+                realized_files.append(
[...]

PR #186, opened 2015-05-07
Avoid ambigous destination value in test case...

peterjc:
Suggestion: If the ``source`` has any wildcards (or is a folder with an implicit wildcard), then ``destination`` must be a directory given with trailing slash (otherwise error).

The include source of "../shared_files/extra_test_data/**" would only find "../shared_files/extra_test_data/extra_test_file.txt" which after strip_components 3 gives "extra_test_file.txt"

Question: did the ".shed.yml" destination of "test-data" mean call this file "test-data" [sic] or use "test-data/extra_test_file.txt" instead? Why?

What if the include source glob matches multiple files? Then might expect destination must be a folder - since alternative means trying to map multiple input files to one in the tar-ball/

Using destination "test-data/" seem unambiguous and preferable.

(I found this issue during re-factoring while working on #180)

### DIFF
--- a/planemo/shed/__init__.py
+++ b/planemo/shed/__init__.py
@@ -902,6 +902,7 @@ class RealizedFile(object):
         if not isinstance(include_info, dict):
             include_info = {"source": include_info}
         source = include_info.get("source")
+        abs_source = os.path.join(path, source)
         destination = include_info.get("destination", None)
         strip_components = include_info.get("strip_components", 0)
         if destination is None:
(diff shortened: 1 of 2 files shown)


peterjc:
P.S. I am willing to try and add code to catch the proposed error condition.

jmchilton:
Okay - I can live with that - what you are proposing seems like a tighter, more reasonable spec. We should make sure shed_lint provides a good description of the problem though.

peterjc:
So changes to both ``shed_lint`` (new ground for me) and ``shed_upload`` (starting to understad) needed?

jmchilton:
If you return an exception - like unmatched globs - I think ``shed_lint`` will just print a message for it - so we should verify this but I don't think it will require changes to ``shed_lint``.

jmchilton:
Awesome - thanks for working on this.

PR #187, opened 2015-05-07
Handle empty tar-balls gracefully on old Python

peterjc:
This is a workaround for http://bugs.python.org/issue6123 where older versions of Python fail to handle an empty tar-ball.

(I found this during re-factoring where an error elsewhere in the shed_upload logic was resulting in empty tar-balls being generated)


peterjc:
(Updated to use Python 3 compatible exception syntax)

jmchilton:
Awesome - thanks for the fix!

PR #190, opened 2015-05-08
Warn on tool id contains a space.

jmchilton:


peterjc:
+1 pending even stricter guidelines

PR #195, opened 2015-05-11
Relax path checking for tool_dependencies.xml etc

peterjc:
Solved problem I had on a branch with ``'./tool_dependencies.xml'`` being ignored for a tool package because it only allowed the exact match ``'tool_dependencies.xml'`` (likewise for tool suites).

On the other hand, with this change these special files don't have to be at the top level of the Tool Shed tar-ball any more. Is a sub-folder acceptable?

### DIFF
--- a/planemo/shed/__init__.py
+++ b/planemo/shed/__init__.py
@@ -837,15 +837,15 @@ class RawRepositoryDirectory(object):
     def _implicit_ignores(self, relative_path):
         # Filter out "unwanted files" :) like READMEs for special
         # repository types.
+        name = os.path.basename(relative_path)
         if self.type == REPO_TYPE_TOOL_DEP:
-            if relative_path != TOOL_DEPENDENCIES_CONFIG_NAME:
+            if name != TOOL_DEPENDENCIES_CONFIG_NAME:
                 return True
 
         if self.type == REPO_TYPE_SUITE:
-            if relative_path != REPO_DEPENDENCIES_CONFIG_NAME:
+            if name != REPO_DEPENDENCIES_CONFIG_NAME:
                 return True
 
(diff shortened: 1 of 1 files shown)


jmchilton:
I would prefer planemo enforce the special files (e.g. ``tool_dependencies.xml``) are at the root of the repository.

Further down in the file, should the following method:

```
    @property
    def relative_dest(self):
        if self.dest_is_file:
            destination = self.dest
        else:
            destination = os.path.join(self.dest, self.stripped_source)
        return os.path.relpath(destination)
```

actually be something like:

```
    @property
    def relative_dest(self):
        if self.dest_is_file:
            destination = self.dest
        else:
            destination = os.path.join(self.dest, self.stripped_source)
        return os.path.normpath(os.path.relpath(destination))
```

It seems like that ``normpath`` call would eliminate the superfluous ``./`` while still enforcing that ``tool_dependencies.xml`` files are in the root of the tar ball. What do you think?

peterjc:
I'm playing with a re-factoring which gets rid of that method -wWork in progress which will get rebased / force-updated: https://github.com/peterjc/planemo/tree/rel_paths - but yes, using ``os.path.normpath(...)`` sounds good.

Note that ``tool_dependencies.xml`` does not have to be at the base path, e.g. https://github.com/peterjc/galaxy_blast/tree/master/tools/blast_rbh has ``tools/blast_rhb/tool_dependencies.xml``, but I concede for the special cases like tool packages there is no need for any sub-folders.

peterjc:
On reflection, it seems reasonable to insist ``tool_dependencies.xml`` is at the tar-ball root for a package repository (likewise for suites). Withdrawing pull request.

PR #200, opened 2015-05-13
Add -m as alias for --message in planemo shed_upload

peterjc:
This matches the ``git commit`` argument for a commit message, and thus should be natural and familiar to planemo users too.

(I found myself trying to use ``planemo shed_upload ... -m "v1.2.3 adding cool stuff"`` and it didn't work)


jmchilton:
Awesome sauce!

Issue #201, opened 2015-05-13
planemo shed_diff -r breaks if sub-directory not recognised


peterjc:
This folder is a very early work in progress, with no ``.shed.yml`` and no ``*.xml`` file either. As expected, ``planemo shed_diff`` fails (although not cleanly):

```
$ ls -1 ~/repositories/pico_galaxy/tools/seq_concatenate/
seq_concatenate.py
$ planemo shed_diff --shed_target testtoolshed  ~/repositories/pico_galaxy/tools/seq_concatenate/
Unable to find repository id, cannot download.
Traceback (most recent call last):
...
Exception: Unable to find repository id, cannot download.
```

The problem is that this also happens in recursive mode. It should report this, but keep going:

```
$ planemo shed_diff --shed_target testtoolshed -r  ~/repositories/pico_galaxy/tools/
...
<real differences snipped>
...
Unable to find repository id, cannot download.
Traceback (most recent call last):
...
Exception: Unable to find repository id, cannot download.
```

peterjc:
I think this needs to be treated as a special error (repository not in Tool Shed), causing ``shed_diff`` to fail unless in recursive mode (report but continue).

Initial hack:

```
$ git diff
diff --git a/planemo/shed/__init__.py b/planemo/shed/__init__.py
index ede7440..45ccca9 100644
--- a/planemo/shed/__init__.py
+++ b/planemo/shed/__init__.py
@@ -226,6 +226,13 @@ def _diff_in(ctx, working, realized_repository, **kwds):
     other = os.path.join(working, label_b)
 
     tsi = tool_shed_client(ctx, read_only=True, **kwds)
+    # In order to download the tarball, require repository ID...
+    repo_id = realized_repository.find_repository_id(ctx, tsi)
+    if repo_id is None:
+        error("%s does not seem to be in the tool shed, can't do shed_diff"
+              % realized_repository.name)
+        # TODO - Should this return an error code which can be checked for in recursive mode?
+        return 0
     download_tarball(
         ctx,
         tsi,
```

PR #208, opened 2015-05-14
recursive shed_diff with directories not yet in Tool Shed

peterjc:
Previously this error condition would mean a recursive ``planemo shed_diff`` would fail on repository not yet uploaded to the Tool Shed. It now prints an error message but continues.

Resolves GitHub issue #201

TODO: Ensure that shed_diff on a single repository gives an error return code, e.g.

```
$ planemo shed_diff ~/repositories/pico_galaxy/tools/seq_concatenate/ && echo "$? returned"
Repository [seq_concatenate] does not exist in the targeted Tool Shed, can't do shed_diff
0 returned
```

### DIFF
--- a/planemo/shed/__init__.py
+++ b/planemo/shed/__init__.py
@@ -226,6 +226,13 @@ def _diff_in(ctx, working, realized_repository, **kwds):
     other = os.path.join(working, label_b)
 
     tsi = tool_shed_client(ctx, read_only=True, **kwds)
+    # In order to download the tarball, require repository ID...
+    repo_id = realized_repository.find_repository_id(ctx, tsi)
+    if repo_id is None:
+        error("Repository [%s] does not exist in the targeted Tool Shed, can't do shed_diff"
+              % realized_repository.name)
+        # TODO - Should this return an error code which can be checked for in recursive mode?
+        return 0
     download_tarball(
         ctx,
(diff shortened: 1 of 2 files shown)


planemo/shed/__init__.py
```diff
@@ -226,6 +226,13 @@ def _diff_in(ctx, working, realized_repository, **kwds):
     other = os.path.join(working, label_b)
 
     tsi = tool_shed_client(ctx, read_only=True, **kwds)
+    # In order to download the tarball, require repository ID...
+    repo_id = realized_repository.find_repository_id(ctx, tsi)
+    if repo_id is None:
+        error("Repository [%s] does not exist in the targeted Tool Shed, can't do shed_diff"
+              % realized_repository.name)
+        # TODO - Should this return an error code which can be checked for in recursive mode?
+        return 0
```
jmchilton: I feel like if you just return a non-zero value here the exit code would be handled correctly by the layers above this.

peterjc: This is the TODO, we need a clean way to signal to the recursive shed_diff that this error should be ignored (i.e. a warning to the terminal only), whereas in non-recursive mode it really is an error which should quit with a non-zero return code.

martenson:
thank you @peterjc

PR #210, opened 2015-05-15
Implement a ``--failed`` flag for the ``test`` command.

jmchilton:
This will use ``tool_test_output.json`` to determine which tests failed the last time ``planemo test`` was executed and only run those previously failing tests. This should speed up test-driven development of tools.

Requires galaxyproject/galaxy#266.

Closes #145.

Issue #177, opened 2015-05-04
Implement Github/GitURL -> Shed persistent mapping backend.


jmchilton:
See https://github.com/galaxyproject/planemo/issues/169#issuecomment-98158780 for more context - but the idea would be to provide the basis for a RESTful interface and web app for managing synchronizing Github repositories to the Tool Shed.

Depending on implementation and usefulness on the command-line, this may well belong in another project to uses Planemo as a library - for instance there no reason to Planemo to have a database requirement.

See also:
- #169 
- #170

Issue #19, opened 2014-11-12
Allow testing tool_dependencies.xml without a tool shed.


jmchilton:
- [ ] Finish work on transcribing tool sheds to Homebrew taps (https://github.com/jmchilton/shed2tap).
 - [ ] Every hour synchronize the tool shed and test tool shed with a brew tap (e.g. https://github.com/jmchilton/homebrew-toolshed). 
 - [ ] Add planemo options for installing a `tool_dependencies.xml` with as if it were installed into a particular tool shed - all with `brew` using these taps.
 - [ ] Implement a Galaxy dependency resolver that uses `tool_dependencies.xml` plus these custom taps to resolve dependencies the way it would be done via tool shed installs.
 - [ ] Add option to `test` command to do resolution in this fashion.

erasche:
From #86 

Possibly should try and find the packages locally (e.g. ../../packages/###)

mdshw5:
Is there any progress on this feature? I'm currently trying to test a tool with tool_dependencies using `planemo serve` but realize that I need to set up a local tool shed and then.. (inform planemo about it?)

jmchilton:
Not really - sorry. If you want to test tools with ``planemo serve`` you need to make sure the dependencies are locally available. It doesn't address tool shed dependencies at this time - to do that testing I guess I would recommend publishing to the test tool shed and testing manually.

mdshw5:
Got it. Thanks!
Matt Shirley

jmchilton:
Closing as won't fix - we are moving toward conda and these options exist today. The remaining issues are largely all Galaxy issues.

PR #213, opened 2015-05-18
Implement shed serve and test commands, other enhancements.

### LINKED ISSUE #150: Many more commands should allow specifing multiple directories.
See #139. At least anything with a ``-r`` option including ``shed_lint``.

### LINKED ISSUE #176: Implement shed_test command.
- Don't test local stuff - test stuff published to shed.
  - Setup an on-the-fly Galaxy instance with ``planemo.galaxy_config`` module.
  - Update to actually configure shed artifacts.
  - Use ``bioblend`` to install the repository.
  - Run the tests (I think ``run_tests.sh --installed`` might be do this).
- Allow specifying a ``shed_target``.
- Use ``.shed.yml`` to resolve id, etc....

jmchilton:
* Implement ``shed_test`` command that installs the latest installable revision of specified repositories and tests them. Closes #176.  a48d811 
* Implement ``shed_serve`` command that likewise installs the latest installable revision of specified repositories  into a fresh Galaxy instance and serves it. 74e3d64
* Enhance a variety of commands to allow specifying multiple paths (``test``, ``serve``, ``shed_*``). Closes #150. 2bd5dfc
* Refactorings to enable this and reduce existing duplication. daa4f59, 4116999

Requires merging of:
- [x] #210 
- [x] galaxyproject/galaxy#267


erasche:
@jmchilton does this get us closer to https://github.com/galaxyproject/planemo/issues/19 ?

jmchilton:
Nope... this is almost exactly the opposite of that. Planemo was originally about fighting the machine and has totally devolved into just another a cog :rage:.

I have been making slow progress on #19 (for instance https://github.com/galaxyproject/planemo/tree/shed2tap2). I tried to work on this again this weekend and I am having real problems synchronizing how Python is handled in the Tool Shed with Homebrew.

erasche:
Ahah. Gotcha. From the description of "installs the latest installable revision of specified repositories into a fresh Galaxy instance and serves it" it sounded a lot like we'd be getting close to testing tools with dependencies.

jmchilton:
This does test tools with dependencies - but once they have been uploaded into the tool shed not before. This can be thought of a stand-alone variant of Galaxy's install and test framework.

erasche:
So, if I'm reading this right, I can use `shed_test` for testing packages, and then `shed_serve` to upload them all to a Galaxy, and somehow then use a `test` to use that same Galaxy with the installed packages?

jmchilton:
Not quite - a potential workflow would be something like this:

 - Use ``test`` and ``lint`` to develop tools locally.
 - Use ``serve`` to review tool. Ready to publish...
 - Use ``shed_init`` to create ``.shed.yml``.
 - Write ``tool_dependencies.xml``.
 - Use ``shed_upload --force_repository_creation --shed_target testtoolshed`` to populate the tool on the test tool shed.
 - Use ``shed_test --target testtoolshed`` to test published tools. This will start a Galaxy in the background, install the tool(s) from the test tool shed and its dependencies, and run the tests with tool shed dependency resolution (instead of local resolution configured with ``test``).
 - Publish to main tool shed.

I don't think I would use ``shed_serve`` in a typical workflow - but if you want to do ``serve`` but with tool shed installed tools instead of local copies - that is what it would do. (More for consistency and as a stepping stone to ```shed_test``.)

The goal with #19 is more to have a ``test`` that works like ``shed_test`` but without having to upload the packages to the tool shed and install and test them. If you had a really robust variant of #19 you can imagine you could eliminate publishing to the test tool shed and running ``shed_test`` from this workflow - you would be confident the thing you upload passes the tests before publishing them.

erasche:
As always I'm thinking from the perspective of CI servers... :)

So I could do:

- launch a TS in a docker container (@bgruening, we can do this yeah?)
- Use `shed_upload --force_repository_creation --shed_target $docker_ts` to populate the tool on the local tool shed.
- Use `shed_test --target $docker_ts` to test published tools. This will start a Galaxy in the background, install the tool(s) from the test tool shed and its dependencies, and run the tests with tool shed dependency resolution (instead of local resolution configured with test).

jmchilton:
Well again - if we could do the mythical #19 you wouldn't need the tool shed - planemo+galaxy would be enough.

This stop gap is so I have something useful to show at the GCC but doesn't help with your exact scenario. It does help with other potential CI jobs - for instance you can imagine automatically pushing everything committed to github to the test tool shed and if the tests pass there with ``shed_test`` then promoting them to the main tool shed. Or giving the repository a green light for publication in a UI built on something like https://github.com/galaxyproject/planemo/issues/177.

erasche:
Sure, not needing a TS would be absolutely brilliant, but I'm willing to use whatever hacks/stopgaps I can in the meantime.

>  for instance you can imagine automatically pushing everything committed to github to the test tool shed and if the tests pass there with shed_test then promoting them to the main tool shed

definitely going to do that for https://github.com/galaxyproject/tools-iuc/issues/129

I'll play around with the new commands soon and see what I can do, I really, really, really want Jenkins/Planemo tests to start being green across the board rather than "oh, this test is failing because the bedtools from brew is a different patch version than the one in the TS"

bgruening:
@erasche yes we have a TS in a docker container if you are interested.
@jmchilton  probably know my plan to strip out the relevant parts of parsing & installing tool_dependencies into it's own library. But this `hack` is great and serves us well over the next month I guess!

Thanks John!
Bjoern

erasche:
@bgruening wunderbar. I'll test this out wednesday and see if I can get jenkins to handle complete testing now. Ahhh so excited.

PR #216, opened 2015-05-20
Implement ``shed_update`` to upload & update repo metadata.

### LINKED ISSUE #199: planemo shed_upload fails if only trying to change metadata
e.g.

```
$ planemo shed_upload --shed_target testtoolshed ~/repositories/pico_galaxy/tools/mira4_assembler/ ; echo "$? returned"
cd '/mnt/galaxy/repositories/pico_galaxy/tools/mira4_assembler' && git rev-parse HEAD
cd '/mnt/galaxy/repositories/pico_galaxy/tools/mira4_assembler' && git diff --quiet
/mnt/galaxy/repositories/planemo/.venv/lib/python2.6/site-packages/requests-2.7.0-py2.6.egg/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
/mnt/galaxy/repositories/planemo/.venv/lib/python2.6/site-packages/requests-2.7.0-py2.6.egg/requests/packages/urllib3/connectionpool.py:768: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
Could not update mira4_assembler
No changes to repository.
255 returned
```

There was a change to the meta-data in ``.shed.yml`` which I was hoping to push to the Test Tool Shed via ``planemo shed_upload`` even though the files themselves had not changed: https://github.com/peterjc/pico_galaxy/commit/238db3900d3a01e82089aaaae4a1919556805504

jmchilton:
This should replace ``shed_upload`` in many potential workflows - but this does require a ``.shed.yml`` file with all metadata populated so ``shed_upload`` is still useful for directly uploading custom tar files for instance. I have filled out the documentation on ``shed_upload`` and ``shed_update`` to reflect this.

Additionally, for consistency with the new ``shed_update`` operation - ``shed_create`` now also performs an upload in addition to simply populating the metadata for a new remote repository. The old behavior of ``shed_create`` can be obtained by supplying a new ``--skip_upload`` flag. ``shed_create`` can also take in a ``-m/--message`` option now as well to reflect this new behavior.

Updated ``publish.rst`` to reflect these simplifications (instructions now easier).

Closes #199.

### DIFF
--- a/planemo/commands/cmd_shed_update.py
+++ b/planemo/commands/cmd_shed_update.py
@@ -38,12 +38,13 @@ def cli(ctx, paths, **kwds):
 
         % planemo shed_update --shed_target testtoolshed path/to/repo
 
-    Another important option is ``--check_diff`` - this will check for
-    differences before uploading new contents to the tool shed. This may
-    important because the tool shed will automatically populate certain
-    attributes in tool shed artifact files (such as ``tool_dependencies.xml``)
-    and this may cause unwanted installable revisions to be created when
-    there are no important changes.
+    Another important option is ``--check_diff`` - this doesn't affect the
+    updating of shed metadata but it will check for differences before
+    uploading new contents to the tool shed. This may important because the
+    tool shed will automatically populate certain attributes in tool shed
+    artifact files (such as ``tool_dependencies.xml``) and this may
+    cause unwanted installable revisions to be created when there are no
+    important changes.
 
     The lower-level ``shed_upload`` command should be used instead if
     the repository doesn't define complete metadata in a ``.shed.yml``.
--- a/planemo/commands/cmd_shed_update.py
+++ b/planemo/commands/cmd_shed_update.py
@@ -1,0 +1,74 @@
+"""
+"""
+import sys
+
+import click
+
+from planemo.cli import pass_context
+from planemo import options
+from planemo import shed
+from planemo.io import info, error
+
+
+@click.command("shed_update")
+@options.shed_publish_options()
+@options.shed_upload_options()
+@options.shed_skip_upload()
+@options.shed_skip_metadata()
+@pass_context
+def cli(ctx, paths, **kwds):
+    """Update repository in shed from a ``.shed.yml`` file.
+
+    By default this command will update both repository metadata
+    from ``.shed.yml`` and upload new contents from the repository
+    directory.
+
+    ::
+
+        % planemo shed_update
+
+    This will update the main tool shed with the repository defined
+    by a ``.shed.yml`` file in the current working directory. Both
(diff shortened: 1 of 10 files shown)


erasche:
+1!

erasche:
@jmchilton what's the timeline for the next planemo release?

planemo/commands/cmd_shed_upload.py
```diff
@@ -36,23 +32,24 @@
...
-    help="If a repository cannot be found for the specified user/repo name "
-         "pair, then automatically create the repository in the toolshed.",
-    is_flag=True,
-    default=False
-)
-@click.option(
-    "--check_diff",
-    is_flag=True,
-    help="Skip uploading if the shed_diff detects there would be no "
-         "'difference' (only attributes populated by the shed would would "
```
peterjc: Typo: double would

planemo/commands/cmd_shed_update.py
```diff
@@ -0,0 +1,74 @@
...
+    ::
+
+        % planemo shed_update --shed_target testtoolshed path/to/repo
+
+    Another important option is ``--check_diff`` - this will check for
+    differences before uploading new contents to the tool shed. This may
+    important because the tool shed will automatically populate certain
+    attributes in tool shed artifact files (such as ``tool_dependencies.xml``)
+    and this may cause unwanted installable revisions to be created when
+    there are no important changes.
```
peterjc: This could be a little clearer: Does the diff happen after the meta-data update, but before the tar-ball upload?

e.g. Given a previously up to date tool (in sync with the tool shed), I fix a typo in the ``.shed.yml`` description, and run ``planemo shed_update ...``. What happens (with or without ``--check-diff``)?

bgruening:
Wonderful!

planemo/commands/cmd_shed_update.py
jmchilton: ``--check_diff`` would only affect the upload - so the metadata would be updated and the tar ball would not be uploaded. I will try to clarify this in the text.  There is a problem with exit codes also though - I think in this case the exit code would be non-zero but probably shouldn't be. That is a smaller bug though - so I will just create an issue for this.

jmchilton:
@erasche new planemo release should happen today or tomorrow.

PR #224, opened 2015-05-27
Allow specifying per-user keys in planemo.yml

davebx:


erasche:
@jmchilton I like what @davebx is proposing, but just a thought... might it not be better to re-organise the data structure to something like:

```yaml
my_mts_account:
  user: username
  pass: password
  key: key
  shed_url: http://toolshed.g2.bx.psu.edu
iuc_mts_account:
  user: username
  pass: password
  key: key
  shed_url: http://toolshed.g2.bx.psu.edu
iuc_tts_account:
  user: username
  pass: password
  key: key
  shed_url: http://testtoolshed.g2.bx.psu.edu
```

and then in `shed_url`, instead of taking in toolshed URLs, just take in the `shed_identifier`, i.e. the top level keys?

jmchilton:
@erasche I would prefer what you outlined I think - it seems to be a concise way to also get custom tool sheds and fits in a backward compatible way with the existing format better I think.

@davebx There are still some Travis problems - let me know if you get a chance to resolve them. If you want me to take a look and fix the issues and make the changes requested by @erasche  - let me know as well and I can take a crack at it.

martenson:
@davebx do you plan to follow up on the travis tests or is this PR obsolete?

jmchilton:
I think this is mostly replaced by #314 which has now been released so I am going to close out this PR, hope that is okay @davebx.

PR #235, opened 2015-06-19
Make specifying --galaxy_install optional.

jmchilton:
If Galaxy cannot be found - just download it dynamically even if --galaxy_install is not specified.

Simplifies one slide of my GCC presentation :).


bgruening:
I like the argumentation and the enhanced user experience. :+1:

Issue #238, opened 2015-06-26
Kitematic (Need Help)


jmchilton:
The planemo machine docs should be updated with kitamatic instructions.

 - [x] Maybe screenshots of getting to booting the planemo image - https://registry.hub.docker.com/u/planemo/box/.
 - [x] Question: The Dockerfile ends with /bin/bash at the end of ``CMD`` so running ``docker run`` yields a shell after supervisor starts up. Does kitematic provide an interface for this or do kitmatic users need to SSH in or use the web ide?

I cannot get kitematic or boot2docker running on my Windows dev server - so help with this would be most welcome.

erasche:
Answer to question (c/p from IRC), you can't get a shell like you do in normal docker from the CLI, however you can click "Exec" and launch a terminal. Almost as good.

erasche:
For kitematic, it'd really work better to switch the second part of `CMD` to `tail -f /var/log...` just for the visual response for users.

erasche:
Trimming images + uploading shortly.

erasche:
![0-launch-kitematic](https://cloud.githubusercontent.com/assets/458683/8384505/050eb474-1c07-11e5-8fce-010453d3a3da.PNG)
![1-booting-kitematic](https://cloud.githubusercontent.com/assets/458683/8384506/05114a2c-1c07-11e5-9ec8-e41859bdfc76.PNG)
![2-finding-planemo](https://cloud.githubusercontent.com/assets/458683/8384509/0521caf0-1c07-11e5-8887-8ae8975c46da.PNG)
![4-downloading planemo](https://cloud.githubusercontent.com/assets/458683/8384511/0522e796-1c07-11e5-9082-2ee82c10255a.png)
![5-starting-planemo](https://cloud.githubusercontent.com/assets/458683/8384510/0522a024-1c07-11e5-9832-1505de4539f8.PNG)

You may want red circles around "Exec" and the link-out button above the galaxy screenshot
![6-planemo-running](https://cloud.githubusercontent.com/assets/458683/8384512/05252948-1c07-11e5-9f15-c3c0aba5ba99.PNG)
![7-powershell-root](https://cloud.githubusercontent.com/assets/458683/8384508/0520f0e4-1c07-11e5-9e85-ce46af1ea796.PNG)
![8-powershell-user](https://cloud.githubusercontent.com/assets/458683/8384507/051dbece-1c07-11e5-8407-83fa5bb11087.PNG)

[...]

PR #242, opened 2015-06-29
Updated docs with newer kitematic images of planemo/server

erasche:
Updates for #238


jmchilton:
Beautiful, beautiful - thanks a bunch!

PR #276, opened 2015-07-31
Add --from_work_dir option in init_tool

lecorguille:
The first part of my first contribution.

I just add an option --from_work_dir because we often succeed  to avoid from_workr_dir argument.

Without --from_work_dir
```
    <command><![CDATA[
        samtools sort -o "$input1" > "$output1"
    ]]></command>
    <outputs>
        <data name="output1" format="bam" from_work_dir="1_sorted.bam" />
    </outputs>
```

With --from_work_dir
```
    <command><![CDATA[
        samtools sort -o "$input1" > "$output1"
    ]]></command>
    <outputs>
        <data name="output1" format="bam" />
    </outputs>
```

I hope I follow enough the good practice to not waste your time. Planemo, git and GitHub were almost new for me 2 days ago :)

Let me know if I was on the wrong way.

Gildas

### DIFF
--- a/planemo/commands/cmd_tool_init.py
+++ b/planemo/commands/cmd_tool_init.py
@@ -84,6 +84,13 @@ REUSING_MACROS_MESSAGE = ("Macros file macros.xml already exists, assuming "
           "with a data input parameter)."),
 )
 @click.option(
+    "--from_work_dir",
+    is_flag=True,
+    default=None,
+    prompt=False,
+    help=("Hardcode the output name"),
+)
+@click.option(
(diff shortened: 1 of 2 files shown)


jmchilton:
Thanks for the contribution. That from_work_dir would be included and not used was a bug anyway. I think the following PR https://github.com/galaxyproject/planemo/pull/277 fixes the bug and probably eliminates the need for your new flag. What do you think - is that an acceptable alternative?

lecorguille:
More stylish ! Well done !
Thanks

PR #277, opened 2015-07-31
Bugfix: tool_init use from_work_dir only if file in command.

jmchilton:
Fixes bug in the resulting tools using the output directly but having dummy from_work_dir files that are largely ignored by Galaxy under normal operation (but causing a somewhat more undefined behavior with Pulsar for instance). The updated example tools clarify this I think.

See also #276 (ping @lecorguille).


lecorguille:
More stylish ! Well done !
Thanks

jmchilton:
Alright - thanks again for the contribution though. I had completely forgotten about this bug.

PR #278, opened 2015-08-05
Allow -t as shorthand for --shed_target and correct string literals

peterjc:
I seem to need ``--shed_target`` for all of the online ``planemo shed_*`` commands I use, therefore ``-t`` would seem a nice shorthand.

However, I appreciate we should be cautious about allocating one letter aliases for Planemo commands to avoid any inconsistencies/clashes (in particular, we should avoid doing this for things that can go in the user's configuration file).

### DIFF
--- a/planemo/options.py
+++ b/planemo/options.py
@@ -21,7 +21,7 @@ def test_data_option():
     return click.option(
         "--test_data",
         type=click.Path(exists=True, file_okay=False, resolve_path=True),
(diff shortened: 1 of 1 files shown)


martenson:
:

erasche:
:+1:

jmchilton:
:+1: - looks good, thanks @peterjc.

Issue #280, opened 2015-08-13
lint -r should return a non-zero exitcode when a xml file can't be read 


mvdbeek:
If there is a malformed xml, planemo lint -r will report an error, but exit with a zero exitcode.
(Without -r planemo properly reports a non-zero exitcode).
An example:

```
(.venv)lbd-dge-02:test marius$ planemo lint -r --fail_level error
Error loading tool with path /Users/marius/src/gedtools/tools/test/MirParser/MirParser.xml
Traceback (most recent call last):
  File "/Users/marius/.venv/lib/python2.7/site-packages/planemo_ext/galaxy/tools/loader_directory.py", line 35, in load_tool_elements_from_path
    tool_elements.append((file, loader.load_tool(file)))
ParseError: not well-formed (invalid token): line 6, column 8
Linting tool /Users/marius/src/gedtools/tools/test/mismatch_frequencies/mismatch_frequencies.xml
Applying linter top_level... CHECK
.. CHECK: Tool defines a version.
.. CHECK: Tool defines a name.
.. CHECK: Tool defines an id name.
Applying linter tests... CHECK
.. CHECK: 1 test(s) found.
Applying linter stdio... CHECK
.. INFO: No stdio definition found, tool will determine an error from stderr.
Applying linter output... CHECK
.. INFO: 3 outputs found.
Applying linter inputs... CHECK
.. INFO: Found 8 input parameters.
Applying linter help... CHECK
.. CHECK: Tool contains help section.
.. CHECK: Help contains valid reStructuredText.
Applying linter command... CHECK
.. INFO: Tool contains a command with interpreter of type [python].
(.venv)lbd-dge-02:test marius$ echo $?
0
```

PR #281, opened 2015-08-13
Return non-zero exitcode if planemo fails to load a xml file due to m…

mvdbeek:
…alformed xml tags.

See #280


jmchilton:
It is an absolute embarrassment I took so long to get to this @mvdbeek - I am really sorry. Thanks for the fix, I have made some modifications - do you want to take a look at #292 and see if it is okay.  The refinements are described in https://github.com/jmchilton/planemo/commit/84c70b3687002847a27847e1179df43bd7ea63fa.

mvdbeek:
Thanks @jmchilton , it looks very good!

PR #284, opened 2015-08-26
output readbility improvements

martenson:
print current repo being diffed
add space to the diff command


jmchilton:
Awesome sauce, thanks @martenson.

PR #288, opened 2015-09-02
Change 127.0.0.1 to 0.0.0.0 to allow serving in docker container.

chambm:
My docker container doesn't have a GUI. Nor for that matter does my Linux host. So I'm connecting from my Windows box, through a Linux VM, to the Docker container hosting 'planemo serve'. I had to make this change for it to work properly.


jmchilton:
I think I don't want to accept this as is. I worry about binding an unsecured Galaxy to 0.0.0.0 by default. I understand the need for it but I would rather make it configurable - by adding a ``--host`` parameter to anything to hosts Galaxy, and falling back to ``default_galaxy_host`` in ``~/.planemo.yml`` if that is not specified, and then falling back to ``127.0.0.1``.

Let me know if this sounds like an acceptable path forward and I can try to make those changes.

chambm:
Yep, that sounds better. I went for the expedient solution because I 
wasn't familiar with the code. :)

-Matt

jmchilton:
Superseded by #301 which includes this commit.

PR #297, opened 2015-09-11
Adding shed_build command to build tarballs.

kellrott:
This is an attempt to build shed tarballs without needing to log into the shed server or do an upload.

### DIFF
--- a/planemo/commands/cmd_shed_build.py
+++ b/planemo/commands/cmd_shed_build.py
@@ -1,0 +1,30 @@
+"""
+"""
+import sys
+
+import click
+import shutil
+
+from planemo.cli import pass_context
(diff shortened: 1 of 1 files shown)


jmchilton:
This looks pretty solid, thanks @kellrott. I'll add some tests.

peterjc:
Is this functionally like setting up a alias for ``planemo shed_upload --tar_only PACKAGE``?

kellrott:
`planemo shed_upload --tar_only ` still tries to log in

peterjc:
Understood - this sounds like a useful addition then (and might mean we can drop the ``--tar_only`` option from ``planemo shed_upload`` later on).

PR #307, opened 2015-09-21
Allow reading shed key from an environment variable.

jmchilton:


jmchilton:
Forgot about this, sorry @peterjc. Merged now.

Issue #303, opened 2015-09-18
Command to convert tool_dependencies.xml recipes into shell scripts


peterjc:
Related to #19 (testing ``tool_dependencies.xml`` without a tool shed), I would like to be able to run an install recipe from a ``tool_dependencies.xml`` file locally and/or turn it into a simple shell script for the current platform.

(The platform specific actions could be turned into bash if statements if preferred)

This seems to overlap with https://github.com/jmchilton/shed2tap

e.g. https://github.com/peterjc/pico_galaxy/blob/master/tools/effectiveT3/tool_dependencies.xml

```
<?xml version="1.0"?>
<tool_dependency>
    <package name="effectiveT3" version="1.0.1">
        <install version="1.0">
            <actions>
                <!-- Set environment variable so Python script knows where to look -->
                <action type="set_environment">
                    <environment_variable name="EFFECTIVET3" action="set_to">$INSTALL_DIR</environment_variable>
                </action>
                <!-- Main JAR file -->
                <action type="shell_command">wget http://effectors.csb.univie.ac.at/sites/eff/files/others/TTSS_GUI-1.0.1.jar</action>
                <!-- If using action type download_file will need to move the file,
                <action type="move_file"><source>TTSS_GUI-1.0.1.jar</source><destination>$INSTALL_DIR/</destination></action>
                -->
                <!-- Three model JAR files -->
                <action type="make_directory">$INSTALL_DIR/module</action>
                <action type="shell_command">wget http://effectors.csb.univie.ac.at/sites/eff/files/others/TTSS_ANIMAL-1.0.1.jar</action>
                <action type="move_file"><source>TTSS_ANIMAL-1.0.1.jar</source><destination>$INSTALL_DIR/module/</destination></action>        
                <action type="shell_command">wget http://effectors.csb.univie.ac.at/sites/eff/files/others/TTSS_PLANT-1.0.1.jar</action>
                <action type="move_file"><source>TTSS_PLANT-1.0.1.jar</source><destination>$INSTALL_DIR/module/</destination></action>
                <action type="shell_command">wget http://effectors.csb.univie.ac.at/sites/eff/files/others/TTSS_STD-1.0.1.jar</action>
                <action type="move_file"><source>TTSS_STD-1.0.1.jar</source><destination>$INSTALL_DIR/module/</destination></action>
                <action type="shell_command">wget http://effectors.csb.univie.ac.at/sites/eff/files/others/TTSS_STD-2.0.1.jar</action>
[...]

PR #310, opened 2015-09-28
Initial implementation of "planemo depbash"

peterjc:
As described on issue #303, this implements a new ``planemo depbash`` command for converting ``tool_dependencies.xml`` install recipes into bash scripts.

There is a lot that could be done better, or added, including:
- refactor to use a visitor pattern instead of my ``.to_bash()`` methods
- expand the command line API with options for paths and filenames
- complete the action coverage (especially the R/Python/Perl environments)
- avoid collisions in the download cache which currently assumes unique filenames

Is this suitable to merge as is, or does it need further work?

Does anyone have a better idea for the command name? I have no strong attachment to "depbash" (dependencies to bash script).

I have no objections to squashing down the git commit history here.

P.S. I have considered determining dependencies is out of scope for this work.


jmchilton:
Thanks for the excellent work @peterjc!

First order of bike shedding :disappointed:, none of the existing commands concatenate words together or shorten words - so I think ``dependency_script``, ``dependency_bash``, ``dependencies_script``, or ``dependencies_bash`` might be better command titles. Thoughts?

peterjc:
I was initially thinking something like ``shed_bash`` but this isn't really a Tool Shed action at all.

All four of your suggestions seem fine to me - ``*_bash`` is more precise, but potentially less immediately obvious than ``*_script``. And on singular vs plural, as it stands this can be used on one or many ``tool_dependencies.xml`` files but equally ``dependency_*`` works as an adjective.

On balance, I lean to ``dependency_script``.

jmchilton:
Agreed on every point - I think ``dependency_script`` is best.

planemo/shed2tap/base.py
```diff
@@ -301,24 +357,181 @@ def from_elem(cls, elem, package):
...
+    if target_filename:
+        downloaded_filename = target_filename
+    else:
+        downloaded_filename = os.path.split(url)[-1]
+        if "?" in downloaded_filename:
+            downloaded_filename = downloaded_filename[:downloaded_filename.index("?")]
+        if "#" in downloaded_filename:
+            downloaded_filename = downloaded_filename[:downloaded_filename.index("#")]
+
+    # Curl is present on Mac OS X, can we assume it will be on Linux?
```
bgruening: We can not assume that this is installed on a Galaxy system (https://wiki.galaxyproject.org/Admin/Config/ToolDependenciesList) for that reason we have a TS package for it, but I guess this is fine for a developer environment that is used for planemo.

peterjc: Good - setting the download target filename with ``wget`` is tricky.

If curl is missing this would fail noisily and clearly at least :)

bgruening:
Thanks Peter for working on this and sorry to be so late to this discussion. Have you considered to reuse the original TS code? My idea was to separate the Galaxy installation code into it's own library that can be reused by planemo. This library can than co-exist with whatever will be the successor of tool_dependencies.
I kind of feel this is a lot code duplication.

I don't want to feel to negative here, this is awesome and we should get this functionality in as soon as possible, I'm just wondering if more code-reuse would be beneficial.

peterjc:
As discussed on #303, long term moving the TS installation code into its own library makes a lot of sense. I'm happy to regard the initial internals of ``planemo dependency_script`` as a useful prototype, which can later call the TS installation code library once that is abstracted out of the Galaxy core.

bgruening:
@peterjc ok :) Still catching up with all the new stuff that happened.

martenson:
bikeshedding, yay! 
@peterjc @jmchilton: what about just `script` (as a verb) or  `scriptify`?

peterjc:
I kinda like ``planemo scriptify`` but ``planemo script`` seems overly ambiguous. Using ``planemo dependency_script`` is verbose but otherwise I like it - very clear.

jmchilton:
I like ``scriptify`` actually. I don't like planemo declaring tool shed dependencies are "the way" to do dependencies.

peterjc:
If there are no other views on the name, I'll happily change it to ``planemo scriptify`` (almost as short as ``planemo depbash``)

nsoranzo:
I think that `scriptify` is almost as ambiguous as `script`, maybe `dep_script` or `dep_scriptify`?

bgruening:
`script` or `scriptify` does not tell me much as user. `dep_script` or even better `dep_bash` is more informative. `dep_bash` would give us the possibility to implement different `dep_*` commands in the future.

peterjc:
OK, ``planemo script`` and ``planemo scriptify`` rejected as too vague.

I've changed this to ``planemo dependency_script``. If @jmchilton agrees, I like the shorter ``planemo dep_script``.

But on to the code, is this good enough to be useful now?

jmchilton:
@peterjc In terms of the code - I would say I am going to let you decide if it is useful enough for now. It seems like a good start to me. So :+1:.

@bgruening and I have discussed stuff like this before (duplicating the tool shed functionality outside the tool shed, extracting a library, etc...) - and my position has been that I don't intend to personally support the effort. If there are things I can do to enable it, I certainly will and I have no intention to cause the code the stagnate, but I think it is more important to spend time replacing this dependency resolution with homebrew than augmenting it. That is just my personal opinion though, not everyone on the team agrees with me.

jmchilton:
As for bike shedding, I still think ``scriptify`` is best and we should attempt to override that veto :smile:. My ranked vote is:

 - ``scriptify``
 - ``dependecy_script``
 - ``dep_script``

``dep_script`` is fine if that is what you would like to do.

peterjc:
After some frustration leading to this hack https://github.com/peterjc/planemo/commit/7073505d0524c01f2777dfafd84a5ba3350c7528 I finally got the basic tests to pass on TravisCI.

Should I do any branch history cleanup, or leave this as it is for any merge?

peterjc:
Minor update to fix the tests since the contents of ``tests/repos/`` changed on the master.

@jmchilton Would you prefer a cleaned-up history? Or just do a ``--squash`` merge?

jmchilton:
My preference would be this be squashed into a single commit with a nice commit message if you are offering :smile:, I don't really enforce this and don't always do it myself and so I have no problem with it being merged as is.

peterjc:
Thanks @jmchilton - done as a squashed merge: https://github.com/galaxyproject/planemo/commit/f798c7e29b2276ce68b828e72fc6a6460c73792b

PR #311, opened 2015-09-30
Fixed detection of venv in makefile

lparsons:
The setup-venv function in the Makefile was not properly detecting the need to create a venv and thus resulted in installing all the packages in the environment that make was executed in.


jmchilton:
Great, thanks!

PR #312, opened 2015-09-30
Updated docs for serve command

lparsons:
Added info about the use of virtualenv to the serve command docs.


jmchilton:
Awesome, thanks!

jmchilton:
Wait - actually that file is auto-generated. I'll make the change - sorry.

lparsons:
Ooops! Sorry about that.

lparsons:
Where is that text in the code? I can see where the options help comes from, but the the general command help.

PR #314, opened 2015-10-02
Implement profiles in sheds section of the ~/.planemo.yml.

jmchilton:
Each shed section can now have a ``username`` which will be used in lieu of the top-level ``shed_username`` in that config. Each ``sheds`` keys can now specify a URL - existing sections (``toolshed``, ``testtoolshed``, and ``local``) will be assigned URLs automatically if they are specified and are somewhat special in that respect.

But new profiles can be created with arbitrary names and used in conjunction with ``--shed_target`` as described by @erasche here https://github.com/galaxyproject/planemo/pull/224#issuecomment-106058005.

All of planemo's unit tests pass - but I was hoping since many people have complained about this problem someone would volunteer to test this for me and report bugs.

### DIFF
--- a/docs/.planemo.yml
+++ b/docs/.planemo.yml
@@ -8,7 +8,6 @@
 #  password: <password>
 
 sheds:
-  username: "<TODO>"
   # For each Tool Shed you wish to target enter the API key
   # or both email and password.
   toolshed:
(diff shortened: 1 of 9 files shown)


erasche:
Will test tomorrow. :)

jmchilton:
Going to just merge this it will make it easier to test.

Issue #320, opened 2015-10-08
Multiple install directories for planemo dependency_script?


peterjc:
The initial `planemo dependency_script` implementation f798c7e (see #303 and #310) produces a single `dep_install.sh` and `env.sh` script even when called on multiple `tool_dependencies.xml` files. The simplistic usage model is to run `bash dep_install.sh` and `source env.sh` prior to running Galaxy or any Galaxy tests.

Issue #319 addresses where to put a separate `env.sh` scripts for each tool under the `tool_dependency_dir` folder set in `galaxy.ini` or via `planemo test --tool_dependency_dir <dep_dir> ...`

This issue is about a similar approach for where to put the files created by the installation itself, i.e. automatically adjusting the `$INSTALL_DIR` for multiple `tool_dependencies.xml` files. This would allow co-installation of clashing filenames (e.g. multiple versions of MIRA or BLAST).

Motivating example https://github.com/peterjc/galaxy_mira/commit/0d41761e7073f8fac4c9a4b2a34a95e839ef76f1 and https://travis-ci.org/peterjc/galaxy_mira/builds/84351148 - in this case `planemo dependency_script` is used on three `tool_dependencies.xml` files to install three versions of MIRA. The clashing binary name `mira` requires installation into separate folders and has implications for determining priority on `$PATH` (which would be addressed by #319). The workaround as shown is to call `planemo dependency_script` three times in a particular order.

peterjc:
According to https://wiki.galaxyproject.org/Admin/Config/ToolDependencies if we put binaries under e.g. ``<dep_dir>/MIRA/3.4.1.1/bin/`` then they will automatically be detected by Galaxy and added to ``$PATH`` if a tool XML declares a dependency using:

```
<requirement type="package" version="3.4.1.1">MIRA</requirement>
```

Many ``tool_dependencies.xml`` files do simply download and compile/uncompress binaries and move them to ``$INSTALL_DIR``. So perhaps it would make sense to ``<dep_dir>/<name>/<version>/bin/`` as an option for where the bash script should put things?

bgruening:
Not sure if this helps but you could also create `<dep_dir>/MIRA/default` as a symlink to a specific folder with an `env.sh` file or a `bin/`.
Otherwise I think `<dep_dir>/MIRA/3.4.1.1/bin/` is a good idea.

peterjc:
[...]

PR #321, opened 2015-10-12
First action determines pwd for tool_dependencies.xml recipes

peterjc:
This should resolve an issue discussed on GitHub issue #320 where most `tool_dependencies.xml` recipes are run in a temporary working directory, but others are run in the target `$INSTALL_DIR`
instead.

Within https://github.com/peterjc/pico_galaxy only `clinod` and `effectiveT3` were affected.

I didn't spot any other cases within:
- https://github.com/peterjc/galaxy_blast
- https://github.com/peterjc/galaxy_mira
- https://github.com/galaxyproject/tools-devteam
- https://github.com/gregvonkuster/galaxy-csg

### DIFF
--- a/planemo/commands/cmd_dependency_script.py
+++ b/planemo/commands/cmd_dependency_script.py
@@ -54,6 +54,14 @@ export INSTALL_DIR=${INSTALL_DIR:-$PWD}
 export INSTALL_DIR=`(cd "$INSTALL_DIR"; pwd)`
 """
 
+# Installation recipes using one of these as the opening action use
+# a temporary working directory, otherwise use $INSTALL_DIR - see
+# class InstallToolDependencyManager in Galaxy's
+# lib/tool_shed/galaxy_install/install_manager.py
+INSTALL_ACTIONS = ['download_binary', 'download_by_url', 'download_file',
+                   'setup_perl_environment', 'setup_python_environment',
+                   'setup_r_environment', 'setup_ruby_environment', 'shell_command']
+
 
 def find_tool_dependencis_xml(path, recursive):
     """Iterator function, quick & dirty tree walking."""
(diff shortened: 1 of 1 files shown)


peterjc:
Four maybes under https://github.com/galaxyproject/tools-iuc though:

I think the the no-op for Mac OS X in ``package_atlas_3_10`` is harmless in https://github.com/galaxyproject/tools-iuc/blob/master/packages/package_atlas_3_10/tool_dependencies.xml

To me ``package_vcflib_8a5602bf07`` looks odd, I left a comment for @martenson: https://github.com/galaxyproject/tools-iuc/commit/348398c421a99b1e2b3128cb9c8404e83218a659#diff-b0ee1f84942da2a3633fe43a6743f60b 

Do these look fine @bgruening - specifically the first step of the generic architecture action?:
- ``package_meme_4_10_0`` - https://github.com/galaxyproject/tools-iuc/blob/master/packages/package_meme_4_10_0/tool_dependencies.xml
- ``package_bcftools_1_2`` - https://github.com/galaxyproject/tools-iuc/blob/master/packages/package_bcftools_1_2/tool_dependencies.xml - OK?

peterjc:
As an alternative to adding this special to planemo, in https://github.com/galaxyproject/galaxy/issues/896 I am suggesting removing the special case from Galaxy instead.

bgruening:
> Do these look fine @bgruening - specifically the first step of the generic architecture action?:
- ``package_meme_4_10_0`` - https://github.com/galaxyproject/tools-iuc/blob/master/packages/package_meme_4_10_0/tool_dependencies.xml

`set_environment_for_install` needs to be below the `download_by_url` action afaik.

- ``package_bcftools_1_2`` - https://github.com/galaxyproject/tools-iuc/blob/master/packages/package_bcftools_1_2/tool_dependencies.xml - OK?

same here, `set_environment_for_install` needs to be below the `download_by_url` action

peterjc:
If @bgruening is happy to adjust the ``tool_dependencies.xml`` for  ``package_bcftools_1_2`` and ``package_meme_4_10_0``, and I also to rework ``clinod`` and ``effectiveT3``, then my proposal in https://github.com/galaxyproject/galaxy/issues/896 seems more promising...

peterjc:
I have reworked the ``tool_dependencies.xml`` for ``clinod`` and ``effectiveT3`` to avoid the special case by starting them with a download action:

- https://github.com/peterjc/pico_galaxy/commit/f90c64301b6f23684bd7ea2e430d0a12d8c164b7
- https://github.com/peterjc/pico_galaxy/commit/dd1725e0a07add1eed8f6ad7efaeaad2a2e3e961

This leaves a couple of entries in https://github.com/galaxyproject/tools-iuc  to review, but it seems this corner case is quite rare. I think this points in favour of "fixing" Galaxy rather than adding this quirk to planemo: https://github.com/galaxyproject/galaxy/issues/896

bgruening:
LGTM!

jmchilton:
Even if we fix all these recipes, we still need to merge this right?

peterjc:
My personal preference is to fix Galaxy to remove this quirk https://github.com/galaxyproject/galaxy/issues/896 and just leave ``planemo dependency_script`` as it is.

It would be easy to check all the existing recipes on the tool shed (run  ``planemo dependency_script`` from this branch and grep for ``# Note this recipe does not use a temp working directory:``), if there is any easy way to bulk download them all...

peterjc:
Galaxy has moved away from the XML dependency recipes and now uses conda instead, closing this.

PR #322, opened 2015-10-12
include basic shed_update xunit report option

martenson:
aims to make jenkins's reports more usable


jmchilton:
Awesome continuation of this idea.

PR #325, opened 2015-10-13
when repo update is ok but metadata is not put it in 'skipped' category

martenson:
this will help with Jenkins tracking

### DIFF
--- a/planemo/reports/update_xunit.tpl
+++ b/planemo/reports/update_xunit.tpl
@@ -15,8 +15,8 @@
             </error>
         {% endif %}
         {% if testcase.result == 3 %}
-            <error type="planemo.FailedMetadata" message="Failed to update repo metadata">
-            </error>
+            <skipped type="planemo.FailedMetadata" message="Repo updated but metadata was not">
+            </skipped>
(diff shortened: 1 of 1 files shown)


erasche:
Skipped? Not sure I agree eith that. If metadata id failing, tjat should be a failure yes?

martenson:
I was always confused by this. The command did everything right but then .shed yml does not have 'description' in it and that should cause it to fail?

nsoranzo:
Fixing `.shed.yml` seems the best solution to me, the synopsis/short description is mandatory.

erasche:
> I was always confused by this. The command did everything right but then .shed yml does not have 'description' in it and that should cause it to fail?

yep ;) making a fix mandatory is the best way to ensure it gets done

peterjc:
Making the synopsis/short description is mandatory seems sensible from an IUC standards point of view, so +1 on that idea.

jmchilton:
``shed_upload`` should work without it and so should ``update --skip_metadata``, find if updating the metadata require a description though - that seems reasonable.

PR #333, opened 2015-10-19
Add docker pull attempt when missing Dockerfile

kellrott:
For the 'docker_build' command, in the case that a Dockerfile is not available, planemo will try to pull the dependency from the registry, using the tag found in the tool description, rather then fail.


jmchilton:
Awesome sauce!

PR #334, opened 2015-10-19
Improving docker_shell

kellrott:
Adding the '-t' flag to the docker shell command (add psuedo terminal) as well as mounting the tool directory and the current working directory, also setting working directory for docker.

### DIFF
--- a/planemo/commands/cmd_docker_shell.py
+++ b/planemo/commands/cmd_docker_shell.py
@@ -10,7 +10,7 @@ such a tag from a Dockerfile located in the tool's directory.
 """
 from __future__ import print_function
 import click
-
+import os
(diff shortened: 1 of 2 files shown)


jmchilton:
This is seems good, thanks. One question I'm not sure the answer - instead of using the fixed paths should we mount the volumes in specific places instead ``/working``, ``/tool``?

kellrott:
I'm mimicking the Galaxy behavior of mounting things to the same path inside the container. But that can be changed. Do you want it that way?

jmchilton:
This is fine for now, it is simple and you are right - very much the naive approach of Galaxy. Thanks for these enhancements @kellrott they make a lot of sense to me.

PR #343, opened 2015-10-22
Add name to all XUnit reports

erasche:
This seems to be required, Jenkins isn't displaying them since they've gone missing.


erasche:
@martenson awesome, thanks, just got the email and patched that before noticing you had.

erasche:
@martenson merge if you're happy and then let's re-trigger those diffs/uploads.

martenson:
thanks @erasche

PR #344, opened 2015-10-22
Capture stdout/stderr for shed_diff and shed_update XUnit reports

erasche:
@jmchilton would you mind putting eyes to this if you have time? @martenson you might be interested in this as well.

# The Good

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="planemo-update"
           tests="1"
           errors="0"
           failures="1"
           skip="0">
    <testcase classname="seqtk" name="" time="6.95435190201">
            <error type="planemo.FailedUpdate" message="Failed to update repository">
            </error>
        <system-out>
        cd '/home/hxr/work/tools-iuc/tools/seqtk' &amp;&amp; git rev-parse HEAD
        cd '/home/hxr/work/tools-iuc/tools/seqtk' &amp;&amp; git diff --quiet
        </system-out>
        <system-err>
        Could not update seqtk
        Unexpected response from galaxy: 400: {"content_alert": "", "err_msg": "No changes to repository."}
        </system-err>
    </testcase>
</testsuite>
```

XUnit reports for shed_update and shed_diff will now contain stdout/stderr, and make those available to browse in Jenkins

# The Bad

Logs under `--report_xunit` are no longer pretty colours. They're boring black and white. Seems like an acceptable loss since you gain awesome XUnit reports.

If `--report_xunit` is not specified, tests are still colourful.

# The Ugly

This PR provides a small class called `planemo.io.Capturing` which lets you capture stdout/stderr for any function calls, and then process that data at your convenience. You don't have to rearchitect any of the existing infrastructure for logging user facing messages, you can just capture it and deal with it as you please. This code smells and it's pretty stinky. Instead of some way to handle message propagation properly and dispatching messages to handlers which care about them, we monkey patch `sys.std*`. I am not proud of this, but it solves the problem simply and quickly with very few changes.

The whole of `cmd_shed_update` and `cmd_shed_diff` are probably in dire need of a refactoring

### DIFF
--- a/planemo/commands/cmd_shed_diff.py
+++ b/planemo/commands/cmd_shed_diff.py
@@ -1,7 +1,6 @@
 """
 """
 import sys
-import time
 import tempfile
 import shutil
 from xml.sax.saxutils import escape
@@ -12,7 +11,7 @@ from planemo.cli import pass_context
 from planemo import options
 from planemo import shed
 from planemo.reports import build_report
-from planemo.io import Capturing, tee_captured_output
+from planemo.io import captured_io_for_xunit
 
 
 @click.command("shed_diff")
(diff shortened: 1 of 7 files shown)


erasche:
(Rebasing + fixing tests now...)

jmchilton:
Timestamps gone crazy, github is not displaying this well for me, but I tried refactoring a lot of the duplicated functionality out - hope this okay @erasche. Another upshot is that less of planemo.io needs to be exposed as well. Not sure it actually works yet though :).

erasche:
@jmchilton oh wow, that is *absolutely* ok. I definitely should've seen that easy place to refactor. I'll give it a test locally.

erasche:
@martenson did you have opinions on this?

martenson:
I am not experienced with stream handling so I cannot comment on the architecture but I like the functionality and simple approach. Thanks @erasche

jmchilton:
This is fine for now, the approach needs to be reworked to be thread safe someday (i.e. log the output instead of capturing it) if we want to build a planemo web app.

erasche:
Agreed. I just didn't have the time/energies for a full rewrite of all echoing to logging. Will have to contribute that someday.

A planemo web app? Are we building our own IDE? Neato.

jmchilton:
Less of an IDE than just a shed management app: https://github.com/galaxyproject/planemo/issues/169#issuecomment-98158780

PR #345, opened 2015-10-23
Try and handle UTF-8

erasche:
Planemo crashes at the very last step trying to output data. Here we encode to xml-safe ascii instead.


erasche:
@jmchilton/@natefoo Travis seems to be stuck in a loop. #wheelfallout  :(

```
The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).

The build has been terminated.
```

jmchilton:
Flying blind then, doh.

natefoo:
Galaxy starts a lot of times, even pre-wheels. I am not sure I understand the entire test process here, I dug into it a little bit last week but will need to do more to understand what's happening.

PR #330, opened 2015-10-16
Xunit improvements

erasche:
- Refactors XUnit report templates into a single more generic template. (http://stackoverflow.com/a/9691131 good resource for what fields are available)
- Unrelated mktemp -> mkstemp
- Include timings in shed_diff (not useful but hey, neato)
- Include diffs in shed_diff (hella useful)

PR #349, opened 2015-10-27
Add command to check_urls

### LINKED ISSUE #228: URL Checking functionality for packages
As a **stopgap** measure (until I've got a distributed, fault-tolerant package storage service online), I'd like to add a feature to planemo to check URL validity, especially in light of issues like https://github.com/galaxyproject/tools-iuc/pull/189

This feature would likely take the form of a `--validate_urls` flag in `shed_lint`, which would urlopen+read the first few bytes of each url, additionally checking the return code for equality to 200

erasche:
Fixes #228. Implemented as a second command, hope that's OK because `shed_lint` doesn't support XUnit (yet). These could probably stand to be merged eventually? Hmm.


erasche:
Will go back and add the `--verify` option later since that'll require re-architecting shed2tap a bit to track sha256sums for packages/urls.

peterjc:
The user-facing documentation could be clearer about which files will it check for URLs, e.g. ``tool_dependencies.xml`` etc?

peterjc:
Clearer now, yes. Should it also check ``repository_dependencies.xml`` as well?

erasche:
Do those ever contain URLs that are downloaded? Grepping through the IUC's codebase they look like just (internal to the TS) package references.

yhoogstrate:
I tested whether the code works with:
 - ftp urls :white_check_mark:
 - urls to file with size < 100 bytes :white_check_mark:
 - redirecting urls (302 response) :white_check_mark:

bgruening:
@erasche `repository_dependencies.xml` should never contain URLs, but tools can. It would be nice to also test all URL's in tools. I especially have help text and further documentation links in my mind.

erasche:
@bgruening any examples? 

@yhoogstrate wow, excellent, thanks for testing!

bgruening:
@erasche bedtools for example: https://github.com/galaxyproject/tools-iuc/blob/master/tools/bedtools/macros.xml

jmchilton:
I'd prefer this not be added in as a separate command - I'd like it added into ``shed_lint`` and ``lint`` with a ``--urls`` flag. I can make these modifications to enable this if you would like.

erasche:
@jmchilton okay, or I'll get to it tonight. I separated the command due to clean XUnit report, whereas adding to lint would require figuring out the XUnit for all the linting stuff.

erasche:
@bgruening I'm not sure about testing documentation links. I'll implement it if you have any methods for "finding URLs in free-text fields".

bgruening:
@erasche every link we can find and check is worth doing it right. It's not so important to find every hidden and strange URL but if we can find as many as possible and check them, this would be great.

I would simple go with a regex, like these ones here:
http://stackoverflow.com/questions/7676255/find-and-replace-urls-in-a-block-of-text-return-text-list-of-urls and catch as many as possible, not caring about false negative ones.

peterjc:
I'd agree with @bgruening - a simple regex to catch likely URLs in things like the tool XML help text or README files seems like a good compromise - if my documentation points to a dead link I'd like to know.

erasche:
@bgruening great, thanks for that link. I'll get this merged into lint and added in tonight.

bgruening:
@erasche thanks, you are the best!

jmchilton:
@erasche Don't hesitate to only add xunit reports for URLs to the linting stuff with a TODO for the rest of it if that makes it easier for now. At least that way I feel like we are moving forward with the API we want for planemo - it just is missing some features (like linting stuff besides urls).

erasche:
@jmchilton okay, understood. :)

erasche:
@jmchilton okay, removed XUnit stuff for the time being. If you have any thoughts on #330 let me know, but I think this should address your issues with the separate command.

jmchilton:
@erasche Out of the office for awhile, but will you review  some style fixes I added in 0cd1d6fbb27f6239c9c82fb278ae33835def43ad and then merge?

erasche:
@jmchilton those all look fine. I don't mind the destroyed indentation, there's only ONE reading of that snippet so .. meh. That snippet will probably be removed later. Sorry, I shouldn't leave dead code around. Otherwise looks gerat.

jmchilton:
Merged - thanks for the help all. @erasche I'll think about the XUnit questions from IRC.

erasche:
okay, sounds good. there's persistent discussion in #330, **no rush**

jmchilton:
Actually there is a missing file - @erasche do you have a ``git add planemo/planemo/linters/urls.py``?

erasche:
oh shoot. Ok, pushing that now.

erasche:
@jmchilton follow up PR in https://github.com/galaxyproject/planemo/pull/358

Issue #232, opened 2015-06-17
Linting rules for tool_type="data_source"


erasche:
- [x] no tests
- [x] ~~sometimes no help (but I'm adding dummy help)~~ (agreed this should still be there)
- [x] ~~basically never citations (since not rerunnable)~~ (agreed these should still be there)
- [x] `<display/>` input type
- [x] `<uihints />` are also present.

jmchilton:
Turned this issue into a check list - hope that is okay.

erasche:
absolutely

Issue #194, opened 2015-05-08
Check for missing <when/> blocks


erasche:
Since Galaxy complains about this, it'd be reasonable for planemo to lint on it as well.

xref https://github.com/galaxyproject/tools-devteam/pull/131 https://github.com/galaxyproject/tools-iuc/pull/130

martenson:
Given that Galaxy falls back with reasonable defaults (and the number of these warnings) this might be better when the warnings are removed form Galaxy itself.

jmchilton:
I feel like it is a good warning - it will catch typos and encourage people to be explicit.

nsoranzo:
Yes, please!

peterjc:
+1 on warning about this in planemo (and keeping the warnings in Galaxy)

peterjc:
I just saw the opposite problem which ``planemo lint`` could also spot: ``<when/>`` blocks which did not match the select, e.g. something like this:

```
        <conditional name="input_type_conditional">
            <param name="input_type" type="select" label="Input Type">
                <option value="single" selected="true">Single Dataset</option>
                <option value="paired_collection">Paired Collection</option>
                <!-- note missing paired_files here -->
            </param>
            <when value="single">
                 <param name="fastq_input1" type="data" format="fastqsanger" label="Select fastq dataset" help="Specify dataset with single reads"/>
            </when>
            <when value="paired_files">
                 <param name="fastq_input_f" type="data" format="fastqsanger" label="Select forward fastq dataset" help="Specify dataset with forward reads"/>
                 <param name="fastq_input_r" type="data" format="fastqsanger" label="Select reverse fastq dataset" help="Specify dataset with reverse reads"/>
            </when>
            <when value="paired_collection">
                 <param name="fastq_input" type="data_collection" format="fastqsanger" label="Select dataset pair" help="Specify paired dataset collection containing paired reads" />
            </when>
        </conditional>
```

peterjc:
Could a small extension to this catch the reverse problem as in my example above?

erasche:
@peterjc that's covered in #350. https://github.com/galaxyproject/planemo/pull/350/files#diff-4179932f7b3dedef8f42d33fb6259e01R61

peterjc:
Great, thanks.

PR #350, opened 2015-10-27
Lint extra

erasche:
Whoops. Forgot to open the PR. Should close out some old issues.

- #232 
- #194 

little to no motivation to add tests ... so ... yeah.

### DIFF
--- a/planemo_ext/galaxy/tools/linters/inputs.py
+++ b/planemo_ext/galaxy/tools/linters/inputs.py
@@ -26,7 +26,7 @@ def lint_inputs(tool_xml, lint_ctx):
     if num_inputs:
         lint_ctx.info("Found %d input parameters.", num_inputs)
     else:
-        lint_ctx.warn("Found not input parameters.")
+        lint_ctx.warn("Found no input parameters.")
 
 
 def lint_repeats(tool_xml, lint_ctx):
--- a/planemo_ext/galaxy/tools/linters/inputs.py
+++ b/planemo_ext/galaxy/tools/linters/inputs.py
@@ -2,6 +2,7 @@ from ..lint_util import is_datasource
 
 
 def lint_inputs(tool_xml, lint_ctx):
+    datasource = is_datasource(tool_xml)
     inputs = tool_xml.findall("./inputs//param")
     num_inputs = 0
     for param in inputs:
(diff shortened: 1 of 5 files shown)


planemo_ext/galaxy/tools/linters/inputs.py
```diff
@@ -23,10 +25,40 @@ def lint_inputs(tool_xml, lint_ctx):
             if "format" not in param_attrib:
                 lint_ctx.warn("Param input [%s] with no format specified - 'data' format will be assumed.", param_name)
         # TODO: Validate type, much more...
+
+    conditional_selects = tool_xml.findall("./inputs//conditional")
+    for conditional in conditional_selects:
+        select = conditional.find('./param[@type="select"]')
+        # Should conditionals ever not have a select?
```
bgruening: It can have a boolean. According to https://github.com/galaxy-iuc/standards/blob/master/docs/best_practices/tool_xml.rst#booleans we do not recommend to use booleans for a conditional but it is possible.

planemo_ext/galaxy/tools/linters/inputs.py
```diff
@@ -23,10 +25,40 @@ def lint_inputs(tool_xml, lint_ctx):
...
+        if not len(select):
+            lint_ctx.info("Conditional without <param type=\"select\" />")
+            continue
+
+        select_options = select.findall('./option[@value]')
+        select_option_ids = [option.attrib['value'] for option in select_options]
```
nsoranzo: Maybe here it would be good to call `lint_ctx.error()` if the `<option>` has no `value` attribute.

planemo_ext/galaxy/tools/linters/inputs.py
```diff
@@ -23,10 +25,40 @@ def lint_inputs(tool_xml, lint_ctx):
...
+        whens = conditional.findall('./when')
+        when_ids = [when.attrib['value'] for when in whens]
```
nsoranzo: Maybe here it would be good to call `lint_ctx.error()` if the `<when>` has no `value` attribute.

planemo_ext/galaxy/tools/linters/inputs.py
erasche: interesting. Thanks for that, I'll update it accordingly

erasche:
Comments addressed.

planemo_ext/galaxy/tools/linters/xml_order.py
```diff
@@ -17,16 +17,37 @@
     'citations',
 ]
 
+DATASOURCE_TAG_ORDER = [
+    'description',
+    'command',
+    'inputs',
+    'request_param_translation',
+    'uihints',
```
jmchilton: Are uihints even respected for data source tools? Hmm... we can revisit that later though.

Can you add macros and configfiles back in the same order as normal tools?

jmchilton:
Good changes, I'll just create an issue to add tests after it is merged.

planemo_ext/galaxy/tools/linters/xml_order.py
erasche: That's an awful good question regarding uihints.

Yes, sorry, I removed everything I didn't see during very brief greps through the tools directory.

erasche:
@jmchilton okay, sounds good. Thanks for being lenient :)

jmchilton:
I'm going to merge this then - we will figure out Travis again someday I guess.

PR #359, opened 2015-10-31
Check urls - shed_lint

erasche:
related: #360

### DIFF
--- a/planemo/shed_lint.py
+++ b/planemo/shed_lint.py
@@ -1,3 +1,4 @@
+from __future__ import absolute_import
 import os
 import yaml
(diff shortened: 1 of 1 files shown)


planemo/shed_lint.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import absolute_import
```
bgruening: Is this still needed?

erasche: Yes, I wasted a good 10 minutes confused by what was happening, until I `import xml; print dir(xml)` and realised that python was importing `planemo.xml`.

PR #394, opened 2015-12-20
Depend on new galaxy-lib on PyPI instead of planemo_ext hacks...

jmchilton:
"Showing  111 changed files  with 5 additions and 9,176 deletions." :smiley:

I think now that Galaxy requires a virtualenv and planemo always sets one up, we shouldn't have Python package bleeding that occurred before the changes to planemo_ext that addressed that problem.


erasche:
Bloody beautiful @jmchilton.
@galaxybot test this

PR #398, opened 2016-01-01
cast repo_id as string for shed compat

erasche:
Seriously the only thing :) Since repo_ids are now returned as `int`s rather thans `str`s, we cast them as a string and we're done.


erasche:
Okay, pretty sure test failures aren't my fault...

bgruening:
Interesting something broke with the installation. I had the some problem today in galaxytools: `ValueError: malformed string`. I simple pip install planemo fixed it for me. Before I was installing from github directly.

jmchilton:
Rebased this against the latest master which fixes though planemo install bugs. I have to stop tolerating @jmchilton turning these projects turning red.

Issue #399, opened 2016-01-11
planemo test broken with conda python


mvdbeek:
Not quite sure if this is an issue with conda, galaxy or planemo ...
```
(/home/galaxy/miniconda)galaxy@vm0148:~$ conda create --name python_2_7_10 python==2.7.10
Fetching package metadata: ........
Solving package specifications: ........
Package plan for installation in environment /home/galaxy/miniconda/envs/python_2_7_10:

The following NEW packages will be INSTALLED:

    openssl:    1.0.2d-0     
    pip:        7.1.2-py27_0 
    python:     2.7.10-2     
    readline:   6.2-2        
    setuptools: 19.1.1-py27_0
    sqlite:     3.8.4.1-1    
    tk:         8.5.18-0     
    wheel:      0.26.0-py27_1
    zlib:       1.2.8-0      

Proceed ([y]/n)? y

Linking packages ...
[      COMPLETE      ]|#############################################################################################################################################################################| 100%
#
# To activate this environment, use:
# $ source activate python_2_7_10
#
# To deactivate this environment, use:
# $ source deactivate
#
(/home/galaxy/miniconda)galaxy@vm0148:~$ source activate python_2_7_10
discarding /home/galaxy/miniconda/bin from PATH
prepending /home/galaxy/miniconda/envs/python_2_7_10/bin to PATH
(python_2_7_10)galaxy@vm0148:~$ pip install planemo
Collecting planemo
  Using cached planemo-0.21.1-py2.py3-none-any.whl
Collecting cwltool (from planemo)
Collecting PyGithub (from planemo)
Collecting Click (from planemo)
  Using cached click-6.2-py2.py3-none-any.whl
Collecting virtualenv (from planemo)
  Using cached virtualenv-13.1.2-py2.py3-none-any.whl
Collecting jinja2 (from planemo)
  Using cached Jinja2-2.8-py2.py3-none-any.whl
Collecting six>=1.7.0 (from planemo)
  Using cached six-1.10.0-py2.py3-none-any.whl
Collecting docutils (from planemo)
Collecting glob2 (from planemo)
Collecting pyyaml (from planemo)
  Using cached PyYAML-3.11.tar.gz
Collecting bioblend (from planemo)
Collecting rdflib-jsonld>=0.3.0 (from cwltool->planemo)
Collecting requests (from cwltool->planemo)
  Using cached requests-2.9.1-py2.py3-none-any.whl
Collecting shellescape (from cwltool->planemo)
[...]

PR #400, opened 2016-01-13
Remove shell_exec

bgruening:
Install conda do not have any `shell_exec` parameter anymore: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/deps/conda_util.py#L261

### DIFF
--- a/planemo/commands/cmd_conda_init.py
+++ b/planemo/commands/cmd_conda_init.py
@@ -22,5 +22,4 @@ def cli(ctx, **kwds):
     http://docs.continuum.io/anaconda/eula.
     """
     conda_context = build_conda_context(**kwds)
-    return conda_util.install_conda(conda_context=conda_context,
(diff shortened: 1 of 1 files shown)


jmchilton:
Opps sorry, let me know when you are ready for a release.

bgruening:
Will do, I guess this needs to be fixed before: https://github.com/galaxyproject/planemo/issues/399
Maybe this is related because Galaxy still starts virtualenv even if called from planemo.

jmchilton:
@bgruening Planemo and Galaxy don't work well inside conda environmnts - this isn't a new a problem and I am happy to do a release that supports running tools in conda environments even if planemo and Galaxy still can't be.

bgruening:
Sounds good to me. Would be nice to get tools-iuc under testing and motivate me a little more to migrate more deps :)

PR #418, opened 2016-03-01
Nice makefile help

erasche:
As promised in https://github.com/galaxyproject/galaxy/pull/1824

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,74 +21,55 @@ TEST_DIR?=tests
 .PHONY: clean-pyc clean-build docs clean
 
 help:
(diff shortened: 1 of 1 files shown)


jmchilton:
Clever :smile:

jxtx:
Except that `grep -P` is a gnuism, not posix, doesn't work on os x.

erasche:
@jxtx right, people use OSX. I'll try and re-write without it.

jxtx:
I might be tilting at windmills, but I must fight GNUisms where I find them. Simple fix in #419.

erasche:
Thanks @jxtx, I appreciate the cross-platform fix! :)

PR #419, opened 2016-03-01
Use egrep for make help, works on posix / os x.

jxtx:


erasche:
Patched into galaxyproject/galaxy#1824

PR #427, opened 2016-03-29
Drop Python 2.6 Support

jmchilton:
As announced on -dev.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,10 +4,15 @@
 import ast
 import os
 import re
+import sys
 try:
     from setuptools import setup
 except ImportError:
     from distutils.core import setup
+
+if sys.version_info < (2, 7):
+    sys.stderr.write("ERROR: planemo requires at least Python Version 2.7\n")
+    sys.exit(1)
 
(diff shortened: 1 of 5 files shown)


peterjc:
Can we add something like this to ``setup.py`` as well please?

```
if sys.version_info < (2, 7):
    sys.stderr.write("ERROR: planemo requires at least Python Version 2.7\n")
    sys.exit(1)
```

setup.py
```diff
@@ -114,7 +114,6 @@ def get_var(var_name):
         'Topic :: Software Development :: Testing',
         'Natural Language :: English',
         "Programming Language :: Python :: 2",
-        'Programming Language :: Python :: 2.6',
         'Programming Language :: Python :: 2.7',
```
nsoranzo: Python 3.4 is missing here.

jmchilton:
@peterjc Done, thanks for the suggestion.

@nsoranzo Let's wait until the tests consistently pass before adding that. It is definitely the goal though.

PR #442, opened 2016-04-12
Add tox linters for import ordering and docstrings.

jmchilton:
This projects fails hugely on both metrics, but tracking is a start.


nsoranzo:
:+1: 

Can we have this for Galaxy too? ;)

jmchilton:
@nsoranzo - I'd add it right now if it were up to me alone :smile:. Let's work toward it - add to it to planemo, pulsar, bioblend, starforge - get it working and then make a case that it is useful.

PR #457, opened 2016-04-19
port docs about collection as an output from bb

martenson:
original here: https://bitbucket.org/galaxy/galaxy-central/pull-requests/634/allow-tools-to-explicitly-produce-dataset/diff

ping @jmchilton @lparsons


jmchilton:
I prefer documentation to just remain forever as dated, unlinked, and unsearchable pull requests. I'm closing this as WONTFIX

Issue #433, opened 2016-04-07
planemo conda_install issue


lecorguille:
Hi,

Some of my tools need the brand new conda package [bioconductor-camera](https://bioconda.github.io/recipes/bioconductor-camera/README.html)
```xml
    <requirements>
        <requirement type="package" version="1.22.0">bioconductor-camera</requirement>
    </requirements>

```

It can be install without trouble: `conda install bioconductor-camera`

But when I use `planemo conda_install .` to build an env (if I understand well), I get this:
```bash
Linking packages ...
Error: ERROR: placeholder '/root/miniconda3/envs/_build_placehold_placehold_placehold_placehold_placehold_p' too short in: ncurses-5.9-4
```

The full path of this uncomplete env is `/home/lecorguille/miniconda2/envs/__package__bioconductor-camera@__version__1.22.0`

I looked around and find those threads without understand which path or filename is too long:
https://github.com/rtfd/readthedocs.org/issues/1902
https://github.com/conda/conda/issues/1087

Any idea?

yguitton:
Sorry, Never heard of that conda dependency, it is not listed in CAMERA dependencies "http://bioconductor.org/packages/release/bioc/html/CAMERA.html"

lecorguille:
@yguitton It's a R dependency: https://github.com/conda/conda-recipes/blob/master/r-base/meta.yaml

But it seem to come from my setups (CentOS and Ubuntu). Because travis test pass on travis-ci.org when I use `planemo conda_install ${TRAVIS_BUILD_DIR}/galaxy/xcms_suite ` (https://github.com/workflow4metabolomics/xcms/blob/master/.travis.yml)

So if you (planemo developers) are considering that it's not link to a planemo issue. I can hear that.

mvdbeek:
It isn't really a planemo issue, the problem is that long link paths are problematic with conda.
So a workaround is to make sure that the directory path where planemo bootstraps galaxy is pretty short.
On OSX, by default $TMPDIR would be something like /var/folders/56/g35vb5h14_v0d6chwtfxvdfr0000gn/T/  and then mkdtemp() would give you a path like /var/folders/56/g35vb5h14_v0d6chwtfxvdfr0000gn/T/ tmpi35zzj which exceeds the limit (since in addition the is a path structure where the conda dependencies are linked to) and you get the error mentioned in the opening post.
At least on OSX I can work around the issue with PR #460

lecorguille:
[...]

PR #460, opened 2016-04-24
Try to shorten config_directory path

mvdbeek:
This circumvents 
```
Error: ERROR: placeholder '/Users/aaronmeurer/anaconda/envs/_build_placehold_placehold_placehold_placehold_' too short in: ncurses-5.9-1
```  
for me.

Conda has limitations in the length of the placeholder path. This can be easily used up on OSX, since OSX provides a long tmpdir path (/var/folders/56/g35vb5h14_v0d6chwtfxvdfr0000gn/T/). If that is the 
case, we can try to use mkdtemp with the /tmp directory.
May also be a workaround for #433.

### DIFF
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -143,6 +143,16 @@ def galaxy_config(ctx, tool_paths, for_tests=False, **kwds):
     if not config_directory:
         created_config_directory = True
         config_directory = mkdtemp()
+        # the following makes sure the transient config_dir path is short
+        # enough for conda linking (https://github.com/conda/conda-build/pull/877)
+        if len(config_directory) > 20:
+            try:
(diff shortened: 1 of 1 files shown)


jmchilton:
What is setting it to not be in ``/tmp`` in your environment? I'm resistant - though not completely opposed to doing this - because we should respect how Planemo's environment is configured when it is possible.

mvdbeek:
I think this is the default behaviour of $TMPDIR on OS X (which we could also test instead of len(config_directory) > 20:).
See http://www.magnusviri.com/OS_X_Admin/what-is-var-folders.html 
Admittedly I could also set config_dir to something short ... so this PR is optional and should be reverted once tjis problem is solved in conda.

jmchilton:
I see - so this will never happen in Linux. I get it now - this looks great, thanks!

PR #462, opened 2016-04-25
Customize the name of the generated html slideshow

remimarenco:
Hi :),

I quickly added a way to add its own html slideshow name. We could think about forcing the extension to `.html` and the location to `docs/` later :).


jmchilton:
Thanks for the fix!

PR #464, opened 2016-05-02
Add IUC standards to Planemo docs.

jmchilton:
No one had any objections to this the other day on IRC but I'll give the IUC one more chance to object :smile:.


bgruening:
:+1: 
I guess asking over the ML or via PR is always good.

PR #484, opened 2016-05-20
Add linter for DOI type citation.

mvdbeek:
I guess this is useful on its own, and may prove more valuable if we support plugging tool information from http://bio.tools into `planemo tool_init`.


jmchilton:
Very nice - I love this!

mvdbeek:
Thanks @jmchilton - seems like I created a [false positive](https://travis-ci.org/galaxyproject/planemo/jobs/131742040#L673) in the tests though

tests/data/tools/fail_doi.xml
```diff
@@ -0,0 +1,31 @@
+<tool id="copy" name="Copy Dataset" version="1.0">
+    <description>copies a dataset</description>
```
jmchilton: The test framework just ensures all ``fail_`` tools fail linting - just rename to invaild_doi or something and the failing test will work.

planemo/lint.py
```diff
@@ -7,8 +8,43 @@
...
+        is_doi(publication, lint_ctx)
+
+
+def find_dois_for_xml(root):
+    dois = []
+    for element in root.root.findall("citations"):
+        for citation in list(element):
+            if "type" in citation.attrib:
+                if citation.attrib["type"] == "doi":
+                    dois.append(citation.text)
```
nsoranzo: I'd substitute the previous 3 lines with:

```python
            if citation.tag == 'citation' and citation.attrib.get('type', '') == 'doi':
                dois.append(citation.text)
```

planemo/lint.py
```diff
@@ -7,8 +8,43 @@
...
+    r = requests.get(url)
+    if r.status_code == 200:
+        if publication_id != doiless_publication_id:
+            lint_ctx.error("%s is valid, but galaxy expects DOI without 'doi:' prefix" % publication_id)
+        else:
+            lint_ctx.info("%s is a valid DOI" % publication_id)
+    elif r.status_code == 404:
+        lint_ctx.error("%s is not a valid DOI" % publication_id)
+    else:
+        lint_ctx.info("dx.doi returned unexpected status code %d" % r.status_code)
```
nsoranzo: `lint_ctx.warn` ?

planemo/lint.py
```diff
@@ -7,8 +8,43 @@
...
+    """
+    Check if dx.doi knows about the publication_id
+    """
+    base_url = "http://dx.doi.org"
+    doiless_publication_id = publication_id.split("doi:", 1)[-1]
+    url = "%s/%s" % (base_url, doiless_publication_id)
+    r = requests.get(url)
+    if r.status_code == 200:
+        if publication_id != doiless_publication_id:
+            lint_ctx.error("%s is valid, but galaxy expects DOI without 'doi:' prefix" % publication_id)
```
nsoranzo: s/galaxy/Galaxy/

tests/data/tools/fail_doi.xml
mvdbeek: Ahh, indeed, that did the trick, thanks!

tests/test_lint.py
```diff
@@ -64,3 +64,8 @@ def test_empty_cdata(self):
         empty_cdata = os.path.join(TEST_TOOLS_DIR, "empty_cdata.xml")
         lint_cmd = ["lint", "--skip", "citations,help", empty_cdata]
         self._check_exit_code(lint_cmd, exit_code=0)
+
+    def test_lint_doi(self):
+        fail_doi = os.path.join(TEST_TOOLS_DIR, "invalid.xml_doi.xml")
```
nsoranzo: s/invalid.xml_doi.xml/invalid_doi.xml/ ?

jmchilton:
Awesome thanks!

planemo/lint.py
```diff
@@ -7,8 +8,42 @@
...
+    """
+    Check if dx.doi knows about the publication_id
+    """
+    base_url = "http://dx.doi.org"
+    doiless_publication_id = publication_id.split("doi:", 1)[-1]
+    url = "%s/%s" % (base_url, doiless_publication_id)
+    r = requests.get(url)
+    if r.status_code == 200:
+        if publication_id != doiless_publication_id:
+            lint_ctx.error("%s is valid, but Galaxy expects DOI without 'doi:' prefix" % publication_id)
```
nsoranzo: Are you sure about this? @bgruening pointed out in https://github.com/galaxyproject/tools-iuc/pull/768#issuecomment-221633995 that the initial 'doi:' seems to work fine.

mvdbeek: I just tested it again, the doi breaks the display of the reference with my terribly out-of-date planemo instance (September 2015 I believe), but newer galaxy can deal with the `doi:`.  I guess we could tone this down to an info?

nsoranzo: Something like:
```python
lint_ctx.warn("%s is valid, but old Galaxy versions expected DOI without 'doi:' prefix" % publication_id)
```

?

mvdbeek: Yes ... but given those old instances will disappear over time, maybe we can remove the warning altogether? Alternatively we should at least figure out how old is `old` :/

nsoranzo: Up to you or @jmchilton, I don't have a strong opinion on this.

jmchilton: I have no clue what changed to enable those - I just rewalked through the code and I can't even figure out why it works at all with the leading doi: - I don't think it should. Anyway - I definitely consider a leading ``doi:`` a tool bug and on a purely stylistic point somewhat redundant given the citation type is "doi". I think ``warn`` is entirely appropriate here.

PR #496, opened 2016-06-16
Use ephemeris to handle shed tools for workflows.

jmchilton:


docs/presentations/2016_workflows.html
```diff
@@ -14,18 +16,26 @@
...
-# Planemo: A Scientific Workflow SDK
+class: title
+# Planemo
+## ~~A Galaxy Tool SDK~~
+## A Scientific Workflow SDK
 John Chilton, Aysam Guerler, and the Galaxy Team
 
+???
+
+Oh look, I get to follow Björn again. That guy ruins the curve - when
```
martenson: I had to google for a bit and still do not understand the phrase 'ruin the curve' correctly I am afraid. Non-native speakers might have problems catching this.

nsoranzo: This is a reference to https://en.wikipedia.org/wiki/Grading_on_a_curve , so "to ruin the curve" is a sarcastic way of saying that somebody did something that other people will have a hard time competing with.

jmchilton: Okay - I'll try to come up with another way to say that for you freedom haters. Thanks!

peterjc: Now you're using other Americanisms just for fun, aren't you?

martenson: @nsoranzo Americanism isn't an actual thing - let's just call it post-medieval Europeanism.

PR #498, opened 2016-06-22
[WIP] revamp the docs

martenson:
up to advanced dev topics, working on that now


docs/installation.rst
```diff
@@ -2,34 +2,42 @@
 Installation
 ============
 
-The recommended approach for installing Planemo is to use Homebrew_ or
-linuxbrew_. To install planemo this way simply use the ``brew`` command as
-follows.
+The recommended approach for installing Planemo is to use Conda_
```
martenson: I think @jmchilton was against this switch, I will revert.

jmchilton:
Very nice so far! Thanks a million.

docs/installation.rst
jmchilton: Probably yes - That conda recipe doesn't get updated nearly as frequently as the Homebrew or PyPI recipe yet. I'm still nervous about using conda environments with Galaxy also - though I guess it should be fine.

I'll try to adjust my workflow with the GCC as impetus and hopefully we can reach a point where that conda recipe is my go to one.

martenson:
We are back to Homebrew for planemo itself (but have Conda for seqtk and BWA)

martenson:
The whole advanced tooldev topics section is now validated and concise. 

The only thing I am unclear now is whether we should add another tutorial introducing devs to Conda-controlled requirements?

martenson:
fixed conflicts

jmchilton:
I'm walking through the new docs on the new image and so far it is simply fantastic - this is a great improvement and I profoundly appreciate the effort you put into this! I can't thank you enough @martenson!

PR #517, opened 2016-07-06
Add the Constructive Solid Geometry category

gregvonkuster:


nsoranzo:
I guess we should wait for the category to be added to MTS and TTS before merging.

blankenberg:
Category existed in MTS when I checked, but I just added it to TTS.

gregvonkuster:
Thanks @blankenberg

PR #521, opened 2016-07-22
Update pip and setuptools on TravisCI; fix travis_init

### LINKED ISSUE #520: pip install planemo fail
You should be already aware of that!
I observe that within my travis test and directly on my server:

```
pip install planemo

[...]

Collecting html5lib (from rdflib>=4.1.0->cwltool->planemo)
  Downloading html5lib-0.999999999.tar.gz (245kB)
    100% |████████████████████████████████| 245kB 673kB/s
    Complete output from command python setup.py egg_info:
    html5lib requires setuptools version 18.5 or above; please upgrade before installing (you have 15.0)

    ----------------------------------------
    Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-UfHR2m/html5lib
```

peterjc:
Fixes #520 by explicitly updating pip and setuptools before installing planemo.

Also fixes a regression I introduced recently in 364fec2a3b5458e49ce3f96b76413627243c049c (typo in function name)

### DIFF
--- a/planemo/commands/cmd_travis_init.py
+++ b/planemo/commands/cmd_travis_init.py
@@ -46,6 +46,7 @@ TRAVIS_SETUP = """#!/bin/bash
 # TODO: Add your instructions here:
 """
 
+
 @click.command('travis_init')
 @options.optional_project_arg()
 @command_function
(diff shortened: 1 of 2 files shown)


nsoranzo:
Upgrading `setuptools` should be enough, we had to do the same for `tools-iuc` TravisCI: https://github.com/galaxyproject/tools-iuc/pull/887

peterjc:
Possibly updating setuptools only is enough right now, but I've been bitten by an old pip before with other TravisCI setups, so this seems a sensible precautionary measure.

bgruening:
CWL errors again. I think it's fine to merge this.
Thanks @peterjc!

PR #527, opened 2016-08-03
Pin older cwltool version - newer one breaks Galaxy.

jmchilton:


nsoranzo:
Should we also pin `html5lib` for https://github.com/galaxyproject/planemo/issues/520 ? It has been fixed in GitHub, but not officially released on PyPI.

PR #528, opened 2016-08-05
Include shed_lint in script run by travis_init

### LINKED ISSUE #514: planemo travis_init should use planemo lint
The ``planemo travis_init`` command at its core will setup a TravisCI configuration which via ``.travis.yml`` will download and run https://github.com/galaxyproject/planemo/blob/master/scripts/travis_test.sh meaning any changes to the master script will immediately take effect on third party Galaxy tool testing.

Right now, apart from the Galaxy and Tool installation magic, the key command is ``planemo test``.

I am proposing that at the start of ``travis_test.sh`` we invoke ``planemo shed_lint`` with ``--fail_level error`` (rather than the more stringent default ``--fail_level warn`` which would likely cause many false positives) as already used in https://github.com/galaxyproject/tools-iuc/blob/master/.travis.yml and https://github.com/galaxyproject/tools-devteam/blob/master/.travis.yml

The lint command is quick, and will flush out problems which can be caught without wasting donated VM run-time on running the test suite.

peterjc:
Untested (since currently my Galaxy GitHub repositories are not setup via ``planemo travis_init``), but intended to resolve #514

### DIFF
--- a/scripts/travis_test.sh
+++ b/scripts/travis_test.sh
@@ -6,4 +6,5 @@ pip install --upgrade pip setuptools
 pip install planemo
 planemo travis_before_install # runs .travis/setup_custom_dependencies.bash
 . ${TRAVIS_BUILD_DIR}/.travis/env.sh # source environment created by planemo
+planemo shed_lint --fail_level error --tools --ensure_metadata --urls -r ${TRAVIS_BUILD_DIR}
 planemo test --install_galaxy --no_cache_galaxy ${TRAVIS_BUILD_DIR}


jmchilton:
Thanks @peterjc - I'm not sure anyone is really using this Travis stuff but if they do this seems like a very good default task.

peterjc:
Just FYI, I've tested the suggested command successfully via:

https://github.com/peterjc/galaxy_blast/commit/7cfa35fc525610fcfc4915477c309401909f0ad6

https://github.com/peterjc/galaxy_mira/commit/4e26a4e3f3d69adaab657cb614e56ebcf9968592

https://github.com/peterjc/pico_galaxy/commit/84ec0adafd10715ef465a85cca33388f194c3b50

jmchilton:
Awesome sauce @peterjc - glad to hear it!

Issue #404, opened 2016-01-22
planemo dependency_script can loose current directory


peterjc:
e.g. https://github.com/galaxyproject/tools-iuc/blob/master/packages/package_samtools_1_2/tool_dependencies.xml has:

```xml
...
                    <action type="shell_command">cd htslib-1.2.1 &amp;&amp; make bgzip tabix</action>
                    <action type="move_file">
                        <source>htslib-1.2.1/bgzip</source>
                        <destination>$INSTALL_DIR/bin</destination>
                    </action>
                    <action type="move_file">
                        <source>htslib-1.2.1/tabix</source>
                        <destination>$INSTALL_DIR/bin</destination>
                    </action>
...
```

When run via Galaxy, `` <action type="shell_command">cd htslib-1.2.1 &amp;&amp; make bgzip tabix</action>`` has no lasting change on the current directory.

Currently via ``planemo dependency_script`` use of  ``cd htslib-1.2.1`` is never reverted (so the later steps break).

Workaround:

```xml
 <action type="shell_command">cd htslib-1.2.1 &amp;&amp; make bgzip tabix &amp;&amp; cd ..</action>
```

Probably the simplest fix is to append ``cd $dep_install_tmp`` to the conversion of any  ``<action type="shell_command">`` step?

peterjc:
This might be a little more complex: It should probably remember the results of the ``change_directory`` action, and also the auto-unpacking and change of directory when downloading a tar ball.

peterjc:
Any thoughts on wrapping the ``shell_command`` actions with ``pushd`` / ``popd``, or simply putting the whole action into a call to bash which would be closer to what Galaxy does anyway for running the ``tool_dependencies.xml`` file?

peterjc:
I'll keep an eye on my TravisCI tool tests using ``planemo dependency_script`` once the next planemo release is out - just in case it flags are any side effects.

PR #537, opened 2016-08-17
Return to working directory after shell_command action

peterjc:
Closes issue #404 with the verbose approach of wrapping the ``shell_command`` with ``pushd .`` and ``popd``.


jmchilton:
Sorry for the transiently failing tests - I'm hoping #536 helps but if it doesn't I'll just remove those two tests.

jmchilton:
Also - thanks for the fix!

PR #539, opened 2016-08-17
Look for sha256sum checksums during shed_lint

peterjc:
Looks at the ``download_by_url`` and ``download_file`` actions only, missing sha256sum is a warning, bad sha256sum is an error (e.g. not 64 characters, or non-hex characters present).

A future ``--verify`` switch could download the file and confirm the checksum matches.

TODO: Test case with a bad sha256sum entry (e.g. extra char)

### DIFF
--- a/planemo/shed_lint.py
+++ b/planemo/shed_lint.py
@@ -70,6 +70,11 @@ def lint_repository(ctx, realized_repository, **kwds):
         path,
     )
     lint_ctx.lint(
+        "lint_tool_dependencies_sha256sum",
+        lint_tool_dependencies_sha256sum,
+        path,
+    )
+    lint_ctx.lint(
         "lint_tool_dependencies_actions",
         lint_tool_dependencies_actions,
         path,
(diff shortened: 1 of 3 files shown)


jmchilton:
The failing test case looks legitimate - I'm fine with this change but devteam and iuc repos are going to start failing Travis checks after we merge this right? @martenson, @nsoranzo, @bgruening, @erasche are we okay with that?

erasche:
:+1: I want to know if my downloads are failing hash checks.

martenson:
I did the exact same steps in my brain a @erasche did. 👍

bgruening:
:+1: for me, we will fix travis if this is the case. Nice change Peter!

peterjc:
Copy and paste error in the test cases, one of the checksums is the wrong one...

jmchilton:
Awesome sauce - thanks! This didn't make the 0.28.0 cut - but it will be 0.29.0. Is there urgency - I can do a release if so?

peterjc:
Thanks John - there's no rush from my personal perspective.

PR #538, opened 2016-08-17
Verify sha256sum checksums in planemo dependency_script

peterjc:
In addition this fixes a problem trying to run the script without setting ``$DOWNLOAD_CACHE`` (documented as defaulting to ``./download_cache``) or if the local cache folder did not exist.

PR #541, opened 2016-08-17
dependency_script - only verify checksum of new downloads

peterjc:
This is a refinement to #538 from earlier today, which has been working nicely but makes using ``planemo dependency_script`` painfully slow if you already have the files downloaded and cached because it will verify their checksums.

I'm wondering about going further and putting this under a ``--verify`` switch (like the commented out one proposed for ``planemo shed_lint``), since as used on my TravisCI setup the checksums are run twice (once via ``planemo dependency_script`` and again via ``bash dep_install.sh``).

### DIFF
--- a/planemo/shed2tap/base.py
+++ b/planemo/shed2tap/base.py
@@ -423,12 +423,14 @@ def _cache_download(url, filename, sha256sum=None):
             # Most likely server is down, could be bad URL in XML action:
             raise RuntimeError("Unable to download %s" % url)
 
-    if sha256sum:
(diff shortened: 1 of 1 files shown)


peterjc:
Updated the ``deb_install.sh`` output to also only run ``shasum -a 256 ...`` if the file has been freshly downloaded.

This should solve the potential overhead on my TravisCI usage - the files should be downloaded and the checksum will be verified once only.

jmchilton:
Awesome - thanks for the refinement. This seems like a good compromise.

PR #552, opened 2016-08-31
Update Conda support for recent Galaxy improvements.

jmchilton:
Rev to latest galaxy-lib so that Galaxy handling of conda in master branch matches planemo's. Update planemo for recent changes - including always setting a condarc file up when using planemo - this is needed to fix channel handling for galaxy-lib update which sets up a $HOME for every conda command. Also update the list of default channels and improve verbose logging around these things.

### DIFF
--- a/planemo/commands/cmd_conda_install.py
+++ b/planemo/commands/cmd_conda_install.py
@@ -6,16 +6,35 @@ from galaxy.tools.deps import conda_util
 from planemo import options
 from planemo.cli import command_function
 from planemo.conda import build_conda_context, collect_conda_targets
-from planemo.io import coalesce_return_codes
+from planemo.exit_codes import EXIT_CODE_FAILED_DEPENDENCIES, ExitCodeException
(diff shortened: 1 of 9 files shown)


planemo/commands/cmd_conda_install.py
```diff
@@ -6,16 +6,34 @@
...
+            if conda_context.can_install_conda():
+                if conda_util.install_conda(conda_context):
+                    error("Attempted to install conda and failed.")
+                else:
+                    failed = False
+            else:
+                error("Cannot install conda, failing conda_install.")
+        else:
+            error("Conda not configured - run planemo conda_init' or pass --conda_auto_init to continue.")
+    if failed:
```
mvdbeek: failed isn't necessarily defined here.

jmchilton: Thanks! It wasn't at the right level - I think I have a fix for it now.

mvdbeek:
👍

PR #565, opened 2016-09-16
Do not fail URL linting in case of too many requests

nsoranzo:


jmchilton:
Nice fix - thanks a bunch!

Issue #567, opened 2016-09-18
`--xsd` should be the default and `--disable-xsd` an option


bgruening:
I think the new xsd option should be the default behavior. @jmchilton any reason not to do so?

jmchilton:
I think we would want to make lxml a required dependency then, which I was hesistant to do before because it required some libraries that weren't commonly installed on base systems but just the other day I did a pip install of it from pypi and got a binary manylinux wheel! It was very painless - so I'm up for this I think.

nsoranzo:
Or even always enabled?

bgruening:
;) I don't think there is any harm in offering a `disable` option for this, e..g if someone ships its own modifications to Galaxy. But the important part is to make it default :)
We can also leave out the disable option!

peterjc:
👍  on making XSD validation the default for linting tool XML files.

peterjc:
Enabling this has flagged all sorts of little things, proving very helpful:

 - https://github.com/peterjc/pico_galaxy/commit/ca5ee85c7efe402d05c02c42a570ce191006ab35
 - https://github.com/peterjc/pico_galaxy/commit/26c0b046e804d0525c4a7aa47b94a3869410d704
 - https://github.com/peterjc/galaxy_blast/commit/f68070f140c7a4811ac3a6e08e62e9b540e7b02a
 - https://github.com/peterjc/galaxy_blast/commit/b31248bf39f97caf0c01a9471dc71abfbf909cf5

peterjc:
Another common error in my tools, using ``required=...`` rather than ``optional=..``,

 - https://github.com/peterjc/pico_galaxy/commit/f7dd42e4c5c0c5d8ec5d0bcbf1e9ff52a8270865
 - https://github.com/peterjc/galaxy_mira/commit/0b9ce339ce9a591da30603cde0fb5f993573a157

peterjc:
There are a number of failures on https://github.com/galaxyproject/tools-iuc with the XSD validation, I will start filing issues...

jmchilton:
@peterjc There is a WIP PR that I think covers all of them https://github.com/galaxyproject/tools-iuc/pull/939. Might be better just to break it up and get some traction on individual parts though.

Issue #574, opened 2016-09-22
Tool XSD missing the version_command element interpreter tag


peterjc:
From https://travis-ci.org/peterjc/galaxy_blast/jobs/161948618 where I am trying linting my tools with ``--xsd`` which has flagged a number of minor errors on my part, this looks like a false positive:

```
.. ERROR: Invalid tmpbugrje found. Errors [/tmp/tmpbugrje:8: element version_command: Schemas validity error : Element 'version_command', attribute 'interpreter': The attribute 'interpreter' is not allowed.
```

Just like the ``command`` element has a (deprecated) ``interpreter`` tag, so too does the ``version_command`` element have an ``interpreter`` tag.

peterjc:
Fixed via #576

PR #576, opened 2016-09-22
Various XSD Changes

jmchilton:
See individual commits.


bgruening:
:+1: nice - Thanks!

peterjc:
So this should fix #567 and #574 - sounds good :)

peterjc:
I'm trying this on my tools today - there are some XSD failures which I need to look at to decide if they are my problems (real errors) or oversights in the XSD (false positives).

jmchilton:
Awesome - thanks for all the help @peterjc.

PR #598, opened 2016-10-31
Try to make planemo work with docker_galaxy

bgruening:


bgruening:
Not that this PR will make the `engine` work, but at least it fixes some linting errors. I still get wired CWL error trying to run `engine`.

jmchilton:
Thanks @bgruening - I'll try to find time to have a look.

PR #635, opened 2017-02-20
Implement a --no_dependency_resolution option.

### LINKED ISSUE #633: Disabling conda dependency resolver
So it seems that's currently not possible. The closest thing to disabling
conda would be providing you're own job_conf.xml.
A flag like `--no_conda_dependency_resolution` analogous to `--no_conda_auto_install` would be nice. 
Usecase is if you're developing a python package, and you would like to test it with planemo before packing it up in pypi followed by creating a conda package from pypi. 
You can also work around it by removing the conda package or the requirements tag, but it would still be nice if planemo could be used in that scenario.

jmchilton:
As requested by @mvdbeek https://github.com/galaxyproject/planemo/issues/633. Prior to the latest release of Planemo Conda would only be enabled with --conda_dependency_resolution despite being on by default in Galaxy.

Fixes #633.

Ping @mvdbeek - I haven't tested it yet - but I wanted to at least take a quick pass. I'll try to come back to this and merge it soon unless you want to test it and merge it first.

### DIFF
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -102,6 +102,9 @@ DEFAULT_DEPENDENCY_RESOLUTION_CONF = """<dependency_resolvers>
 </dependency_resolvers>
 """
 
+NO_DEPENDENCY_RESOLUTION_CONF = """<dependency_resolvers>
+</dependency_resolvers>
+"""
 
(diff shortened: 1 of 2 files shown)


mvdbeek:
Works like a charm, thanks a lot @jmchilton!

jmchilton:
Great to hear, thanks for the merge!

PR #650, opened 2017-03-02
First crack at containers documentation and related usability improvements

jmchilton:
- Implement ``planemo lint --biocontainer`` to mirror ``planemo lint --conda_requirements``. Just checks that a tool is ready to go with best practice namespaces for BioContainers tools.
- Implement test case for new linting flag.
- Add new section of documentation to advanced tool topics for containers. Work through examples of both an existing working container (seqtk) and one that must be built on the fly. Demonstrate ``planemo mull``.
- Hack around a bug in Conda 4.2 that makes it so ``planemo mull`` doesn't work out of the box on Mac OS X.
- Add more options and more documentation to the ``planemo mull`` command.

xref https://github.com/galaxyproject/training-material/issues/232

### DIFF
--- a/docs/_writing_dependencies_conda.rst
+++ b/docs/_writing_dependencies_conda.rst
@@ -1,3 +1,5 @@
+.. _dependencies_and_conda:
+
 Dependencies and Conda
 ===========================================
 
--- a/docs/_writing_dependencies_docker.rst
+++ b/docs/_writing_dependencies_docker.rst
@@ -1,3 +1,322 @@
 Dependencies and Docker
 ===========================================
 
+For years Galaxy has supported running tools inside containers. The details
+of how to run Galaxy tools inside of containers varies depending on the
+Galaxy job runner but details can be found in Galaxy's job_conf.xml sample file.
+
+This document doesn't describe how to run the containers, it describes how Galaxy
+figures out which container to run for a given tool. There are currently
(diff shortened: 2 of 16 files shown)


bgruening:
Reads very well John! Wondering if this not also should be part of the Admin Docs in Galaxy.

jmchilton:
There should be corresponding admin docs in Galaxy discussing how to use them from a deployers perspective for sure. Those docs shouldn't really mention Planemo. This is developer-centric documentation that covers Planemo and doesn't even mention ``enable_beta_mulled_containers`` or ``containers_resolvers_config_file`` - the Galaxy options.

I don't think the Galaxy side is quite ready for detailed documentation like this - I think we need some GUI tools, scripting, and a broad set of tools that have containers known to work first.

bgruening:
> I don't think the Galaxy side is quite ready for detailed documentation like this - I think we need some GUI tools, scripting, and a broad set of tools that have containers known to work first.

Might, can be. But the first part is also interesting for admins that want to know how these things work internally. But yes, this has time. just wanted to say. this looks really good!!!

jmchilton:
Yeah - I appreciate the review and the note about documentation. I have created a Galaxy issue for tracking this here https://github.com/galaxyproject/galaxy/issues/3710. Thanks as always and have . a great weekend.

I'm closing this PR because I merged directly after fixing the documentation problems.

PR #662, opened 2017-04-12
Create and use a temporary shed_data_manager_conf.xml

nsoranzo:
Otherwise installing a data manager from the Tool Shed under `planemo serve` adds it to the `shed_data_manager_conf.xml` file of the Galaxy repo.

Also:
- Fix Python3 compatibility
- Remove duplicated statements


jmchilton:
Wonderful - thanks so much!

PR #665, opened 2017-04-27
Fix filter_paths() to not partial match paths

nsoranzo:
E.g. in https://github.com/galaxyproject/tools-iuc/pull/1278 having 'tools/taxonomy' in `.tt_blacklist` and executing `planemo ci_find_repos --exclude_from .tt_blacklist` wrongly filters out also 'tools/taxonomy_krona_chart'.

### DIFF
--- a/planemo/io.py
+++ b/planemo/io.py
@@ -311,7 +311,17 @@ def filter_paths(paths, cwd=None, **kwds):
         return os.path.normpath(path)
 
     def exclude_func(exclude_path):
-        return lambda p: norm(p).startswith(norm(exclude_path))
+        def path_startswith(p):
+            """Check that p starts with exclude_path and that the first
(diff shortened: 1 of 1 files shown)


jmchilton:
Sorry about that - thanks for the fix! I guess I should convert those to regex or globs perhaps.

nsoranzo:
Hopefully we'll remove all tools from the blacklist before that is needed :)

nsoranzo:
@jmchilton Can you do a release with this included? It's needed for https://github.com/galaxyproject/tools-iuc/pull/1278 .

jmchilton:
@nsoranzo 0.40.1 is now available on PyPI.

nsoranzo:
Thanks @jmchilton, https://github.com/galaxyproject/tools-iuc/pull/1278 is now fixed!

PR #672, opened 2017-05-18
Add empty_extra_files_path validator for Galaxy tools xsd

gregvonkuster:
Also update docs to include empty_dataset


gregvonkuster:
I guess this is no longer needed due to https://github.com/galaxyproject/planemo/pull/688.

Issue #644, opened 2017-02-24
Update tool_init to use detect_errors.


jmchilton:
This will require redoing the docs also but it is worth it - those stdio blocks are noisy and no longer a best practice.

jmchilton:
Done with #693.

Issue #502, opened 2016-06-27
workshop


bgruening:
- booleans as select parameter
- quote many text, data parameters

PR #674, opened 2017-05-20
small steps to remove stdio from the docs in favor of detect_errors

bgruening:
xref: https://github.com/galaxyproject/planemo/issues/644 and https://github.com/galaxyproject/planemo/issues/502


bgruening:
@jmchilton the failing test is because BWA returns error code of 1 while printing the help. How do you want me to solve this? Hacking in cheetah, like `| grep "bwa"` or reverting to stdio for these few tools?

jmchilton:
Yeah - I'd revert the changes for those few tools. It is fine to use stdio if an exit code of 1 doesn't indicate an error - at least in this context we expect that to be the exit code so the stdio if fine. The rest of this is super awesome though - thanks for going through and doing all of that!

One problem is that the line numbers of highlighting things in the RST need to be updated when you change the seqtk example tools and tool_init output. We can merge as is and then create an issue for that if you'd like and assign me - I need to rewalk through all the tutorials with a new copy of planemo machine anyway before the GCC - I can fix those as I go.

jmchilton:
I'm going to cherry-pick e0803b7be3c0fc699e3fc30f650f16a00b671ca6 and add it to my next docs update. Otherwise I redid the tool builder and seqtk changes in #693 - but without the quotes changes since tool_init doesn't do it that way (we should fix tool_init to though down the road) and I also changed the macro version of tool_init to not add a stdio block.

Hope this okay.

Issue #646, opened 2017-02-27
Mulled Support - Part II


jmchilton:
Continued from #585 and #583.

- [x] Implement linting of mulled container's existence.
- [x] Work through seqtk based example for documentation.
- [x] Support multi-container registry for biocontainers.
   - [x] Setup a composite recipes path in auto-mulled (e.g. `composite-recipes/`).
      - [x] Use `mulled-build-files` to monitor that directory as part of the Travis build process for that repository.
      - [x] Implement a check for existing published images in `mulled-build-files`.
    - [x] Implement `mulled_register` command to open a PR against `composite-recipes/` for a given tool or tools.
- [x] Redo mulled concept with a version 2 of the hash - that hashes just the package names in the repository name and then hashes the versions in image version followed by a build number.

jmchilton:
Mulled enhancements for doing this last piece are implemented in https://github.com/galaxyproject/galaxy-lib/pull/58 and https://github.com/galaxyproject/galaxy-lib/pull/59.

jmchilton:
Moved some of these TODO items to #695 - this remains the GCC 2017 target issue.

PR #675, opened 2017-05-22
Implement container_register for tool repositories.

jmchilton:
xref - last bullet of #646

### DIFF
--- a/planemo/commands/cmd_conda_install.py
+++ b/planemo/commands/cmd_conda_install.py
@@ -11,6 +11,7 @@ from planemo.io import coalesce_return_codes
 
 @click.command('conda_install')
 @options.optional_tools_or_packages_arg(multiple=True)
+@options.recursive_option()
 @options.conda_target_options()
 @options.conda_global_option()
 @options.conda_auto_init_option()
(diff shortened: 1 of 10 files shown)


peterjc:
Is it worth fixing the typo in the commit message?

peterjc:
The documentation is somewhat cryptic without being familiar with the area, but that also means I can't suggest how to improve this:

```python
   """Register multi-requirement containers as needed.

    BioContainers publishes all Bioconda packages automatically as individual
    container images. These however are not enough for tools with multiple
    best-practice requirements. Such requirements should be recorded and published
    so that a container can be created and registered for these tools.
    """
```

jmchilton:
@peterjc Agreed this is not documented well and I'm still trying to figure out how to describe what is going on. There was a bit of a conversation about this in the tools-iuc Gitter just now and there is some discussion on the README of [this repo](https://github.com/jmchilton/mulled-monitor) if you want to dig a little deeper. At a high level, the goal is that all tools that have Conda-ified requirement tags will automatically have containers registered and published quay.io/biocontainers for each unique combination of requirements. This new command opens a PR against a registry of such combinations that when merged will result in a container being published. I want to just monitor repositories though and take care of these details without intervention on a per tool basis - so to leverage this tool developers will just need to add their repository to the list in this file (https://github.com/jmchilton/mulled-monitor/blob/master/monitor.sh#L3). I'm going to present this as the GCC - so hopefully I figure out how to talk about this before then.

bgruening:
Wuhu!!!!

jmchilton:
Thanks @bgruening - I've updated #646 with your v2 mulled idea.

Issue #494, opened 2016-06-04
planemo shed_update should not fail if there is nothing to update


bgruening:
If the repository is already up-to-date the TS will return the following message:

```
Unexpected response from galaxy: 400: {"content_alert": "", "err_msg": "No changes to repository."}
```

This should not be an error as it breaks travis/jenkins integration.

nsoranzo:
:+1: This an issue especially when uploading a suite of tools where not all tools are updated.

PR #680, opened 2017-05-26
Return 0 and a warning if repo has no changes

nsoranzo:
Fix https://github.com/galaxyproject/planemo/issues/494 .

### DIFF
--- a/planemo/shed/__init__.py
+++ b/planemo/shed/__init__.py
@@ -15,6 +15,7 @@ from tempfile import (
     mkstemp,
 )
 
+import bioblend
 import six
 import yaml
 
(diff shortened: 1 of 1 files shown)


planemo/shed/__init__.py
```diff
@@ -312,6 +314,10 @@ def upload_repository(ctx, realized_repository, **kwds):
             str(repo_id), tar_path, **update_kwds
         )
     except Exception as e:
+        if isinstance(e, bioblend.ConnectionError) and e.status_code == 400 and \
+                e.body == '{"content_alert": "", "err_msg": "No changes to repository."}':
```
martenson: I would check *only* the `err_msg` here. Content alert has nothing to do with this I think.

nsoranzo: BioBlend returns this `ConnectionError` exception whose `body` attribute is the stringified version of the JSON returned by Galaxy. I preferred to be strict to not match other real errors, we can relax this if needed.

PR #693, opened 2017-06-18
Remove stdio from generated tools - just use exit_code for everything.

jmchilton:
Other small formatting tweaks - but less than #674 since I'm trying to keep these tools the way they are when tool_init generates them - at least for those options. This also doesn't strip stdio from other places like the bwa test tool that broke the tests in #674.

Cleans up docs pre-GCC.


mvdbeek:
Very nice!

jmchilton:
Thanks @mvdbeek!

PR #696, opened 2017-06-21
Update BWA-MEM examples in the TDD docs.

jmchilton:
I originally updated BWA-MEM for Conda in https://github.com/galaxyproject/planemo/commit/9af67bfd46509cc6d59ff6f2d86e2ddca613b2f9 but I didn't update the remaining doc examples.

Also adding diff script to see diffs between versions and cleaning up some of the content of the tutorial itself (including dropping a warning about an older error that has been fixed).

### DIFF
--- a/docs/writing/bwa-mem_v2.xml
+++ b/docs/writing/bwa-mem_v2.xml
@@ -2,8 +2,8 @@
 <tool id="bwa_mem_test" name="Map with BWA-MEM" version="0.0.1">
     <description>- map medium and long reads</description>
     <requirements>
-        <requirement type="package" version="0.7.7">bwa</requirement>
-        <requirement type="package" version="0.1.16">samtools</requirement>
+        <requirement type="package" version="0.7.15">bwa</requirement>
+        <requirement type="package" version="1.3">samtools</requirement>
     </requirements>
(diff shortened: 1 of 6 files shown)


docs/writing/bwa-mem_v2.xml
```diff
@@ -2,8 +2,8 @@
 <tool id="bwa_mem_test" name="Map with BWA-MEM" version="0.0.1">
     <description>- map medium and long reads</description>
     <requirements>
-        <requirement type="package" version="0.7.7">bwa</requirement>
-        <requirement type="package" version="0.1.16">samtools</requirement>
+        <requirement type="package" version="0.7.15">bwa</requirement>
+        <requirement type="package" version="1.3">samtools</requirement>
```
mvdbeek: we're using 1.3.1 in the devteam wrapper, in case you want to keep this synced.

jmchilton:
Thanks a ton for the review @mvdbeek - I really appreciate the second set of eyes on this so close to GCC.

docs/writing/bwa-mem_v2.xml
mvdbeek: Either way conda picks the latest version that satisfies the request, so you'll end up with 1.3.1 regardless.

PR #697, opened 2017-06-21
Update bwa test data generation to allow old or new samtools to be used.

jmchilton:


jmchilton:
Thanks again @mvdbeek !

PR #704, opened 2017-06-24
handle None versions

bgruening:
Handles package definitions without a version number: https://github.com/galaxyproteomics/tools-galaxyp/blob/master/tools/peptideshaker/searchgui.xml#L10

might fix: https://github.com/BioContainers/multi-package-containers/pull/329
completely untested ...

### DIFF
--- a/planemo/commands/cmd_container_register.py
+++ b/planemo/commands/cmd_container_register.py
@@ -159,7 +159,14 @@ class RegistryTarget(object):
 
     def write_targets(self, ctx, target_filename, mulled_targets):
         with open(target_filename, "w") as f:
-            contents = ",".join(["%s=%s" % (t.package_name, t.version) for t in mulled_targets])
+            target_strings = list()
+            for target in mulled_targets:
+                if target.version:
+                    target_str = "%s=%s" % (target.package_name, target.version)
+                else:
+                    target_str = target.package_name
+                target_strings.append(target_str)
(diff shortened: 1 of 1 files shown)


planemo/commands/cmd_container_register.py
```diff
@@ -159,7 +159,14 @@ def handle_pull_request(self, ctx, name, target_filename, packages_str, tools_st
 
     def write_targets(self, ctx, target_filename, mulled_targets):
         with open(target_filename, "w") as f:
-            contents = ",".join(["%s=%s" % (t.package_name, t.version) for t in mulled_targets])
+            target_strings = list()
```
mvdbeek: how about `contents = ",".join(["%s=%s" % (t.package_name, t.version) if t.version else "%s" % t.package_name for t in mulled_targets])` ?

mvdbeek:
Is it a good idea to allow this though? Should we not better fix the tools?

planemo/commands/cmd_container_register.py
bgruening: I don't like these large list comprehensions, they are less readable and this section is not performance critical.

bgruening:
I don't think such things are a good idea. But Galaxy allows this so this should work here as well. However, I have on my todo list to enhance planemo lint and if this all is merged I will fix the galaxyp tools.

mvdbeek:
Sounds good then!

PR #705, opened 2017-06-28
writing_appliance script example: smiting perl from yet one more place

NickSto:
The example script in https://planemo.readthedocs.io/en/latest/writing_appliance.html#wrapping-a-script was fine, but it was written in Perl (gasp). Why not Python, I thought?

Full disclosure: I haven't tested it, since I the build environment hasn't worked yet on my machine. But it's just a change in existing docs, what could go wrong? (famous last words)


docs/writing/gc_content.py
```diff
@@ -0,0 +1,20 @@
...
+    for line_raw in infile:
+      line = line_raw.rstrip('\r\n')
+      if line.startswith('>'):
+        if name is not None:
+          outfile.write(name+': '+str(gc/float(total))+'\n')
+        name = line[1:]
+        gc = 0
+        total = 0
+      else:
+        gc += line.count('G') + line.count('C')
```
erasche: not case insensitive, unlike the perl version.

docs/writing/gc_content.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python2
```
erasche: python2? why not python3? ;)

docs/writing/gc_content.py
NickSto: True, true. Didn't know if it was more important to be complete or keep it simple. I can make the edit if you guys want.

docs/writing/gc_content.py
NickSto: Haha, I'd love to, but I figured let's not go crazy here.

NickSto: Actually, this made me do some research.. can OS X people confirm whether python2 is on your PATH?

peterjc: By default, no. At least only ``python`` is on the ``$PATH`` on macOS Sierra 10.12.5, not ``python2``.

```
$ uname
Darwin
$ which python2
$ python2
-bash: python2: command not found
$ which python
/usr/bin/python
$ python --version
Python 2.7.10
```

jmchilton:
Comments addressed in https://github.com/galaxyproject/planemo/pull/1709 - thanks for the fix and the review!

Issue #107, opened 2015-03-24
Implement shed_lint.


jmchilton:
Lint a shed repository instead of just tools. This is a meta-issue to track 4 issues open by @peterjc.
- [ ] Check sample files #96.
- [x] Check for a README #89.
- [x] Check rst in any `.rst` files #91.
- [x] Lint `tool_dependencies.xml`.
- [x] ~Cross reference `tool_dependencies.xml` with tools #90.~
- [x] Lint `.shed.yml`
- [x] Lint `repository_dependencies.xml`.
- [x] Check that packages belong only to the category `Tool Dependency Packages`. 8c2a45a14605e9695575ad1ce80d0791ba571d3d
- [x] Check tool dependency actions that are difficult to verify with XSD. 8117e038917f9c2914e7871d519e1af448cec54b
- [x] Check for required type based on repository type. #156
- [x] ~Check for changeset and tool shed attributes in `tool_dependencies.xml` files. (https://github.com/galaxyproject/tools-iuc/pull/149#issuecomment-104311715)~

erasche:
How would you feel about checking for .shed.yml? The, admittedly, under debate, current example is: https://gist.github.com/erasche/25a2405f259f9d2fb14f

jmchilton:
Some initial work here https://github.com/galaxyproject/planemo/pull/130 addressing linting ``tool_dependencies.xml``, ``repository_dependencies.xml`` and ``.shed.yml`` but not the rest of this - but there is at least a spot for it now.

peterjc:
@erasche did you mean finding a ``.shed.yml`` means the folder is a Tool Shed repository? Or when checking a folder thought to be a Tool Shed repository, it ought to have a ``.shed.yml``? This is important for recursion - how should ``planemo shed_list -r ...`` decide what to look at?

erasche:
@peterjc nmo, I meant checking that a `.shed.yml` file was present in what looks like a tool folder/maybe validating the contents of it. See #130 and "Bare minimum to lint .shed.yml files."

peterjc:
Thanks for the clarification @erasche - basic linting of ``.shed.yml`` is happening now (ticked on the TODO list here), but I agree that can probably be more through.

See also #156 for dependency packages.

erasche:
#96 would've caught https://github.com/galaxyproject/tools-iuc/issues/148

erasche:
```
[...]

PR #708, opened 2017-07-04
Elevate missing .shed.yml to error

erasche:
From https://github.com/galaxyproject/planemo/issues/107#issuecomment-312948255

This change will probably cause repositories which passed linting before, to fail now, if people were passing directories which weren't things which should be uploaded to the tool shed. Is that acceptable, or should this be hidden behind a strictness falg or something?

### DIFF
--- a/planemo/shed_lint.py
+++ b/planemo/shed_lint.py
@@ -289,7 +289,7 @@ def lint_shed_yaml(realized_repository, lint_ctx):
     path = realized_repository.real_path
     shed_yaml = os.path.join(path, ".shed.yml")
     if not os.path.exists(shed_yaml):
-        lint_ctx.info("No .shed.yml file found, skipping.")
+        lint_ctx.error("No .shed.yml file found.")
         return
     try:
         yaml.load(open(shed_yaml, "r"))


peterjc:
How will this interact with recursion and different levels of testing for tools? e.g. ``tools/tool_a/`` vs ``tools/suite_b/tool_c/`` where not all the folders should have a ``.shed.yml`` file.

erasche:
http://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/repositories.html#github-repositories according to this, suites should be in a separate directory, which sidesteps the problem ;)

But yes, you raise a good point. I don't have a good answer for this, and I'm open to ideas. I just believe that a missing .shed.yml is super important.

peterjc:
I agree this is important, and as long as it doesn't break existing widespread folder structures, its a very good thing. Planemo already has some non ``.shed.yml`` file base tool detection, but I don't know exactly how that works.

jmchilton:
I was a bit confused by this at first - because I thought planemo defines a repository as a directory containing a ``.shed.yml`` file - and for the most part it does. But the shed operations it looks like will just use the supplied path as is if it doesn't find any ``.shed.yml`` files.

```
def _find_raw_repositories(ctx, path, **kwds):
    name = kwds.get("name", None)
    recursive = kwds.get("recursive", False)

    shed_file_dirs = find_matching_directories(
        path, SHED_CONFIG_NAME, recursive=recursive
    )

    config_name = None
    if len(shed_file_dirs) == 1:
        shed_file_dir = shed_file_dirs[0]
        try:
            config = shed_repo_config(ctx, shed_file_dir, name=name)
        except Exception as e:
            error_message = PARSING_PROBLEM % (shed_file_dir, e)
            exception = RuntimeError(error_message)
            _handle_realization_error(exception, **kwds)
            return [exception]
        config_name = config.get("name", None)

    if len(shed_file_dirs) > 1 and name is not None:
        raise Exception(NAME_INVALID_MESSAGE)
    if config_name is not None and name is not None:
        if config_name != name:
            raise Exception(CONFLICTING_NAMES_MESSAGE)
    raw_dirs = shed_file_dirs or [path]
```

This behavior is because of the last line here. I really was trying to take a very light touch with .shed.yml files originally and everything in there truly was optional - non-optional fields could be specified at the command-line for instance.

With that in mind, two things.
- I don't think a missing .shed.yml would be caught by tools-iuc's Travis setup with this change since the jq directory wouldn't have been picked up as a directory/repository to lint. So I'm not sure this helps https://github.com/galaxyproject/tools-iuc/pull/1402. Maybe I'm wrong though?
- Following the linting levels for tools - in my view of things this isn't an error, it is a warning - the repository can still be used - it isn't broken per se. Error the way I've tried to use it (and maybe I've failed to be consistent at times) means usable, completely broken - not just in poor form.
- If we change this from error to warn, the check becomes a much weaker check than what is produced with ``--ensure_metadata`` and that makes sense I think - the whole point of ``.shed.yml`` is provide some default metadata properties.

Therefore - I think the existing INFO or WARN would be correct - the problem is that you are linting in such a way that you are explicitly ignoring warnings. I think it would be better to just fail on warnings as is the planemo default. I've opened a tools-iuc PR here https://github.com/galaxyproject/tools-iuc/pull/1407 that would cause metadata problems with repositories to fail the build while still "ignoring" warnings in the tools themselves - I think that is better right?

PR #711, opened 2017-07-21
Replace double quotes in seqtk example 

bebatut:
Spotted during the ISMB workshop


jmchilton:
But tool_init doesn't produce data this way right? I think we shouldn't make this change until tool_init produces this data right?

nsoranzo:
Ops, I suppose you're right, revert? Or fixing tool_init is easy enough?

PR #716, opened 2017-08-14
Fix use of conda_context

nsoranzo:
In particular, the changes to `best_practice_search()` in `planemo/conda.py` are needed to support the changes to `best_search_result()` in https://github.com/galaxyproject/galaxy/pull/4409 .


jmchilton:
The way I'm reading this I don't need a galaxy-lib update right - this should work as is?

nsoranzo:
Yes, this will support the future changes in galaxy-lib.

PR #720, opened 2017-09-06
Fix ruamel.yaml version

### LINKED ISSUE #719: dependency conflicts cause scripts/test_conformance_cwl.bash to fail
```
error: ruamel.yaml 0.15.31 is installed but ruamel.yaml<0.15,>=0.12.4 is required by set(['schema-salad'])

```
from https://ci.commonwl.org/job/galaxy-planemo-conformance/82/console

mvdbeek:
This should fix #719.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,3 +16,4 @@ ephemeris>=0.2.0
 galaxy-lib>=17.9.0
 html5lib>=0.9999999,!=0.99999999,!=0.999999999,!=1.0b10,!=1.0b09 ; python_version == '2.7'
 cwltool==1.0.20170224141733 ; python_version == '2.7'
+ruamel.yaml<0.15,>=0.12.4


jmchilton:
Cool - thanks @mvdbeek - I'll work on getting a release out with this in it.

nsoranzo:
Shouldn't that be fixed upstream in cwltool?

jmchilton:
@nsoranzo I imagine it was but we have a fixed older version of cwltool as a dependency, I'll remove this line when I merge my cwltool update as part of #714. Does that sound like the right approach?

jmchilton:
xref https://github.com/common-workflow-language/cwltool/pull/450

nsoranzo:
Perfect, thanks @jmchilton!

jmchilton:
@nsoranzo Cool - thanks! I've made that change in #714 and rebased it as https://github.com/galaxyproject/planemo/pull/714/commits/d2ea830458ba0d54c910a931602638033b44e2eb.

I'm going to see what the tests say about #723 and if things look okay I think I will do a release so this is fixed and Conda is fixed as well. Then I will merge #714 later this week (or maybe later today) and create a new set of problems.

jmchilton:
Well so things went sideways because the new galaxy-lib introduced test errors (these can be seen in #723 but would be experienced in the wild anyway since we don't pin an upper bound on galaxy-lib). So I just cleaned up and merged my CWL WIP with https://github.com/galaxyproject/planemo/commit/2462099b7d3536cdb91761ddf9da246922085daf and did a release so either the dependencies are fixed when coming from PyPI ... hopefully ... but probably not and we shall see.

nsoranzo:
@jmchilton I couldn't see any new test error in https://travis-ci.org/galaxyproject/planemo/jobs/272496388:
- jobloaderctx is caused by using an old cwltool, is fixed by commit 2462099b7d3536cdb91761ddf9da246922085daf
- fleeqtk not found: I'm sure I've seen this in previous Travis builds
- `AttributeError: 'module' object has no attribute 'numproc'`: should be fixed by https://github.com/galaxyproject/galaxy/pull/4565

Am I missing something?

jmchilton:
I just hadn't seen 2462099 previously - my :eyes: aren't 🦅 (they call me :poop::eyes:). I didn't think it would have occurred with older galaxy-libs before the two I released today.

nsoranzo:
jobloaderctx error is a side-effect of pinning ruamel.yaml in this very PR, in fact it started with https://travis-ci.org/galaxyproject/planemo/jobs/272481174

PR #725, opened 2017-09-12
Add 'Accept' header when lint doc URLs

nsoranzo:
Fix linting of http://www.microrna.org/


jmchilton:
Thanks! I'm thinking I'll release a new planemo today.

PR #727, opened 2017-09-15
Fix `--conda_auto_install` help

nsoranzo:
And rebuild documentation.


jmchilton:
This breaks the build for the linting of documentation - but I think all the new linted doc errors it found are valid things that need to be fixed. Thanks for the contribution!

jmchilton:
Doc lint fixes in #731.

PR #731, opened 2017-09-15
Doc lint fixes.

jmchilton:


docs/appliance.rst
```diff
@@ -257,6 +257,7 @@ environments such as Amazon Web Services and Google Compute Engine.
 
 .. _Galaxy: http://galaxyproject.org/
 .. _Docker: https://www.docker.com/
+.. _Conda: http://conda.pydata.org/
```
nsoranzo: It's https://conda.io/ nowadays.

planemo/network_util.py
```diff
@@ -33,10 +33,11 @@ def wait_http_service(url, timeout=None):
...
 def wait_net_service(server, port, timeout=None):
-    """ Wait for network service to appear
-        @param timeout: in seconds, if None or 0 wait forever
-        @return: True of False, if timeout is None may return only True or
-                 throw unhandled network exception
+    """ Wait for network service to appear.
+
+    @param timeout: in seconds, if None or 0 wait forever
+    @return: True of False, if ``timeout`` is ``None`` may return only True or
+    throw unhandled network exception
```
nsoranzo: Shouldn't that be indented?

jmchilton: I tried the old way, I tried two spaces, 4 spaces, Sphinx hates everything but this.

peterjc: Typo in the line above?

``True of False`` --> ``True or False``

nsoranzo: Shouldn't these start with a `:` http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists
```
:param timeout: ...
:return: ...
```

jmchilton: Revised the docs to implement this I think.

PR #738, opened 2017-09-26
Catch also socket.error when linting URLs

nsoranzo:
Noticed this today that SourceForge is down, see https://travis-ci.org/galaxyproject/tools-iuc/jobs/280065650


jmchilton:
Great - thanks!

Issue #763, opened 2018-02-05
planemo tool_init --help_from_command bytes under Python 3


peterjc:
Example using planemo installed under conda using Python 3.6.3:

```
$ planemo --version
planemo, version 0.47.0
$ planemo tool_init -i head -n head --help_from_command "head --help"
Tool written to head.xml
$ more head.xml 
<tool id="head" name="head" version="0.1.0">
    <requirements>
    </requirements>
    <command detect_errors="exit_code"><![CDATA[
        TODO: Fill in command template.
    ]]></command>
    <inputs>
    </inputs>
    <outputs>
    </outputs>
    <help><![CDATA[
        b"Usage: head [OPTION]... [FILE]...\nPrint the first 10 lines of each FILE to standard output.\nWith more t
han one FILE, precede each with a header giving the file name.\nWith no FILE, or when FILE is -, read standard inpu
t.\n\nMandatory arguments to long options are mandatory for short options too.\n  -c, --bytes=[-]K         print th
e first K bytes of each file;\n                             with the leading '-', print all but the last\n         
                    K bytes of each file\n  -n, --lines=[-]K         print the first K lines instead of the first 1
0;\n                             with the leading '-', print all but the last\n                             K lines
 of each file\n  -q, --quiet, --silent    never print headers giving file names\n  -v, --verbose            always 
print headers giving file names\n      --help     display this help and exit\n      --version  output version infor
mation and exit\n\nK may have a multiplier suffix:\nb 512, kB 1000, K 1024, MB 1000*1000, M 1024*1024,\nGB 1000*100
0*1000, G 1024*1024*1024, and so on for T, P, E, Z, Y.\n\nGNU coreutils online help: <http://www.gnu.org/software/c
oreutils/>\nFor complete documentation, run: info coreutils 'head invocation'\n"
    ]]></help>
    <citations>
    </citations>
</tool>
```

Notice the help text was captured as a bytes string, and presented as such in the XML output (``b"Usage: ...``)

I have not delved into the code base to see how other bytes-string/unicode issues have been dealt with in planemo for stdout/stderr from command line tools.

peterjc:
Pull request soon, just doing local testing of a simple fix now...

peterjc:
Should be fixed in the next release via #764, closing.

PR #764, opened 2018-02-05
py3: subprocess with universal_newlines=True

peterjc:
This is a simple way to ensure we get a string back (and not bytes) under Python 3, intended to solve #763 where the output is used directly in XML output.


jmchilton:
Awesome - thanks for the fix @peterjc !

PR #767, opened 2018-02-22
Fix workflow testing test & port picking for Galaxy serve.

jmchilton:


martenson:
You probably do not want to target master here.

jmchilton:
Merged directly after the green.

PR #770, opened 2018-02-22
Add a native Galaxy workflow (.ga) testing test.

jmchilton:


erasche:
Let us know and we'll be happy to test! We've got a ton of workflows we'd love to validate on our instance, especially for training.

jmchilton:
sqllite locking errors - ugh - not sure I can merge this. The next two things to do down this line of work are:

- Run all these workflow tests against postgres instead sqlite - #772.
- Add workflow example tests that use tool shed tools (I think the gxformat2 will install them but not the .ga files or the reverse of that - at any rates the guts are there we just need a simple test case to guide fixing up the interface for listing toolshed tools from a workflow).

bgruening:
Danke John!

PR #775, opened 2018-02-24
WIP: Testing for training assembly intro workflow 

jmchilton:
Update ephemeris usage for a modern version to work with both .ga and .yml workflows.

Had to modify the workflow slightly to define a workflow output for testing.

xref http://galaxyproject.github.io/training-material/topics/assembly/tutorials/general-introduction/tutorial.html

This seems to be falling apart on a Conda problem Planemo can't really solve? Anyone want to trigger a rebuild of that with a newer Conda for me?

```
DependencyException: Conda dependency failed to build job environment. This is most likely a limitation in conda. You can try to shorten the path to the job_working_directory.\n", "err_msg": "Uncaught exception in exposed API method:", "err_code": 0}
(2/3) Installing repository fastqc from devteam to section "Tools from workflows" at revision 9337dd1fbc66 (TRT: 0:00:34.882887)
	repository fastqc installed successfully (in 0:00:05.288841) at revision 9337dd1fbc66
(3/3) Installing repository fasta_stats from simon-gladman to section "Tools from workflows" at revision 20ca2574216a (TRT: 0:00:40.275837)
	repository fasta_stats installed successfully (in 0:00:03.969259) at revision 20ca2574216a
Installed repositories (2): [(u'fastqc', None), (u'fasta_stats', None)]
Skipped repositories (0): []
Errored repositories (1): [(u'velvetoptimiser', '')]
All repositories have been processed.
Total run time: 0:00:44.245543
```

Maybe I'll try again with a different workflow?

PR #776, opened 2018-02-24
Fix resolving & installing shed repositories from workflows.

jmchilton:
Used by run/serve/test of workflows.

- Update for modern ephemeris usage.
- Update to work with .ga files an not just explicit tool: sections in format 2 workflows.

Test case bundled into the not yet working #775. This part of that PR is working.


jmchilton:
Thanks for linting fix @bgruening !

Issue #100, opened 2015-03-17
Use of shell commands when alternatives are available


erasche:
Maybe supporting windows users isn't a priority (I have no need for it, despite submitting this), but there are a number of places where shell commands could/should be refactored into more portable shutil type commands.

E.g. [cmd_tool_init @ L208](https://github.com/galaxyproject/planemo/blob/ec6e30f8f7106bb749d13d8796145507e6c3d87d/planemo/commands/cmd_tool_init.py#L208), mkdir/cp/mv commands could be replaced.

jmchilton:
I know what you are saying would be more... professional - but I really like ``io.shell``. It prints the command in a pretty color so you can see what it is doing and how to do it yourself - it is more fun and a bit irreverent.

I will cede at least that ``tool_init`` and ``lint`` should absolutely work in Windows... ugh... you are right and I am not happy about it.

erasche:
Awww, yeah, I can definitely empathize with the pleasant visual feedback and interactivity of those. I might even go so far as to suggest you retain the extremely useful feedback/explanation of what planemo is doing by wrapping the specific copy/move/tar commands in a small class since there aren't so many unique commands that it'd be unthinkable to do so.

If nothing else, it's a very low priority issue until someone actually needs it. I only discovered this whilst reading the code.

jmchilton:
@nsoranzo is making good progress replacing fun, irreverent bugs with working code in https://github.com/galaxyproject/planemo/pull/774/commits/e09ac2f9ea283a164084f8f30b5a31ee98c2f280.

PR #785, opened 2018-02-27
Print fake shell commands.

jmchilton:
A bunch of raw shell scripting was replaced with Python system calls in 953958932384ff7ebd1a09806e6c290c556f9dea, that is great but the cool reporting of the shell commands was lost. This restores the reporting as fake shell commands corresponding to what is happening (sort of).

In the spirit of something @erasche proposed doing in #100.

### DIFF
--- a/planemo/io.py
+++ b/planemo/io.py
@@ -45,6 +45,26 @@ def communicate(cmds, **kwds):
     return output
 
 
+def copy(source, dest):
+    info("cp '%s' '%s'" % (source, dest))
+    shutil.copy(source, dest)
+
+
+def mkdir(path):
+    info("mkdir '%s'" % path)
+    os.mkdir(path)
+
+
+def makedirs(path):
+    info("mkdir -p '%s'" % path)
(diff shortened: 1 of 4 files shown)


planemo/io.py
```diff
@@ -45,6 +45,26 @@ def communicate(cmds, **kwds):
...
+
+
+def mkdir(path):
+    info("mkdir '%s'" % path)
+    os.mkdir(path)
+
+
+def makedirs(path):
+    info("mkdir -p '%s'" % path)
+    os.makedirs(path)
```
nsoranzo: A slight difference with between `mkdir -p` and `os.makedirs()` is that the former succeeds even if the directory already exists, while the latter fails. You may want to guard this with an `if not os.path.exists(path):`

nsoranzo:
If you still want this, I'd prefer human-friendly messages to fake shell commands.

jmchilton:
Planemo has matured and vision of it being nerdy has failed - just going to close this out.

PR #787, opened 2018-02-28
Revise README and touch up documentation in general.

jmchilton:
- Drop mention of Brew, brew-based tool development is probably a dead end at this point.
- Drop mention of the Travis support - per tool repositories never took off in favor of mega-repositories. Travis support should be upgraded to reflect this and not referenced until it is.
- Include less dated information in the readme and link out to the more well maintained RTDs for Conda and Docker.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,7 +34,7 @@ This quick start demonstrates using ``planemo`` commands to help
 develop Galaxy tools.
 
 -----------------
-Obtaining Planemo
+Obtaining
 -----------------
 
 For a traditional Python installation of Planemo, first set up a virtualenv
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -59,11 +59,11 @@ You can choose from open virtualization format (OVA_, .ova), Docker_,
 or Vagrant_ appliances.
 
 --------------
-Planemo Basics
+Basics
```
peterjc: Should the section over/underlines be edited to match (six characters long)

jmchilton: I don't think it matters? The older TravisCI section went over also. Is there a best practice around this?

Issue #788, opened 2018-03-02
Planemo not ready for Galaxy 18.01


jmchilton:
Planemo serve with 17.09 builds this paster:

```
python ./scripts/paster.py serve /private/var/folders/78/zxz5mz4d0jn53xf0l06j7ppc0000gp/T/tmpdabxUz/galaxy.ini --server-name=planemo28411 --pid-file=planemo28411.pid --log-file=planemo28411.log --daemon
```

With 18.01 it builds this:

```
python ./scripts/paster.py serve /private/var/folders/78/zxz5mz4d0jn53xf0l06j7ppc0000gp/T/tmpQFqHKF/galaxy.ini --server-name=planemo87041 --pid-file=planemo87041.pid --log-file=planemo87041.log --pid-file galaxy.pid --log-file galaxy.log --daemon
```

And can't find the log file.

jmchilton:
xref https://github.com/galaxyproject/galaxy/commit/09feaf6f4d46509c7401970013b5a8097a5d4c4e#r27076912

PR #789, opened 2018-03-02
Fix Planemo serve for Galaxy 18.01.

jmchilton:
Fixes https://github.com/galaxyproject/planemo/issues/788


jmchilton:
Merged as part of #790

PR #795, opened 2018-03-08
Restructure serve testing code for reuse.

jmchilton:
Want to build test cases in the external engine code that standup Galaxy with ``planemo serve`` and then test against it. This is an atomic step in that direction and that branch has grown unwieldy.


jmchilton:
Tests failing because #797 I think, probably unrelated to this commit.

PR #806, opened 2018-04-24
Sync galaxy.xsd with latest Galaxy updates

nsoranzo:
Fix https://travis-ci.org/TGAC/earlham-galaxytools/builds/370515703


nsoranzo:
Tests are failing because the ToolShed is down.

PR #808, opened 2018-04-29
catch link redirections

bgruening:


planemo/lint.py
```diff
@@ -148,6 +148,9 @@ def validate_url(url, lint_ctx, user_agent=None):
             if e.code == 429:
                 # too many requests
                 pass
+            elif e.code >= 300 and e.code < 400:
```
nsoranzo: Maybe that's a bit too forgiving, I'll see if we can solve the issue with nature.com without allowing other broken redirects.

nsoranzo: The above code works, BUT `requests` is HTTP-only, so FTP URLs fail with something like:
```
Applying linter tool_urls... FAIL
.. ERROR: Error 'No connection adapters were found for 'ftp://ftp.ncbi.nih.gov/pub/taxonomy/taxdump.tar.gz'' accessing ftp://ftp.ncbi.nih.gov/pub/taxonomy/taxdump.tar.gz
```

bgruening: Jupp, thats why I ended up with the more relaxed mode. Maybe the more relaxed one is ok, if we turn this into a warning?

nsoranzo: I think using `requests` for HTTP and `urllib2` for FTP would be the best of both worlds, `requests` is so much better than Python standard library. I've implemented this in https://github.com/galaxyproject/planemo/pull/809 .

PR #809, opened 2018-05-01
Use requests.get() when validating http URLs

nsoranzo:
Tested on all dirs in https://github.com/TGAC/earlham-galaxytools and 40 dirs in tools-iuc.

Alternative to #808.

### DIFF
--- a/planemo/lint.py
+++ b/planemo/lint.py
@@ -2,18 +2,10 @@
 from __future__ import absolute_import
 
 import os
-import socket
 
 import requests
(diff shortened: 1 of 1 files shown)


nsoranzo:
No idea why the codeml tests fail (due to an empty line diff per output file), they pass on my laptop.

jmchilton:
Looks good to me!

bgruening:
Thanks @nsoranzo!

PR #814, opened 2018-05-11
Improvements to waiting on Galaxy to startup.

jmchilton:


bgruening:
Thanks John. Maybe this is also helpful: https://github.com/galaxyproject/ephemeris/blob/master/ephemeris/sleep.py#L39

jmchilton:
Thanks for the suggestion @bgruening, I've Created an issue for that but I'm rushing this out though because it worked to make workflow testing against docker-galaxy-stable work on Travis (https://github.com/jmchilton/planemo-workflow-test-template/pull/1). Hope that is okay.

PR #826, opened 2018-05-16
Fix "planemo conda_search".

### LINKED ISSUE #824: Planemo conda_search broken, not using best practice channels.

jmchilton:
There was a change in the galaxy-lib interface it used that broke it, and the underlying CLI seems to require * to do a more SQL-like search now so I've updated that as well.

Fixes #824.

### DIFF
--- a/planemo/commands/cmd_conda_search.py
+++ b/planemo/commands/cmd_conda_search.py
@@ -23,4 +23,5 @@ def cli(ctx, term, **kwds):
     Implicitly adds channels Planemo is configured with.
     """
     conda_context = build_conda_context(ctx, handle_auto_init=True, **kwds)
-    conda_context.exec_command("search", [term])
+    args = conda_context._override_channels_args + ["*%s*" % term]
+    conda_context.exec_command("search", args)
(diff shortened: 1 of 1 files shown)


planemo/commands/cmd_conda_search.py
```diff
@@ -23,4 +23,5 @@ def cli(ctx, term, **kwds):
     Implicitly adds channels Planemo is configured with.
     """
     conda_context = build_conda_context(ctx, handle_auto_init=True, **kwds)
-    conda_context.exec_command("search", [term])
+    args = conda_context._override_channels_args + ["*%s*" % term]
```
nsoranzo: This is not going to work on older Conda:
```
$ ~/miniconda3/bin/conda --version
conda 4.3.33
$ ~/miniconda3/bin/conda search '*samt*'

CommandArgumentError: Failed to compile regex pattern for search: *samt*
regex error: error('nothing to repeat at position 0',)

```

jmchilton:
Thanks for the review and the catch @nsoranzo.

PR #841, opened 2018-05-22
Add --no_shed_install option.

jmchilton:
@bgruening's original advice was to base this on whether ``--galaxy_admin_key`` was an actual admin key or not and just disable this functionality if not. My issue with this is that Planemo assumes a default admin key (that works with ``planemo serve`` defaults), so if the admin key is wrong and required I think it should be a hard failure not a silent behavior change. To put this another way, Planemo is tool development SDK, the correct assumption is that the target Galaxy is something it can just install tools into without worry - if that is not a correct assumption you need to change the behavior of Planemo and to do that the ``--no_shed_install`` flag makes a lot of sense I think.

I understand the critique that there are too many commands with too many options, and I have done things to make the default behaviors better (for instance none of the --conda_*** flags really need to be used for typical runs now). Rather than restricting functionality or making too clever assumptions I prefer the to go the route of documentation that focused on the actual commands to run and the relevant options for that use case. Therefore I've updated the documentation related to external Galaxy testing - and that doesn't cover twenty options - it covers only three - the most important three for external Galaxy testing including this new one.

### DIFF
--- a/docs/test_format.rst
+++ b/docs/test_format.rst
@@ -33,9 +33,9 @@ same testing file format is used for both kinds of artifacts.
       job: pear_job.yml
       outputs:
         assembled_pairs:
-          file: test-data/pear_assembled_results1.fastq
+          path: test-data/pear_assembled_results1.fastq
         unassembled_forward_reads:
-          file: test-data/pear_unassembled_forward_results1.fastq
+          path: test-data/pear_unassembled_forward_results1.fastq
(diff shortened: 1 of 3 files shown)


bgruening:
Tested and works for me. I still would argue that the admin access is not necessary but this is a great feature and I was able to test workflows against usegalaxy.eu.

PR #847, opened 2018-05-27
new database version

bgruening:
xref: https://github.com/jmchilton/galaxy-downloads


jmchilton:
Thanks!

jmchilton:
There is a bug in Planemo which is causing older target Galaxies to grab this version of latest, I need to revert the latest change I think for now. Sorry.

bgruening:
:(
But yes, I updated all travis.yml to use 18.05.

PR #860, opened 2018-08-03
Allow naming history from command line

erasche:
For testing of workflows specifically. Allows overriding the default "CWL Target History".

Rationale: EU is testing workflows and I considered to automatically make the history publicly available (especially in case of failure), and having these named something more distinct/appropriate would make life easier.


bgruening:
Failing test seems to be unrelated. Thanks @erasche!

PR #864, opened 2018-08-31
Update cwltool and galaxy-lib dependencies for Python 3.7

nsoranzo:
Fix https://github.com/galaxyproject/planemo/issues/863 .

Need a release which includes https://github.com/common-workflow-language/cwltool/pull/831 .

PR #866, opened 2018-09-07
Sync galaxy.xsd from galaxy repo

nsoranzo:
We need a new Planemo release that doesn't fail the linting when `fatal_oom` is used, see e.g. https://github.com/galaxyproject/tools-devteam/pull/503 .


nsoranzo:
Test failures are unrelated and would be fixed by #864.

PR #867, opened 2018-09-12
change the channel priority of conda (again)

bgruening:
* removed `r` channel
* conda-forge has not the highest priority


bgruening:
We would need a new release of planemo and then we need to update https://github.com/galaxyproject/planemo-monitor/blob/master/monitor.sh#L3

PR #870, opened 2018-09-13
Fix Planemo writing a file called gx_venv_None.

jmchilton:
Was meant to be gx_venv.


planemo/galaxy/run.py
```diff
@@ -53,7 +53,7 @@ def setup_venv(ctx, kwds):
 def locate_galaxy_virtualenv(ctx, kwds):
     if not kwds.get("no_cache_galaxy", False):
         workspace = ctx.workspace
-        galaxy_branch = kwds.get("galaxy_branch", "master")
+        galaxy_branch = kwds.get("galaxy_branch", None) or "master"
```
nsoranzo: The `, None` is redundant.

PR #874, opened 2018-09-29
Allow selection of python version when starting managed Galaxy

mvdbeek:
Make sure python 3 is on PATH and enjoy `planemo {test, serve} --galaxy_branch release_18.09 --galaxy_python_version 3`.


mvdbeek:
Hmm, this is Paste failing now due to deleting from a dict while iterating over it. The fix is simple and I've done it locally, but Paste looks pretty much unmaintained.

mvdbeek:
Hmm, so for some reason when installing dependencies by running common_startup.sh we get the wrong shebang (it points to the python in the virtualenv in which planemo was installed), which is a problem for the gunicorn entrypoint. I haven't figured out what's going on there, seems very strange.

mvdbeek:
So I'm having `__PYVENV_LAUNCHER__` set in my python 3 virtualenv. This get's propagated to new commands lauchned by plaenmo becasue in galaxylib we create a new env with `os.environ.copy` and update it. Just setting env['__PYVENV_LAUNCHER__'] = '' seems to solve the issue. Seems this problem might occur more frequently in the future based on https://github.com/pypa/virtualenv/issues/845#issuecomment-415189976

planemo/git.py
```diff
@@ -59,18 +59,18 @@ def checkout(ctx, remote_repo, local_path, branch=None, remote="origin", from_br
...
-    if bare:
-        bare_arg = "--bare"
+    mirror_arg = ""
+    if mirror:
+        mirror_arg = "--mirror"
     branch_arg = ""
     if branch is not None:
         branch_arg = "--branch '%s'" % branch
-    cmd = "git clone %s %s '%s' '%s'" % (bare_arg, branch_arg, src, dest)
+    cmd = "git clone %s %s '%s' '%s'" % (mirror_arg, branch_arg, src, dest)
```
nsoranzo: This should be better a list, no modifications in the rest of the code seem necessary.

mvdbeek:
yay, all green!

.travis.yml
```diff
@@ -1,29 +1,33 @@
...
+    - python: 2.7
+      env: TOX_ENV=py27-lint-readme
+    - python: 2.7
+      env: TOX_ENV=py27-lint-docs
+    - python: 2.7
+      env: TOX_ENV=py27-lint-docstrings
+    - python: 2.7
+      env: TOX_ENV=py27
+    - python: 3.6
+      env: TOX_ENV=py36-quick
```
nsoranzo: It would be useful to keep `py3?-lint` I think, since flake8 performs slightly different checks for the various Python versions.

Also, maybe put `py36-quick` before `py27` again, to have a faster feedback.

nsoranzo:
> So I'm having `__PYVENV_LAUNCHER__` set in my python 3 virtualenv. This gets propagated to new commands lauched by planemo because in galaxy-lib we create a new env with `os.environ.copy` and update it. Just setting `env['__PYVENV_LAUNCHER__'] = ''` seems to solve the issue. Seems this problem might occur more frequently in the future based on https://github.com/pypa/virtualenv/issues/845#issuecomment-415189976

@mvdbeek A bugfix for this has been proposed in https://github.com/python/cpython/pull/9516

planemo/galaxy/config.py
```diff
@@ -1240,6 +1254,7 @@ def _install_with_command(ctx, config_directory, command, env, kwds):
     else:
         pip_install_command = ""
     setup_venv_command = setup_venv(ctx, kwds)
+    env['__PYVENV_LAUNCHER__'] = ''
```
mvdbeek: I guess the cleaner fix would be to omit `__PYVENV_LAUNCHER__` in galaxy-lib, that'd also be more in line with https://github.com/python/cpython/pull/9516. What do you think @nsoranzo ?

nsoranzo: Where exactly in galaxy-lib ? And at which of the joined commands below `__PYVENV_LAUNCHER__` starts creating issues?

mvdbeek: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/deps/commands.py#L88

It's hard to tease apart exactly where in the joined command that happens, but the issue is that shebang for all scripts in the newly created virtualenv points to planemo's virtualenv instead of the newly created virtualenv.

planemo/galaxy/config.py
```diff
@@ -1240,6 +1254,7 @@ def _install_with_command(ctx, config_directory, command, env, kwds):
     else:
         pip_install_command = ""
```
nsoranzo: I just noticed that, since commit 9867e5644a5f3a8f9d58c6896ad34df936bb4e48, `pip_installs` and `pip_install_command` could be removed, if desired.

planemo/io.py
```diff
@@ -182,6 +182,7 @@ def kill_pid_file(pid_file):
 
     pid = int(open(pid_file, "r").read())
     kill_posix(pid)
+    os.unlink(pid_file)
```
nsoranzo: This should probably be in a `try/except Exception: pass`, since the killed process can remove the pid file before exiting, in which case  `os.remove` would raise `OSError` or `FileNotFoundError`.

mvdbeek: Good point, thanks!

nsoranzo:
Thanks @mvdbeek!

Should we enable this for tools-iuc testing? :smiling_imp:

mvdbeek:
Sure! We can start with a blacklist and work our way down like we did with condafying tools.

PR #889, opened 2018-11-08
Sync galaxy xsd and fix tests

mvdbeek:
And test everything under python 3 if in py36 environment. Otherwise we get an old python 2 version from travis.


nsoranzo:
`galaxy.xsd` is included in galaxy-lib since v18.9.0 , not sure how difficult is to use that one instead.

nsoranzo:
The `os.makedirs` errors can be fixed by using `galaxy.util.path.safe_makedirs`, I'm going to push a commit here soon.

nsoranzo:
Actually the `os.makedirs` errors are due to `generate_tuto_from_wf` test not cleaning after itself when failing, so using `safe_makedirs` in the other tests may make debugging more confusing, I think.

PR #893, opened 2018-11-11
skip workflow outputs that do not have a `label` set

bgruening:
This will skip `workflow_outputs` that are not set or `"label": null`


bgruening:
Merci!

jmchilton:
De rien!

PR #895, opened 2018-11-18
 Do not skip Galaxy client build for ``planemo serve``

### LINKED ISSUE #845: Planemo serve doesn't serve up static contents with default command line invocation
`planemo serve <tool_xml>`. Using `--galaxy_branch release_18.05` works fine. Possibly we just need to update ~~the~~ galaxy's master branch ?

nsoranzo:
Fix #845 .

Also, convert an existing `gx_repo` git repository from bare to mirror before doing a `git remote update`.


planemo/galaxy/serve.py
```diff
@@ -35,6 +37,7 @@ def _serve(ctx, runnables, **kwds):
     if port is None:
         port = network_util.get_free_port()
         kwds["port"] = port
+    kwds['galaxy_skip_client_build'] = False
```
jmchilton: This should be done in cmd_serve and cmd_shed_serve I think, that way when serve is done here for workflow testing for instance this isn't rebuilt. Does that sound fair?

nsoranzo: Good point, fixed in a rebase!

nsoranzo:
Finally green!! :tada:

bgruening:
Thanks all!

nsoranzo:
Time for a new Planemo release supporting Galaxy 18.09?

jmchilton:
I let you all down on this - sorry :(. Just released this.

PR #900, opened 2018-11-23
Allow composite inputs to workflows. 

jmchilton:


tests/data/wf6-composite-inputs.gxwf-test.yml
```diff
@@ -0,0 +1,11 @@
+- doc: Test Composite Inputs
+  job:
+    input1:
+      class: File
+      filetype: imzml
+      composite_data:
+        - Example_Continuous.imzML
```
bgruening: Should that be `location: URL` or `path: Example_Continuous.imzML`?

jmchilton:
Rebased and this seems to work in my hands with https://github.com/galaxyproject/galaxy/pull/9312. I guess we're going to forever be targeting dev releases galaxy-tool-util for Planemo.

I'll take a look at cleaning up the syntax now - @bgruening's suggestion above is probably a good one - even if the under the hood it is really just a list of paths.

tests/data/wf6-composite-inputs.gxwf-test.yml
jmchilton: Going to leave this as an open issue - it is an important issue though and needs to be done.

PR #908, opened 2019-01-03
Fix display of tool ids in planemo html report

mvdbeek:
If I understand `rSplit(testMethod, "_", 1);` correctly
the purpose here is to split the test_index from the tool id.

The test index is added in `_register_job_data` by doing
`tool_id + "-" + str(test_index)`, so just splitting at
`-` should have the desired effect.


mvdbeek:
changes
screen shot 2019-01-03 at 20 15 50
into
screen shot 2019-01-03 at 20 16 47

PR #909, opened 2019-01-09
single quotes for file names

bernt-matthias:
While giving a planemo tutorial I realized that tool_init creates double quoted file names. I guess single quotes are better.

PR #910, opened 2019-01-10
Fix doc linting

mvdbeek:


peterjc:
Looks good to me too (have just seen the problem via #909)

PR #917, opened 2019-04-03
Drop amqp workaround

mvdbeek:
If there's no message to be sent there's no activity,
so I don't think this was needed.
If there's a message it should be pretty lightweight too,
I can run 32 processes with a sqlite database and not run into locking
issues when sending messages.


mvdbeek:
@jmchilton can you have a look at this ? if there are concerns we could also setup a separate sqlite database for the queue.

jmchilton:
Sorry for the delay, I don't catch PRs not on galaxyproject/galaxy - I need a better workflow. I trust you if you say it isn't a problem, happy to merge. Thanks!

mvdbeek:
Awesome, thanks!

PR #921, opened 2019-04-16
Use ``yaml.safe_load()`` instead of ``load()``

nsoranzo:
Fix warnings like:

```
planemo/shed/__init__.py:436: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  config = yaml.load(f)
```

### DIFF
--- a/planemo/bioc_tool_builder.py
+++ b/planemo/bioc_tool_builder.py
@@ -234,7 +234,7 @@ class Requirement(object):
                                            "r-" + name.lower(),
                                            "meta.yaml")
             with open(recipe_path, 'r') as f:
-                doc = yaml.load(f)
+                doc = yaml.safe_load(f)
                 if not version:
                     version = doc["package"]["version"]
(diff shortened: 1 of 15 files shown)


martenson:
This is pure refactoring, right? afair `safe()` is just a wrapper around `safe_load()` now.

nsoranzo:
You mean **load()** is just a wrapper?

And no, if you read at link in the message above it clearly states:

> Since load cannot break backward compatibility to call safe_load, the maintainers of PyYAML decided to simply deprecate the plain usage of load, and require that the user intentionally declare the Loader they desired.

and

> FullLoader: Loads the full YAML language. Avoids arbitrary code execution. This is currently (PyYAML 5.1) the default loader called by yaml.load(input) (after issuing the warning).

while `safe_load()` use the `SafeLoader`.

martenson:
My memory failed me then. 👀

PR #926, opened 2019-04-25
Training: Fix empty repeat + some formatting

bebatut:
Hi,

This PR fixes a bug identified during a workshop: generation of new tutorials from workflow was failing if a repeat parameter was empty in the workflow.

Additionally, I removed some debug prints but also the references in the template (move now to bibtex on the training material website). I also added an extra print to help users to change default contributors/maintainers.

Bérénice


planemo/training/tool_input.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, tool_inp_desc, wf_param_values, wf_steps, level, should_be_th
             if not should_be_there:
                 info("%s not in workflow" % self.name)
             else:
-                raise ValueError("%s not in workflow" % self.name)
+                raise ValueError("%s not in workflow for %s" % (self.name, self.name))
```
nsoranzo: Both `self.name` ? Doesn't seem right.

planemo/training/tutorial.py
```diff
@@ -555,9 +555,7 @@ def format_wf_steps(wf, gi):
     steps = wf['steps']
 
     for s in range(len(steps)):
-        print('format_wf_steps')
         wf_step = steps[str(s)]
-        pprint(wf_step)
```
nsoranzo: Need to remove line 8 `from pprint import pprint` now that is unused.

planemo/training/tool_input.py
bebatut: Nice catch!

nsoranzo:
@mvdbeek Should when change https://github.com/galaxyproject/planemo/blob/master/tests/data/data_manager/data_manager_fetch_genome_dbkeys_all_fasta/data_manager/data_manager_fetch_genome_all_fasta_dbkeys.xml#L88 to fix https://travis-ci.org/galaxyproject/planemo/jobs/547779422#L4221 ?

jmchilton:
I'm sorry I missed this - @bebatut can you resolve the conflict and merge it?

PR #930, opened 2019-05-05
doc: add a little warning for <param ... multiple="true">

bernt-matthias:
I stumbled over this multiple times... 

Wondering what happens if the user selects multiple collections in such an input field

### DIFF
--- a/docs/_writing_collections.rst
+++ b/docs/_writing_collections.rst
@@ -10,8 +10,8 @@ types - ``list`` and ``paired``.
 * A **list** is a collection of datasets (or other collections) where
   each element has an ``identifier``. Unlike Galaxy dataset names which are
   transformed throughout complex analyses - the ``identifier`` is generally
-  perserved and can be used for concepts such as ``sample`` name that one wants to
-  perserve in the earlier mapping steps of a workflow and use it
+  preserved and can be used for concepts such as ``sample`` name that one wants to
+  preserve in the earlier mapping steps of a workflow and use it
   during reduction steps and reporting later.
 
 * The **paired** collection type is much simpler and more specific to sequencing
@@ -29,7 +29,7 @@ dataset having a list idenifier, each pair of datasets does.
 Consuming Collections
 -------------------------------
 
-Many Galaxy tools can be used without modification in conjuction with collections.
+Many Galaxy tools can be used without modification in conjunction with collections.
 Galaxy users can take a collection and `map over` any tool that
 consumes individual datasets. For instance, early in typical bioinformatics
 workflows you may have steps that filter raw data, convert to standard
@@ -48,7 +48,7 @@ identifier and index for the output entries match those of the supplied input.
 If a tool's functionality can be applied to individual files in isolation, the
 implicit mapping described above should be sufficient and no knowledge of collections
 by tools should be needed. However, tools may need to process multiple
-files at once - in this case explict collection consumption is required. This
+files at once - in this case explicit collection consumption is required. This
 document outlines three cases:
 
  * consuming pairs of datasets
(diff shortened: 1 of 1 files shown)


docs/_writing_collections.rst
```diff
@@ -125,13 +125,17 @@ Processing Lists (Reductions)
...
+``list`` but whenever possible, it might actually be better to not explicitly
 consume lists as a tool author. Parameters of type ``data`` can include a
 ``multiple="True"`` attribute to allow many datasets to be selected
 simultaneously. While the default UI will then have Galaxy users pick
-individual datsets, they can choose a collections as the tool can
+individual datasets, they can choose a collections as the tool can
 process both. This has the benefit of allowing tools to
 process either individual datasets or collections.
+A noteworthy difference is that if ``multiple="True"`` is used the elements of
+the collection are passed to the tool as (python) list, i.e., it is not
```
nsoranzo:
```suggestion
the collection are passed to the tool as a (Python) list, i.e. it is not
```

docs/_writing_collections.rst
```diff
@@ -125,13 +125,20 @@ Processing Lists (Reductions)
...
 consume lists as a tool author. Parameters of type ``data`` can include a
 ``multiple="True"`` attribute to allow many datasets to be selected
 simultaneously. While the default UI will then have Galaxy users pick
-individual datsets, they can choose a collections as the tool can
+individual datasets, they can choose a collections as the tool can
 process both. This has the benefit of allowing tools to
 process either individual datasets or collections.
+A noteworthy difference is that if ``multiple="True"`` is used the elements of
+the collection are passed to the tool as a (Python) list, i.e. it is not
+possible
```
nsoranzo:
```suggestion
possible:
```

docs/_writing_collections.rst
```diff
@@ -125,13 +125,20 @@ Processing Lists (Reductions)
...
+
+- to find out if a collection was passed, 
+- to access properties of the collection (name,...), or
+- to write tests that pass collections to the parameter (which would allow to name the elements explicitly)
```
nsoranzo:
```suggestion
- to write tests that pass a collection to the parameter (which would allow to name the elements explicitly).
```

docs/_writing_collections.rst
```diff
@@ -125,13 +125,22 @@ Processing Lists (Reductions)
...
 process either individual datasets or collections.
+A noteworthy difference is that if ``multiple="True"`` is used the elements of
+the collection are passed to the tool as a (Python) list, i.e. it is not
+possible:
+
+- to find out if a collection was passed, 
+- to access properties of the collection (name,...), or
+- to write tests that pass a collection to the parameter (which would allow to name the elements explicitly).
+
+Another drawback is that the `${on_string}` of the label contains the list of data sets in the collection (which can be confusing, since these data sets are in most cases hidden) and not the name of the collection.
```
nsoranzo: That's probably something which we may want to fix?

docs/_writing_collections.rst
```diff
@@ -125,13 +125,22 @@ Processing Lists (Reductions)
 -------------------------------
 
 The ``data_collection`` parameter type can specify a ``collection_type`` or
-``list`` but whenever possible, it is actually better to not explicitly
+``list`` but whenever possible, it might actually be better to not explicitly
```
jmchilton: This represented an IUC recommendation... I think I agree with the language changes here - but I was always more pro-collection inputs then everyone else. So I'm nervous about making this less of a recommendation.

bernt-matthias: Agreed. I changed the wording to "it is recommended".

docs/_writing_collections.rst
```diff
@@ -125,13 +125,22 @@ Processing Lists (Reductions)
...
+``list`` but whenever possible, it is recommended to not explicitly
 consume lists as a tool author. Parameters of type ``data`` can include a
 ``multiple="True"`` attribute to allow many datasets to be selected
 simultaneously. While the default UI will then have Galaxy users pick
-individual datsets, they can choose a collections as the tool can
-process both. This has the benefit of allowing tools to
+individual datasets, they can choose a collections as the tool can
+process both. This has the benefit of allowing tools to 
 process either individual datasets or collections.
+A noteworthy difference is that if a parameter of type ``data`` wth ``multiple="True"`` is used the elements of
```
nsoranzo:
```suggestion
A noteworthy difference is that if a parameter of type ``data`` with ``multiple="true"`` is used, the elements of
```

jmchilton:
Thanks so much!

Issue #939, opened 2019-06-25
planemo test can't handle same tool used twice in a workflow


davidchristiany:
I am using planemo test on a galaxy workflow with this command:
```
planemo test --galaxy_url <url> --galaxy_user_key <user_key> --no_shed_install --engine external_galaxy workflow.ga
```

It is working when my workflow only use each tool once:
```
All 1 test(s) executed passed.
workflow_frogs.ga_0: passed
```

But when I am using the same tool twice, I get an error:

```
galaxy.tools.verify INFO: ## files diff on '/home/dchristiany/proteore_project/ProteoRE/workflows/expected/Filters_excluded.tsv' and '/tmp/tmpp1xH50Filters_excluded.tsv': lines_diff = 0, found diff = 785`
`galaxy.tools.verify INFO: ## files diff on '/home/dchristiany/proteore_project/ProteoRE/workflows/expected/Filters_sequences.fasta' and '/tmp/tmp4lsKOhFilters_sequences.fasta': lines_diff = 0, found diff = 86`
`galaxy.tools.verify INFO: ## files diff on '/home/dchristiany/proteore_project/ProteoRE/workflows/expected/Filters_report.html' and '/tmp/tmpLbQaXhFilters_report.html': lines_diff = 0, found diff = 8`
`Traceback (most recent call last):
  File "/home/dchristiany/.venv/bin/planemo", line 10, in <module>
    sys.exit(planemo())
  File "/home/dchristiany/.venv/local/lib/python2.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/dchristiany/.venv/local/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/dchristiany/.venv/local/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/dchristiany/.venv/local/lib/python2.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/dchristiany/.venv/local/lib/python2.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/dchristiany/.venv/local/lib/python2.7/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
[...]

PR #944, opened 2019-07-16
Use ``unicodify()`` on exceptions and subprocess outputs

nsoranzo:
Fix traceback reported in https://github.com/galaxyproject/planemo/issues/939#issuecomment-505766544 .

Also fix 2 old broken tests.

### DIFF
--- a/planemo/commands/cmd_run.py
+++ b/planemo/commands/cmd_run.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 import json
 
 import click
+from galaxy.util import unicodify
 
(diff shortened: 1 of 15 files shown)


nsoranzo:
@mvdbeek Thanks for the merge!

I think we should apply the same changes (in particular adding `sort_keys=True` to `json.dump`) to all data managers in your https://github.com/galaxyproject/tools-iuc/pull/2032 , which should simplify the DM testing.

mvdbeek:
Yeah, I agree.

PR #958, opened 2019-09-25
Docker template changes

bedroesb:
This is to be aligned with the latest changes in :  https://github.com/galaxyproject/training-material/pull/1332


bedroesb:
Travis fails, but don't think it has something to do with my changes

nsoranzo:
Thanks @bedroesb !

PR #959, opened 2019-09-27
Add support for suite of repos with different owners

nsoranzo:
Useful when merging existing tools from different owners into a common
directory, producing a suite.

Example use case: https://github.com/galaxyproject/tools-iuc/pull/2612


bgruening:
Uh, this is cool. Thanks @nsoranzo!

PR #961, opened 2019-10-01
explicit TMPDIR note

martin-raden:
info what env var to set in order to change the temporary directory used by planemo

### DIFF
--- a/docs/commands/serve.rst
+++ b/docs/commands/serve.rst
@@ -26,7 +26,7 @@ option or force planemo to download a disposable instance with the
 ``planemo`` will run the Galaxy instance in an existing virtualenv if one
 exists in a ``.venv`` directory in the specified ``--galaxy_root``.
 Otherwise, the Galaxy instance will run in a clean virtualenv created in
-``/tmp``.
+``/tmp`` (or the directory from ``$TMPDIR``).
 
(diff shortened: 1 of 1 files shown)


docs/commands/serve.rst
```diff
@@ -26,7 +26,7 @@ option or force planemo to download a disposable instance with the
 ``planemo`` will run the Galaxy instance in an existing virtualenv if one
 exists in a ``.venv`` directory in the specified ``--galaxy_root``.
 Otherwise, the Galaxy instance will run in a clean virtualenv created in
-``/tmp``.
+``/tmp`` (or the directory from ``$TMPDIR``).
```
nsoranzo: I think Planemo nowadays caches the virtualenv in `~/.planemo/gx_venv/` , no?

martin-raden: I can confirm that the virtualenv goes to `~/.planemo/gx_venv/`

it seems the problem is, if you run on a system without installed dependencies, the pip update etc. run by planemo is causing troubles if `/tmp` is not writeable nad you do not set the `TMPDIR` path

so most likely the wrong spot to put this information

thanks @nsoranzo to check it..

martin-raden:
wrong docu spot

PR #964, opened 2019-10-08
Move most tests to Python 3.7. Updates for release

nsoranzo:
- Drop Python 3.4.
- Refactor `tox.ini`
- Add Python3 support to some files
- Update docs
- Update `galaxy.xsd` from upstream

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,22 +9,22 @@ matrix:
   include:
     - python: 2.7
       env: TOX_ENV=py27-lint
+    - python: 3.7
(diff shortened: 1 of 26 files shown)


bgruening:
Thanks @nsoranzo, looks good to me. Will test tm locally as well.

jmchilton:
I'm not good at Github - my landing page is just github.com/galaxyproject/galaxy so I miss a lot. Just seeing this now - it is way awesome - thanks a bunch @nsoranzo!

PR #974, opened 2019-11-25
Add merge_test_reports command

mvdbeek:
That will be useful when running all tool tests divided up into chunks and then combining the output. Could also be done with `jq`, but couldn't quite figure out how to get this right.

### DIFF
--- a/planemo/galaxy/test/actions.py
+++ b/planemo/galaxy/test/actions.py
@@ -131,6 +131,19 @@ def handle_reports_and_summary(ctx, structured_data, exit_code=None, kwds={}):
     return exit_code if exit_code is not None else summary_exit_code
 
 
+def merge_reports(input_paths, output_path):
+    reports = []
+    for path in input_paths:
+        with io.open(path, encoding='utf-8') as f:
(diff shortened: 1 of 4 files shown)


planemo/galaxy/test/actions.py
```diff
@@ -131,6 +131,19 @@ def handle_reports_and_summary(ctx, structured_data, exit_code=None, kwds={}):
...
 
+def merge_reports(input_paths, output_path):
+    reports = []
+    for path in input_paths:
+        with io.open(path, encoding='utf-8') as f:
+            reports.append(json.load(f))
+    tests = []
+    for report in reports:
+        tests.extend(report["tests"])
+    merged_report = {"tests": tests for report in reports}
```
bgruening:
```suggestion
    merged_report = {"tests": tests}
````

mvdbeek: Good one ... what was I thinking ?

bgruening: Looks like you wanted to do the short way and then decided to do the more readable way :)

PR #976, opened 2019-12-08
Implement github workflow and fix profile commands if psql unavailable 

mvdbeek:


mvdbeek:
workflow for this commit is running at: https://github.com/mvdbeek/planemo/commit/0298f8494ce4fe36f681ec5f9a53fd83b0bc1bb1/checks?check_suite_id=347681728

jmchilton:
Amazing!

jmchilton:
Have you seen tox-gh-actions? https://github.com/galaxyproject/gxformat2/commit/fea27f0e1179c3467e16f1ab062daf7cf724c3b3

There isn't a ton of cruft doing it explicitly the way you have here - so I'm not sure it is worth introducing the black box magic tox-gh-actions does. But the magic is kind of cool at the same time.

martenson:
> It is a cool way to skip running the doc tests and readme test on all but one profile 

The speed should be worth the blackbox I think. :+1:

mvdbeek:
I don't think it is, we're talking seconds vs hours in build time.

martenson:
Sorry, I am thinking Galaxy, where linting is long. 🙈

mvdbeek:
Still makes sense to look into it if we also want to test a wider range of python versions, we could skip the full-blown serve

PR #977, opened 2019-12-08
Build single requirement container, log if requirement not in best-pr…

mvdbeek:
…actice channels

This also revealed quite a headache: We frequently don't specify exact versions in tool requirements (e.g `<requirement type="package" version="3.7">conda</requirement>`) instead of the full 3.7.1. This is very widespread in Galaxy tools. A single requirement like this will prevent building a container :(.

### DIFF
--- a/planemo/commands/cmd_container_register.py
+++ b/planemo/commands/cmd_container_register.py
@@ -81,10 +81,9 @@ def cli(ctx, paths, **kwds):
                 ctx.log("Target [%s] is not available in best practice channels - skipping" % conda_target)
                 best_practice_requirements = False
             if not exact:
-                ctx.log("Target version [%s] for package [%s] is not available in best practice channels - skipping",
+                ctx.log("Target version [%s] for package [%s] is not available in best practice channels - please specify the full version",
                         conda_target.version,
                         conda_target.package)
-                best_practice_requirements = False
 
         if not best_practice_requirements:
             continue
--- a/planemo/commands/cmd_container_register.py
+++ b/planemo/commands/cmd_container_register.py
@@ -74,29 +74,40 @@ def cli(ctx, paths, **kwds):
         ctx.vlog("Handling conda_targets [%s]" % conda_targets)
         mulled_targets = conda_to_mulled_targets(conda_targets)
         mulled_targets_str = "- " + "\n- ".join(map(conda_build_target_str, mulled_targets))
-        if len(mulled_targets) < 2:
-            ctx.vlog("Skipping registeration, fewer than 2 targets discovered.")
-            # Skip these for now, we will want to revisit this for conda-forge dependencies and such.
-            continue
-
         best_practice_requirements = True
         for conda_target in conda_targets:
             best_hit, exact = best_practice_search(conda_target, conda_context=conda_context)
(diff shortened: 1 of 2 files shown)


planemo/commands/cmd_container_register.py
```diff
@@ -74,30 +74,40 @@ def cli(ctx, paths, **kwds):
...
-        if os.path.exists(target_filename):
-            ctx.vlog("Target file already exists, skipping")
+        if len(mulled_targets) < 1:
+            ctx.log("Skipping registration, no targets discovered.")
             continue
 
+        name = v2_image_name(mulled_targets)
+
+        if len(mulled_targets) == 1:
+            target_filename = os.path.join(registry_target.output_directory, "hash.tsv")
```
bgruening: @mvdbeek what about putting those also into separate files. Like here: https://github.com/BioContainers/multi-package-containers/pull/795/files

In this way we really can distinguish bettwen manually added containers and automatic ones.

mvdbeek: I'm not in favor of this. Much easier to just parse one file, no chance of duplication. If there is a reason to discriminate automatically built from not automatically built I don't think the difference should be a single file vs in one big file.

bgruening: You don't need to parse a file, you look up the path as it is done already with the multi-containers and like it is working for years. If the file is there, skip it. No parsing needed imho.

mvdbeek: We should be parsing the file anyway to check if the build already exists. This also avoids any potential character issue (utf-8 on non-utf8 systems, or other fairly arbitrary character restrictions on samba/cifs. That wasn't an issue with multi-containers because of the hash.

mvdbeek: I mean it doesn't matter too much. j think a single file is more user-friendly, we can always revisit this later if it is really an issue. I'll update the code.

mvdbeek: Just updated it, thanks @bgruening!

planemo/commands/cmd_container_register.py
```diff
@@ -163,17 +165,21 @@ def handle_pull_request(self, ctx, name, target_filename, packages_str, tools_st
...
-            for target in mulled_targets:
-                if target.version:
-                    target_str = "%s=%s" % (target.package_name, target.version)
-                else:
-                    target_str = target.package_name
-                target_strings.append(target_str)
-            contents = ",".join(target_strings)
-            f.write(contents)
-            ctx.vlog("Wrote requirements [%s] to file [%s]" % (contents, target_filename))
+        with open(target_filename, "a") as f:
```
bgruening: I guess if the file already exists we should not append, isn't it? Is `a` really needed here?

mvdbeek: It's a leftover from appending to hash.tsv

bgruening:
@mvdbeek I think we need a new planemo release. Then we could start testing and migrating the monitor repo ...

mvdbeek:
We need to wait for the cwltool upgrade so that we can use a yet-to-be released galaxy-tool-util package with the mulled updates.

PR #978, opened 2019-12-19
Replace deprecated galaxy-lib requirement with galaxy-tool-util

nsoranzo:
Depends on https://github.com/common-workflow-language/cwltool/pull/1228 and https://github.com/galaxyproject/galaxy/pull/9146 (with corresponding new PyPI releases).


mr-c:
As mentioned elsewhere, https://github.com/common-workflow-language/cwltool/releases/tag/1.0.20191225192155 has been released with https://github.com/common-workflow-language/cwltool/pull/1228 incorporated

nsoranzo:
The call to cwltool's `load_tool.validate_document()` method in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/cwl/schema.py#L42 was broken in commit https://github.com/common-workflow-language/cwltool/commit/526f36f93655bfb098f766ff020708b5a707513a
Any help on a fix for that in galaxy-tool-util would be appreciated!

jmchilton:
It is almost certainly fixed in the CWL branch of Galaxy already.

jmchilton:
https://github.com/galaxyproject/galaxy/compare/dev...common-workflow-language:cwl-1.0-rebase-20191231?expand=1#diff-9f5ed620db2b8c1f6fbb4d19cfb8a33d

mvdbeek:
Hmm, somethings hanging in the full test suite, possibly `test_local_conda_dependencies_version`. I have no idea what is going on there.

jmchilton:
This process ``30877 s008  S+     0:00.00 /bin/bash -e /Users/john/miniconda3/conda-bld/fleeqtk_1579101053099/test_tmp/run_test.sh`` seems to hang locally. I don't know if this always hung on Macs though.

jmchilton:
Replacing the fleeqtk test with 'true' ... https://github.com/jmchilton/planemo/runs/391338751 ... lets see what happens.

mvdbeek:
The original seqtk recipe now has an additional patch ... perhaps that might be needed now too? https://github.com/bioconda/bioconda-recipes/blob/master/recipes/seqtk/meta.yaml#L12

jmchilton:
Might be better easier just patch fleeqtk if we think this is the problem.

nsoranzo:
I also see the `fleeqtk seq` hang during conda build locally. I've pushed a commit which changes the test a bit, seems to work for me.

mvdbeek:
Yeah, the additional patch doesn't make it a great example anymore. The original seqtk recipe builds fine, so let's patch fleeqtk

jmchilton:
If this doesn't fix things - lets cherry-pick this ( https://github.com/jmchilton/planemo/commit/371f3d2804092e7c39e2b5cf7b67a1a068980fd2 ) - I think that should build a new fleeqtk 1.4 with the patch applied.

jmchilton:
That was a long journey - thanks for the effort all!

PR #980, opened 2020-01-11
Make container_register build files with headers and include base_image

mvdbeek:
I think this is all we need to make significant progress with https://github.com/BioContainers/multi-package-containers/issues/940 and https://github.com/galaxyproject/tools-iuc/issues/2701

### DIFF
--- a/planemo/commands/cmd_container_register.py
+++ b/planemo/commands/cmd_container_register.py
@@ -3,6 +3,10 @@ import os
 import string
 
 import click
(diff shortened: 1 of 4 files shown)


mvdbeek:
Green and ready for review!

jmchilton:
Awesome - thanks a bunch! What else can I do to help?

bgruening:
I think a new release would be nice at some point. And then we need to update planemo-monitor, it has planemo pinned to an older version, imho.

PR #982, opened 2020-01-17
Add planemo list_repos command

mvdbeek:
This command lists all repositories defined by .shed.yml files on a file
system.

### DIFF
--- a/planemo/commands/cmd_list_repos.py
+++ b/planemo/commands/cmd_list_repos.py
@@ -1,0 +1,35 @@
+"""Module describing the planemo ``ci_find_repos`` command."""
+from __future__ import print_function
+
+import click
+
+from planemo import options
+from planemo.ci import print_as_yaml
+from planemo.cli import command_function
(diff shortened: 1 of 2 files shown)


planemo/commands/cmd_list_repos.py
```diff
@@ -0,0 +1,35 @@
...
+from planemo import options
+from planemo.ci import print_as_yaml
+from planemo.cli import command_function
+from planemo.shed import (
+    _realize_effective_repositories,
+    find_raw_repositories,
+)
+
+
+@click.command('ci_list_repos')
```
jmchilton: This is confusing - this has nothing to do with CI right? Why prefixed this way and why was print_as_yaml put in there? The ci module has to do with like diffing Travis branches and stuff doesn't it?

mvdbeek: Yep, forgot to push a commit that fixes that ...

PR #985, opened 2020-01-19
Drop planemo database seed option

mvdbeek:
It's easier and faster to let Galaxy create a database.

Let's see how much breaks here.

### DIFF
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -190,21 +190,6 @@ DOWNLOAD_GALAXY = (
     "wget -q https://codeload.github.com/galaxyproject/galaxy/tar.gz/"
 )
 
-DOWNLOADS_URL = ("https://raw.githubusercontent.com/"
-                 "jmchilton/galaxy-downloads/master/")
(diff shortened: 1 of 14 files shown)


planemo/galaxy/config.py
```diff
@@ -1003,84 +978,6 @@ def _database_connection(database_location, **kwds):
...
-    newest_migration = _newest_migration_version(galaxy_root, galaxy_branch)
-    for migration in DOWNLOADABLE_MIGRATION_VERSIONS:
-        if newest_migration >= migration:
-            download_migration = migration
-            break
-
-    if download_migration:
-        download_name = "db_gx_rev_0%d.sqlite" % download_migration
-        download_url = DOWNLOADS_URL + download_name
-        ctx.cache_download(download_url, database_location)
```
nsoranzo: The `Context.cache_download()` method (defined in `planemo/cli.py`) is now unused, not sure if we want to drop that too.

mvdbeek: I may reuse that for downloading the prebuilt galaxy client (https://github.com/mvdbeek/galaxy-client-cache/blob/master/.github/workflows/build_client.yaml)

Issue #857, opened 2018-07-13
symbolic links in test-data


bernt-matthias:
In https://github.com/galaxyproject/tools-iuc/pull/1963 I tried to use symbolic links in the test-data directory which led to: 

```
Traceback (most recent call last):
  File "/tmp/tmpH13feS/galaxy-dev/test/functional/test_toolbox.py", line 88, in test_tool
    self.do_it(tool_version=tool_version, test_index=test_index)
  File "/tmp/tmpH13feS/galaxy-dev/test/functional/test_toolbox.py", line 41, in do_it
    verify_tool(tool_id, self.galaxy_interactor, resource_parameters=resource_parameters, test_index=test_index, tool_version=tool_version, register_job_data=register_job_data)
  File "/tmp/tmpH13feS/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 653, in verify_tool
    stage_data_in_history(galaxy_interactor, tool_id, testdef.test_data(), test_history)
  File "/tmp/tmpH13feS/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 59, in stage_data_in_history
    upload_wait = galaxy_interactor.stage_data_async(test_data, history, tool_id)
  File "/tmp/tmpH13feS/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 267, in stage_data_async
    file_name = self.test_data_path(tool_id, fname)
  File "/tmp/tmpH13feS/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 232, in test_data_path
    assert response.status_code == 200
AssertionError
```

Issue #865, opened 2018-09-02
Allow test-data to be symlink


mvdbeek:
At least the simple case where the test-data directory is a symlink to somewhere else fails.
This might be a Galaxy issue, not sure.

mvdbeek:
fwiw the erros is:
```
Traceback (most recent call last):
  File "/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpmgem8dw6/galaxy-dev/test/functional/test_toolbox.py", line 88, in test_tool
    self.do_it(tool_version=tool_version, test_index=test_index)
  File "/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpmgem8dw6/galaxy-dev/test/functional/test_toolbox.py", line 41, in do_it
    verify_tool(tool_id, self.galaxy_interactor, resource_parameters=resource_parameters, test_index=test_index, tool_version=tool_version, register_job_data=register_job_data)
  File "/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpmgem8dw6/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 653, in verify_tool
    stage_data_in_history(galaxy_interactor, tool_id, testdef.test_data(), test_history)
  File "/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpmgem8dw6/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 59, in stage_data_in_history
    upload_wait = galaxy_interactor.stage_data_async(test_data, history, tool_id)
  File "/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpmgem8dw6/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 267, in stage_data_async
    file_name = self.test_data_path(tool_id, fname)
  File "/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpmgem8dw6/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 232, in test_data_path
    assert response.status_code == 200
AssertionError
```

mvdbeek:
Hmm, I guess this is actually a security feature to not let the test-data be outside of the test-data dir.
[...]

PR #988, opened 2020-01-26
Make symlinks in tool tree work for planemo test

mvdbeek:
Fixes https://github.com/galaxyproject/planemo/issues/857 and https://github.com/galaxyproject/planemo/issues/865
Symlinks were problematic when the symlink pointed out of the tool
directory. This was not a problem when uploading to the tool shed,
because we'd dereference the symlinks when creating the repository
archive. So here we just copy the tool tree with shutil.copytree,
which dereferences symlinks by default.

### DIFF
--- a/planemo/commands/cmd_test.py
+++ b/planemo/commands/cmd_test.py
@@ -1,6 +1,4 @@
 """Module describing the planemo ``test`` command."""
-from distutils.dir_util import copy_tree
-
 import click
 
 from planemo import options
@@ -96,13 +94,10 @@ def cli(ctx, paths, **kwds):
         else:
             ctx.vlog("Running traditional Galaxy tool tests using run_tests.sh in Galaxy root %s" % engine_type)
             kwds["for_tests"] = True
+            if kwds.get('update_test_data'):
+                non_copied_runnables = for_paths(paths)
+                kwds['test_data_target_dir'] = _find_test_data(non_copied_runnables, **kwds)
             with galaxy_config(ctx, runnables, **kwds) as config:
                 return_value = run_in_config(ctx, config, **kwds)
-                if kwds.get('update_test_data'):
-                    non_copied_runnables = for_paths(paths)
-                    target = _find_test_data(non_copied_runnables, **kwds)
-                    if config.test_data_dir != target:
-                        ctx.vlog("Copying %s to %s" % (config.test_data_dir, target))
-                        copy_tree(config.test_data_dir, target)
 
     ctx.exit(return_value)
--- a/planemo/commands/cmd_test.py
+++ b/planemo/commands/cmd_test.py
@@ -80,12 +80,12 @@ def cli(ctx, paths, **kwds):
     with temp_directory(dir=ctx.planemo_directory) as temp_path:
         # Create temp dir(s) outside of temp, docker can't mount $TEMPDIR on OSX
         runnables = for_paths(paths, temp_path=temp_path)
-        is_cwl = all([r.type in [RunnableType.cwl_tool, RunnableType.cwl_workflow] for r in runnables])
-        if kwds.get("engine", None) is None:
+        is_cwl = all(r.type in {RunnableType.cwl_tool, RunnableType.cwl_workflow} for r in runnables)
+        if kwds.get("engine") is None:
             kwds["engine"] = "galaxy" if not is_cwl else "cwltool"
 
         engine_type = kwds["engine"]
-        test_engine_testable = (RunnableType.galaxy_tool, RunnableType.galaxy_datamanager, RunnableType.directory)
+        test_engine_testable = {RunnableType.galaxy_tool, RunnableType.galaxy_datamanager, RunnableType.directory}
         enable_test_engines = any(r.type not in test_engine_testable for r in runnables)
         enable_test_engines = enable_test_engines or engine_type != "galaxy"
         if enable_test_engines:
--- a/planemo/commands/cmd_test.py
+++ b/planemo/commands/cmd_test.py
@@ -1,4 +1,6 @@
 """Module describing the planemo ``test`` command."""
+from distutils.dir_util import copy_tree
+
 import click
 
 from planemo import options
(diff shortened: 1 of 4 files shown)


planemo/commands/cmd_test.py
```diff
@@ -73,23 +77,32 @@ def cli(ctx, paths, **kwds):
...
+            ctx.vlog("Running traditional Galaxy tool tests using run_tests.sh in Galaxy root %s" % engine_type)
+            kwds["for_tests"] = True
+            with galaxy_config(ctx, runnables, **kwds) as config:
+                return_value = run_in_config(ctx, config, **kwds)
+                if kwds.get('update_test_data'):
+                    non_copied_runnables = for_paths(paths)
+                    target = _find_test_data(non_copied_runnables, **kwds)
+                    if config.test_data_dir != target:
+                        ctx.vlog("Copying %s to %s" % (config.test_data_dir, target))
+                        copy_tree(config.test_data_dir, target)
```
nsoranzo: Shouldn't lines 99-106 be inside `run_in_config()` ?

mvdbeek: I guess we could, but we only want to do this for planemo test, which is why I placed it here (If we do this for planemo serve edits to the tool files won't be reloaded)

nsoranzo: `planemo serve` does not call `run_in_config()`, only `test` and `shed_test` do.

The reason to put it there would be that it's where the normal test data update is executed: https://github.com/galaxyproject/planemo/blob/master/planemo/galaxy/test/actions.py#L98-L100

mvdbeek: That also gives us the chance to only copy the files once. Le't see how https://github.com/galaxyproject/planemo/pull/988/commits/8656796d9fd9ae3900a84bc80e2894137499365f fares

nsoranzo:
I've tested this with a fake tool and it works correctly if the `test-data` directory is a symlink.
On the other hand, if `test-data` is a normal directory and some output test files are symlinks, with this PR the symlinks are replaced with updated real files, so I'm not sure it's the proper fix for #857 (according to @bernt-matthias's comment on #865).

mvdbeek:
Yes, that is what I intended, I don't think you'll want to overwrite symlink targets.

mvdbeek:
As far as I understand #857 this was mostly an issue with having test data point out of the runnable directory, which Galaxy doesn't allow. That will work fine with this PR, you'll just have to take care with `--update_test_data`

nsoranzo:
OK, let's see if @bernt-matthias has anything to add.

I seem to be able to replicate the same behaviour with just these changes:
```diff
diff --git a/planemo/galaxy/test/actions.py b/planemo/galaxy/test/actions.py
index 92cda85..4cf9c8b 100644
--- a/planemo/galaxy/test/actions.py
+++ b/planemo/galaxy/test/actions.py
@@ -3,9 +3,9 @@
 import io
 import json
 import os
+from distutils.dir_util import copy_tree
 
 import click
-from galaxy.tool_util.deps.commands import shell
 from galaxy.util import unicodify
 
 from planemo.exit_codes import (
@@ -96,8 +96,7 @@ def run_in_config(ctx, config, run=run_galaxy_command, **kwds):
         action
     )
     if kwds.get('update_test_data', False):
-        update_cp_args = (job_output_files, config.test_data_dir)
-        shell('cp -r "%s"/* "%s"' % update_cp_args)
+        copy_tree(job_output_files, config.test_data_dir)
 
     _check_test_outputs(xunit_report_file_tracker, structured_report_file_tracker)
     test_results = test_structures.GalaxyTestResults(
```

Are the other changes needed? In case, if you can add some comments explaining the reason, thanks!

mvdbeek:
> I seem to be able to replicate the same behaviour with just these changes:

The diff you posted will not expand symlinks in the runnable location, so when the test tool interactor asks for the test files from Galaxy Galaxy will deny this https://github.com/mvdbeek/galaxy/blob/fdfdcc50b45ea692e747db7cf05f9adac235ae72/lib/galaxy/tools/__init__.py#L972. That is the cause of #857 and #865.

I mostly wrote this for dockerized tests where on top of these restrictions the symlink target won't be mounted in leading to broken symlinks, not just in the test data but also for symlinked scripts as in ruvseq. If you want to check this for yourself I'd suggest running `planemo test --biocontainers` on ruvseq (symlinked script) or samtools_cram_to_bam test 4 (symlinked test table data).

We don't have any iuc tools around that would elicit #857, but you can simulate that with
```
cd tools/column_maker/test-data
mv column_maker_out1.interval ../../
ln -s ../../column_maker_out1.interval .
planemo test column_maker.xml
```
That'll fail with
```
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpwttc77ji/galaxy-dev/test/functional/test_toolbox.py", line 99, in test_tool
    self.do_it(tool_version=tool_version, test_index=test_index)
  File "/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpwttc77ji/galaxy-dev/test/functional/test_toolbox.py", line 36, in do_it
    verify_tool(tool_id, self.galaxy_interactor, resource_parameters=resource_parameters, test_index=test_index, tool_version=tool_version, register_job_data=register_job_data)
  File "/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpwttc77ji/galaxy-dev/lib/galaxy/tool_util/verify/interactor.py", line 789, in verify_tool
    raise e
JobOutputsError: Test file (column_maker_out1.interval) is missing. If you use planemo try --update_test_data to generate one.
```

jmchilton:
This is a lot of magic outside of Galaxy 😟. I don't know what to do otherwise though, thanks for the fixes.

mvdbeek:
Thanks @jmchilton! If it is any consolation this brings the test closer to what is being uploaded to the toolshed. Another alternative would have been to do some staging on the Galaxy side, which I think pulsar does anyway ?

PR #993, opened 2020-01-30
Fix quay repository presence check for single target builds

mvdbeek:
Since the v2_image_name for a single target doesn't include the build
suffix given by bioconda (e.`--hc9558a2_0`) we don't get an exact
match from the quay.io api.

### DIFF
--- a/planemo/commands/cmd_container_register.py
+++ b/planemo/commands/cmd_container_register.py
@@ -3,13 +3,13 @@ import os
 import string
 
 import click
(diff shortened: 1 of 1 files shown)


mvdbeek:
this isn't actually meant for merging, I think the logic actually belongs in galaxy-tool-util but I wanted to have a check to see if / what else is missing for planemo monitor.

mvdbeek:
OK, we already had everything in place to do this right in galaxy-tool-util.

PR #1001, opened 2020-03-10
More fine grained options for --shed_install

AndreasSko:
`--shed_install` uses the default options for installing tools from Ephemeris, resulting in automatic dependency resolution/installation on first run (see: https://github.com/galaxyproject/ephemeris/blob/6a01ac3de3f8e40145365523e3ae8807b9072282/src/ephemeris/shed_tools.py#L103). In some situations, where installing dependencies is not necessary (for example when using containers), disabling this would result in faster execution time.

With this change, the options `--install_tool_dependencies`, `--install_resolver_dependencies`, and `--install_repository_dependencies` are introduced, which enables to define tool installation using `--shed_install` more fine grained. The default options stay the same, so there should be no breaking change to existing usage.

### DIFF
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -726,7 +726,12 @@ class BaseGalaxyConfig(GalaxyInterface):
 
     def _install_workflow(self, runnable):
         if self._kwds["shed_install"]:
-            install_shed_repos(runnable, self.gi, self._kwds.get("ignore_dependency_problems", False))
+            install_shed_repos(runnable,
+                               self.gi,
+                               self._kwds.get("ignore_dependency_problems", False),
+                               self._kwds.get("install_tool_dependencies", False),
+                               self._kwds.get("install_resolver_dependencies", True),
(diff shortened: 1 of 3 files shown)


AndreasSko:
cc @bgruening

bgruening:
@AndreasSko a few tests are failing, can you please look at those?

AndreasSko:
```
Exception [Problem executing commands psql --tuples-only --username postgres -P pager=off --command 'drop database plnmoprof_profile1234;' - (, psql: could not connect to server: No such file or directory
Is the server running locally and accepting connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?)]
```
Seems strange..

AndreasSko:
Is it possible that the tests/pipeline itself is broken currently? Tested the master branch on my fork and it failed with the same problem: https://github.com/AndreasSko/planemo/actions/runs/53662366

jmchilton:
Those look like pre-existing problems - thanks for the enhancements! They look great!

PR #1003, opened 2020-03-16
Fixes for CI.

jmchilton:


jmchilton:
Still a red test in there but a lot better.

PR #1007, opened 2020-03-23
[WIP] Nested collection inputs

jmchilton:
Looks like they don't work at all - I underestimated how custom all the lower level code I wrote is.


jmchilton:
Implemented in https://github.com/galaxyproject/planemo/pull/1029 instead.

PR #1009, opened 2020-04-08
test: fix inconsistent test numbering

bernt-matthias:
planemo test output during test and html report count test number starting from 1. But the final summary started counting at 0.

### DIFF
--- a/planemo/galaxy/test/structures.py
+++ b/planemo/galaxy/test/structures.py
@@ -183,9 +183,7 @@ def case_id(testcase_el=None, raw_id=None):
         tool_and_num = raw_id.split("TestForTool_")[-1]
         if ".test_tool_" in tool_and_num:
             name, num_str = tool_and_num.split(".test_tool_", 1)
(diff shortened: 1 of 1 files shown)


jmchilton:
Can you clarify this a bit? I'm nervous this isn't the right thing to do

bernt-matthias:
Sure. For the first test of a tool there is an outputs like: 

`( TOOLNAME ) > Test-1 ... 2020-04-21 19:54:17,068 INFO  [galaxy.tools] Validated and populated state for tool request (18.386 ms)`

In the html output the test is given as 

`TOOLNAME (Test #1)`

The final output for the first test is 

`TOOLNAME[0]: passed`

So counting is 1-based for the first 2 outputs and 0-based for the last one. I would __hope__ that my change only changes the output. The comment which I removed somehow suggests that this was intended .. maybe in the past html output was also 0-based?

bernt-matthias:
Just checked again for the `master` branch. Seems to be fine now and the PR is not needed anymore.

PR #1021, opened 2020-06-11
change default python version

bernt-matthias:
is it really that simple, or is there something else that should be changed?

### DIFF
--- a/planemo/options.py
+++ b/planemo/options.py
@@ -134,7 +134,7 @@ def galaxy_python_version():
         '--galaxy_python_version',
         use_global_config=True,
         default=None,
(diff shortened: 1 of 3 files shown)


planemo/options.py
```diff
@@ -134,7 +134,7 @@ def galaxy_python_version():
         '--galaxy_python_version',
         use_global_config=True,
         default=None,
-        type=click.Choice(['2', '2.7', '3', '3.5', '3.6', '3.7', '3.8']),
```
nsoranzo, line 137: To support older Galaxy releases, we may want to keep 2/2.7 here.

bernt-matthias: True. Didn't think about this.

PR #1022, opened 2020-07-07
sort tests by id when merging

bernt-matthias:
Was a little paper cut for me when looking at larger reports (e.g. IUC CI results) that were not sorted alphabetically.


jmchilton:
Awesome - thanks!

PR #1023, opened 2020-07-17
Don't say Galaxy can't run on Python 3

mvdbeek:


lldelisle:
Thanks

docs/installation.rst
```diff
@@ -34,8 +34,8 @@ consider using the ``virtualenv`` argument ``-p`` to point at a non-Conda Python
 natively on your system or using a tool such pyenv_. ``virtualenv`` can be installed via Conda, pyenv_,
 or a package manager - it should make no difference.
 
-Planemo in theory runs under Python 3 but Galaxy does not, so it is best to use a Python 2 with Planemo
-for now.
+Planemo runs on Python 2.7 and Python 3.5 or newer. Planemo can be used to run multiple versions of Galaxy,
+but please note that the last Galaxy release that fully supports Python 2.7 is 19.09.
```
nsoranzo, line 38:
```suggestion
but please note that the last Galaxy release that fully supports Python 2.7 is 20.01.
```

mvdbeek: I don't think that's right, is it ? We stopped doing large parts of the CI on python 2 after the 19.09 release, and we wrote in the 20.01 release notes:
```
Deprecation Notice
With the release of Galaxy 20.01 we strongly encourage all deployers to run Galaxy under Python 3. For instructions on how to transition to Python 3 please see here.
```

nsoranzo: Python 2.7 was still supported and partially tested for 20.01, and we didn't introduce any breaking change that I know of. My institute's production instance is 20.01 under py2.7 too.

PR #1032, opened 2020-07-23
Update installation.rst

mblue9:
Hello, maybe you don't want this but am just suggesting it as it is what worked smoothly for me i.e. using python venv instead of virtualenv. With the virtualenv method I ended up in Stack Overflow trying to fix a similar situation to [this](https://stackoverflow.com/questions/31133050/virtualenv-command-not-found).

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -19,7 +19,7 @@ When installed this way, planemo can be upgraded as follows:
 
 ::
 
-    $ source planemo/bin/activate
+    $ . planemo/bin/activate
     $ pip install -U planemo
 
 To install or update to the latest development branch of planemo with ``pip``, 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -12,7 +12,7 @@ install with ``pip``. Planemo requires pip 7.0 or newer.
 ::
 
     $ python3 -m venv planemo
-    $ source planemo/bin/activate
+    $ . planemo/bin/activate
     $ pip install planemo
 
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -5,21 +5,21 @@ Installation
...
+For a traditional Python installation of Planemo, first set up a virtual environment
+for ``planemo`` (this example creates a new one in ``planemo``) and then
 install with ``pip``. Planemo requires pip 7.0 or newer.
 
 ::
 
-    $ virtualenv .venv; . .venv/bin/activate
-    $ pip install "pip>=7" # Upgrade pip if needed.
+    $ python3 -m venv planemo
+    $ source planemo/bin/activate
```
nsoranzo, line 15: `.` is POSIX and more portable than `source` (which is not available in e.g. dash)

```suggestion
    $ . planemo/bin/activate
```

docs/installation.rst
```diff
@@ -5,21 +5,21 @@ Installation
...
     $ pip install planemo
 
 When installed this way, planemo can be upgraded as follows:
 
 ::
 
-    $ . .venv/bin/activate
+    $ source planemo/bin/activate
```
nsoranzo, line 22:
```suggestion
    $ . planemo/bin/activate
```

mblue9:
Ok great, thanks for the info @nsoranzo !

bgruening:
Thanks @mblue9!

mblue9:
Great! (thanks @rlupat for the tip about python venv)

Issue #996, opened 2020-02-04
planemo 0.70 traverses complete subtree even if tools are listed


bernt-matthias:
I like to run `planemo test A.xml B.xml ...`  because sometimes the tool directory is not clean, i.e. there is a lot of other things besides tools, test-data, etc. 

Used to work until 0.62.1. 

Now planemo needs a large amount of time until anything happens (strace indicates that it tries to copy to tmp) and finally stumbles over a broken symlink in my case. 

```
Traceback (most recent call last):
  File "/home/berntm/.venv3/bin/planemo", line 8, in <module>
    sys.exit(planemo())
  File "/home/berntm/.venv3/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/berntm/.venv3/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/berntm/.venv3/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/berntm/.venv3/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/berntm/.venv3/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/berntm/.venv3/lib/python3.6/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/home/berntm/.venv3/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/berntm/.venv3/lib/python3.6/site-packages/planemo/cli.py", line 190, in handle_blended_options
    return f(*args, **kwds)
  File "/home/berntm/.venv3/lib/python3.6/site-packages/planemo/commands/cmd_test.py", line 80, in cli
    runnables = for_paths(paths, temp_path=temp_path)
  File "/home/berntm/.venv3/lib/python3.6/site-packages/planemo/runnable.py", line 149, in for_paths
    return [for_path(path, temp_path=temp_path) for path in paths]
  File "/home/berntm/.venv3/lib/python3.6/site-packages/planemo/runnable.py", line 149, in <listcomp>
    return [for_path(path, temp_path=temp_path) for path in paths]
[...]

PR #1037, opened 2020-07-24
Only copy files if they don't exist

mvdbeek:
I think that should make https://github.com/galaxyproject/planemo/issues/996 much less of an issue.

https://docs.python.org/2.5/dist/module-distutils.fileutil.html describes the `update` argument.


mvdbeek:
Not insisting this is the right way to do it at all, but this is a simple improvement on the current situation. Maybe staging inputs via pulsar embedded is the way to go ? Or maybe we can improve on this and somehow only copy the test-data folder ?

PR #1038, opened 2020-07-26
Bump minimum bioblend to 0.14.0

mvdbeek:
Otherwise workflow testing fails wih
```
'GalaxyInstance' object has no attribute '_make_url'
```


bgruening:
Thanks @mvdbeek!

Issue #920, opened 2019-04-16
Deprecate/remove `conda_lint` and `bioc_conda_recipe_init` commands


nsoranzo:
Planemo is not (and probably cannot) keep up with conda recipe best practices.

jmchilton:
🥲 - but I get it - and implemented already.

PR #1043, opened 2020-07-27
Remove older, likely unused commands.

jmchilton:
Probably a bunch of supporting code that needs to be removed also - but lets see if anyone objects to any of these being removed first.


nsoranzo:
In https://github.com/galaxyproject/planemo/issues/920 I suggested to also remove `conda_lint` .

jmchilton:
Let me remove these others ones I'm more confident about first - that will be at the top of the next round of cuts 😅.

PR #1047, opened 2020-08-01
Fixes to run tool_factory command.

fubar2:
fubar2/toolfactory now passes tests using planemo - it was rebuilt with galaxyxml and is substantially simplified. Here are some changes including an updated planemo_ext/tool_factory_2 and some setup tweaks so it seems to get installed. I can't get tox tests all working on my fork. Dunno why. 
Also a suggested pr.yml to only run a test when a pr happens. ci.yml now only runs on push and distributes the package. This way you can let PR's run a test before even looking at them?
cheers...
ross


fubar2:
Temporary halt - seeing strange gremlins I need to chase down.....

jmchilton:
I removed the tool factory in #1043 because I thought it was dead but I'm excited to see life! Happy to re-add it if you get those gremlins solved - thanks so much for working on this @fubar2.

fubar2:
Hi, John,

I've had 4 years of retirement and not paying attention - so I feel like I just woke up after oversleeping for 4 years. I did enjoy BCC2020 and that's how I kind of got re-motivated.

Yikes! Lots to learn - but when I noticed that you'd included it in the planemo source tree, I felt so bad that I'd ignored it for so long that I'm determined to resurrect it. Lazarus indeed... :)

It does work but I'll do some more testing before the PR again...

PR #1050, opened 2020-08-03
Make splitting job and test files optional

mvdbeek:
I think that makes writing tests easier and I think @bgruening agrees (https://github.com/galaxyproject/training-material/pull/2021#discussion_r464233812)

### DIFF
--- a/planemo/commands/cmd_workflow_test_init.py
+++ b/planemo/commands/cmd_workflow_test_init.py
@@ -17,8 +17,9 @@ from planemo.galaxy.workflows import (
 @options.required_workflow_arg()
 @options.force_option()
 @options.workflow_output_artifact()
+@options.split_job_and_test()
 @command_function
-def cli(ctx, workflow_path, output=None, force=False, **kwds):
+def cli(ctx, workflow_path, output=None, force=False, split_test=False, **kwds):
     """Initialize a Galaxy workflow test description for supplied workflow.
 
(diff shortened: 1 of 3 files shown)


jmchilton:
I'm totally fine hiding this behind a flag and I think I generally agree the single file version is simpler and should perhaps be best practice. I had this https://github.com/galaxyproject/planemo/pull/1052 in mind when I implemented this - I wanted to build some more support and documentation running workflows via Planemo. In that context, the standalone jobs are more useful as examples. We don't need to complicate testing because of my side quest though.

mvdbeek:
Sounds like `planemo run` could also (optionally ?) output a runnable script like workflow2executable does ... that would be some nice convergence.

jmchilton:
I tried to rebase your branch in https://github.com/galaxyproject/planemo/pull/1064/files but somehow it lost the commit author information 😢.

mvdbeek:
No worries, that's fine.

bgruening:
> Sounds like planemo run could also (optionally ?) output a runnable script like workflow2executable does ... that would be some nice convergence.

That would be great! However, not sure this should be part of planemo run. How do we structure all the workflow-run functionality? It would be super great to look up the invocations, get a link to the reports, restart all failing jobs in a workflow. @simonbray has some functionality in his gxwf project. But we are struggling with how we can fit this back into planmo from the CLI point of view.

PR #1051, opened 2020-08-03
Implement stubs for outputs

mvdbeek:
Would be awesome if we could figure out the class (File, Collection,
Parameter) and more details about each, but that seems like more than we
can do without having the tools available.

### DIFF
--- a/planemo/commands/cmd_workflow_test_init.py
+++ b/planemo/commands/cmd_workflow_test_init.py
@@ -6,7 +6,11 @@ import yaml
 
 from planemo import options
 from planemo.cli import command_function
-from planemo.galaxy.workflows import job_template, new_workflow_associated_path
+from planemo.galaxy.workflows import (
+    job_template,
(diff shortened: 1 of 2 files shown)


jmchilton:
I suspect if you think this is a good idea, you're right and this is probably the right thing to do - there was a weird psychology thing that held me back. I think if people have to delete a line because they don't have a test for a particular output they will feel like ... we're judging them and they're doing something against best practices - and so workflow testing will feel heavier than it needs to be. I was thinking if they were commented out lines people would feel empowered as they added tests for each output they wanted to instead of defeated as they removed tests they couldn't right. I was planning to take another pass where I did that - maybe by treating the file as text and adding comments hackily or using ruamel and adding comment objects. Would you object to me doing that on top of this?

mvdbeek:
Absolutely not, I also wanted to do this, but as you noticed it's either hacky or we need to invest a bit more effort to nicely embed comments in the dumped yaml.

PR #1052, opened 2020-08-03
More documentation/support around running workflows.

jmchilton:
Nowhere near done but I want to discuss ``planemo run`` with workflows - including from the command line and using Planemo as a library. Against running servers and with Planemo managed ones, against dockerized Galaxy, against postgres, with CVMFS, etc... To incentivize people to use ``planemo run`` I added ``planemo workflow_job_init target.ga`` to build job files.

Also fixes ``--force`` for ``workflow_test_init``.

### DIFF
--- a/docs/automating_workflows.rst
+++ b/docs/automating_workflows.rst
@@ -1,0 +1,31 @@
+Automating Galaxy Workflows
+=============================
+
+The BioBlend_ library can be used to invoke and monitor Galaxy workflows.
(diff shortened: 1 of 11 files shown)


jmchilton:
The docs aren't even close to being done so I won't link them in - the rest of these fixes though are good so I'm going to merge this. Also I added docs to the test format page about how to test composite inputs and nested collection inputs and these are really important.

PR #1057, opened 2020-08-03
Add test case for non-data parameters.

jmchilton:


tests/data/wf9-int-input.gxwf.yml
```diff
@@ -0,0 +1,20 @@
+class: GalaxyWorkflow
+name: "Test Workflow with integer parameter"
+
+inputs:
+  lines:
+    type: integer
```
bgruening, line 6: @jmchilton are those types case-sensitive? `integer` vs `File`

jmchilton: I believe so according to the schema http://galaxyproject.github.io/gxformat2/v19_09.html#GalaxyType - I’m sure there is a good reason they have to have different casings (base type coming from jsonld and File being an extension defined in CWL’s schema salad meta schema - or something like that). workflow_lint should catch spelling issues in the workflow and I’m going to write some input job linting that will check types and such soon.

bgruening: Mh, ok. Its somehow confusing. But if gxformat2 will catch it! Top!

PR #1060, opened 2020-08-04
Add mypy type checking.

jmchilton:
Various code cleanups (dropping Python 2 compat. mostly) to be enable this.

### DIFF
--- a/planemo/runnable.py
+++ b/planemo/runnable.py
@@ -6,7 +6,7 @@ import abc
 import collections
 import os
 from distutils.dir_util import copy_tree
-from enum import Enum
+from enum import auto, Enum
 
 import yaml
 from galaxy.tool_util.cwl.parser import workflow_proxy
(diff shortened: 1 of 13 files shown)


bgruening:
:1st_place_medal: this is awesome. @jmchilton can you outline how you want to proceed with mypy maybe we can help.

jmchilton:
I don't know 🤷‍♂️ - this isn't a real priority just something I was hacking on between rounds of fighting with CI on actually important PRs.

I think what is here right now is okay and will catch a few errors here and there and will encourage best practices to get us ready for typing. If I had to guess on another big next task it would probably be to include type annotations with galaxy-util, galaxy-tool-util, and bioblend. I had to disable import checking because a variety of libraries don't support it - so starting at the bottom of the dependency chain and working up would probably be the next step?

planemo/runnable.py
```diff
@@ -6,8 +6,8 @@
 import collections
 import os
 from distutils.dir_util import copy_tree
+from enum import auto, Enum
```
nsoranzo, line 9: @jmchilton `enum.auto` is available only on Python >=3.6 (in fact CircleCI builds are broken). Should we remove Python 3.5 support? It reached EOL on 2020-09-30.

jmchilton: Yeah - go for it!

nsoranzo: Addressed in https://github.com/galaxyproject/planemo/pull/1099

PR #1064, opened 2020-08-05
Make splitting job and test files optional

jmchilton:
I think that makes writing tests easier and I think @bgruening agrees (https://github.com/galaxyproject/training-material/pull/2021#discussion_r464233812)

### DIFF
--- a/planemo/commands/cmd_workflow_test_init.py
+++ b/planemo/commands/cmd_workflow_test_init.py
@@ -18,29 +18,35 @@ from planemo.io import can_write_to_path
 @options.required_workflow_arg()
 @options.force_option()
 @options.workflow_output_artifact()
+@options.split_job_and_test()
 @command_function
-def cli(ctx, workflow_path, output=None, **kwds):
+def cli(ctx, workflow_path, output=None, split_test=False, **kwds):
     """Initialize a Galaxy workflow test description for supplied workflow.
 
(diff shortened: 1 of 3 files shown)


planemo/commands/cmd_workflow_test_init.py
```diff
@@ -18,29 +18,35 @@
...
          'doc': 'Test outline for %s' % path_basename,
-         'job': os.path.basename(job_output),
-         'outputs': output_stubs_for_workflow(workflow_path)
+         'job': job,
+         'outputs': output_stubs_for_workflow(workflow_path),
     }]
+    if split_test:
+        job_output = new_workflow_associated_path(workflow_path, suffix="job1")
+        if not can_write_to_path(job_output, **kwds):
+            ctx.exit(1)
```
mvdbeek, line 46: Does that print some information about why this failed ?

jmchilton: I think so:

``error("%s already exists, exiting." % path)``

jmchilton: It'd be good to also check the directory exists and is writable... but yeah...

bgruening:
Indeed, I think as the default behavior this is better. For power users, the split one is probably best. I like it! Thanks both!

PR #1065, opened 2020-08-07
Add autoupdate command

simonbray:
A command for updating the requirements in XML tool files by checking for newer conda versions. Thanks to @lorrainealisha75 for working on this.

ping @mvdbeek, in this version a `@GALAXY_VERSION@` token should no longer be necessary.

ping @wm75, I added a `--dry-run` flag as you suggested.


planemo/commands/cmd_autoupdate.py
```diff
@@ -0,0 +1,64 @@
...
+    is_tool_load_error,
+    yield_tool_sources_on_paths
+)
+
+from planemo import options, autoupdate
+from planemo.cli import command_function
+from planemo.config import planemo_option
+
+
+def dry_run_option():
```
simonbray, line 22: not sure if this should be moved to `options.py` or not

jmchilton: It is totally fine here, especially if only one command is using it.

planemo/autoupdate.py
```diff
@@ -0,0 +1,131 @@
...
+            requirements['main_req'] = requirement.text
+        else:
+            requirements['other_reqs'][requirement.text] = requirement.attrib.get('version')
+    return requirements
+
+
+def get_latest_versions(version_dict):
+    """
+    Update a dict with current conda versions for tool requirements
+    """
```
jmchilton, line 71: This is pretty awesome!

bgruening:
@simonbray some updated to the docs would be great. We can also use this in the IUC discussion then. Thanks.

simonbray:
@bgruening where should these be added? It seems the docs are autogenerated from the code.

planemo/autoupdate.py
simonbray: Thank you @jmchilton - I think this is ready for a review

bgruening:
@simonbray a similar page to this one maybe? https://github.com/galaxyproject/planemo/blob/master/docs/publishing.rst

We can also wait until we have a CI setup ready and then add documentation for a potential CI setup as well?

planemo/autoupdate.py
simonbray: Actually, I take that back, there are still some things I can fix. Though any overall comments about the implementation would be welcome.

simonbray:
I ran it on the tools-iuc repository to test it out, have a look here: https://github.com/galaxyproject/tools-iuc/compare/master...simonbray:autoupdate

It looks mostly okay, but there are a few bugs (e.g. https://github.com/galaxyproject/tools-iuc/compare/master...simonbray:autoupdate#diff-937fe802b3fe3877efcbd5e6a1f6e36eL2 which I'll take a look at).

simonbray:
Testing of latest version on tools-iuc repo: https://github.com/galaxyproject/tools-iuc/compare/master...simonbray:autoupdate8

simonbray:
https://github.com/galaxyproject/tools-iuc/compare/master...simonbray:autoupdate9

simonbray:
https://github.com/galaxyproject/tools-iuc/compare/master...simonbray:autoupdate10

bgruening:
@simonbray looking at the example commits, I think we should skip pure `python` or `perl` updates?

simonbray:
@bgruening do you have an example where this is happening? The non-main requirements should only be updated when a new version of the main requirement is detected.

simonbray:
Or you mean that there should be an option to specify a list of packages that never get updated? (with `python,perl,r` maybe as default)

bgruening:
I was referring to this one: https://github.com/galaxyproject/tools-iuc/compare/master...simonbray:autoupdate10#diff-f2f4ef5baa9e8c4a1dd72d8c1a9acb73R10

But I see the version bump happened somewhere else. Cool. I think this is exactly how it should be! Thanks a lot, that looks super cool.

simonbray:
Random idea that I want to write down before I forget it: this command should be able to update workflows as well as tools.

bgruening:
@jmchilton what do you think about this? In particular in regards to your recent "update workflow" works?

@simonbray it would be nice to have this in planemo soon so we can finally add this functionality into IUC and close the loop: 

`upstream-github-release` → `auto-conda-bump` → `auto-biocontainer` → `auto-Galaxy-tool` → `auto-mulled-container` → `auto-tool-installation-from-TS` → `auto-tool-testing` → `auto-workflow-bump` → `auto-workflow-testing`

jmchilton:
Seems like a very solid plan - it will be interesting to see what types of errors we can catch at each step. Let me know if there is anything I can do to help.

simonbray:
> @simonbray it would be nice to have this in planemo soon so we can finally add this functionality into IUC and close the loop

I think this is probably quite close to merging - I will have another look at it.

@lorrainealisha75 has also been working on implementing a CI bot here: https://github.com/planemo-autoupdate/autoupdate/blob/main/.github/workflows/update.yml

Where is the code for workflow autoupdates, is this part of planemo or somewhere else? I had the idea this could be done using the same `autoupdate` subcommand (similar to `run`, `test` etc. working with both tools and workflows). But I suppose this isn't really necessary.

bgruening:
@simonbray please have a look at https://github.com/galaxyproject/galaxy/pull/11032 and related PRs. There are new APIs that can upgrade workflows and can upgrade tools, report on changed params or dropped connections etc ...

simonbray:
Nice! So if `autoupdate` got run on a workflow it could simply load it onto the chosen Galaxy (either local or external) and call this `refactor` API.

bgruening:
That is my understanding of the jmchilton mgic :)

docs/autoupdate.rst
```diff
@@ -13,17 +13,17 @@ Planemo provides a ``autoupdate`` subcommand which can be used to perform this t
...
 
 Formatting tools
 =============================
 
- ``autoupdate`` assumes that XML tools are formatted in a certain way - in accordance with the `IUC best practices`_
-
-<https://planemo.readthedocs.io/en/latest/standards/docs/best_practices/tool_xml.html/>`__. In particular:
+``autoupdate`` assumes that XML tools are formatted in a certain way - in accordance with the `IUC best practices`_. In particular:
 
   - the tool ``version`` attribute must be set to ``@TOOL_VERSION@+galaxy0`` or ``@TOOL_VERSION@+galaxy@GALAXY_VERSION``
```
davelopez, line 28:
```suggestion
  - the tool ``version`` attribute must be set to ``@TOOL_VERSION@+galaxy0`` or ``@TOOL_VERSION@+galaxy@GALAXY_VERSION@``
```
;)

By the way, should this be @VERSION_SUFFIX@ instead of @GALAXY_VERSION@? The latest documentation says it should be @VERSION_SUFFIX@ but it seems a significant amount of tools use @GALAXY_VERSION@

simonbray: Yes you are right - we decided on `@VERSION_SUFFIX@` in the end but I didn't update here. Thanks!

nsoranzo: It's `@VERSION_SUFFIX@` , @GALAXY_VERSION@ was the whole `+galaxyN` .

simonbray:
Not sure about the failing test - it doesn't seem related and I can't reproduce it locally: https://travis-ci.org/github/galaxyproject/planemo/jobs/763104995

simonbray:
@jmchilton maybe you could have a look at this when you have time?

jmchilton:
I don't love that you're custom parsing XML outside the abstractions established in ``galaxy.tool_util.parser`` but you're also rewriting the XML explicitly so this wouldn't work with other future tool implementations, etc.. so I think this is ultimately totally fine in this case.

Really awesome work, thanks so much!

simonbray:
> I don't love that you're custom parsing XML outside the abstractions established in galaxy.tool_util.parser

I definitely don't love it either but because of the need to rewrite the wrapper I'm not sure there's a good alternative.

Either way, thanks a lot for the merge!

Issue #1071, opened 2020-08-26
planemo run fails with "Provided API key is not valid."


simonbray:
I am just testing with this workflow: https://github.com/usegalaxy-eu/workflow-testing/tree/master/example1

Running `planemo run wf3-shed-tools.ga wf3-shed-tools-job.yml --galaxy_url https://usegalaxy.eu/ --galaxy_user_key APIKEY --engine external_galaxy` gives:

```
bioblend ERROR: GET: error 403: b'{"err_msg": "Provided API key is not valid.", "err_code": 403001}', 0 attempts left
Traceback (most recent call last):
  File "/home/simon/Repos/.planemo_gh/bin/planemo", line 8, in <module>
    sys.exit(planemo())
  File "/home/simon/Repos/.planemo_gh/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/simon/Repos/.planemo_gh/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/simon/Repos/.planemo_gh/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/simon/Repos/.planemo_gh/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/simon/Repos/.planemo_gh/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/simon/Repos/.planemo_gh/lib/python3.7/site-packages/click/decorators.py", line 73, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/home/simon/Repos/.planemo_gh/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/simon/Repos/.planemo_gh/lib/python3.7/site-packages/planemo/cli.py", line 98, in handle_blended_options
    return f(*args, **kwds)
  File "/home/simon/Repos/.planemo_gh/lib/python3.7/site-packages/planemo/commands/cmd_run.py", line 41, in cli
    run_result = engine.run(path, job_path)
  File "/home/simon/Repos/.planemo_gh/lib/python3.7/site-packages/planemo/engine/interface.py", line 60, in run
[...]

PR #1072, opened 2020-08-28
Fixes for CLI when external_galaxy is used as the engine

simonbray:
Two minor changes:
* when `--galaxy_url` is specified, the engine should be `external_galaxy` by default without this having to be additionally specified.
* when tools/workflows are run on an external galaxy, planemo should not attempt to install shed repositories (at the very least, not by default). See #1071.

### DIFF
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -761,7 +761,7 @@ class BaseGalaxyConfig(GalaxyInterface):
                 self._install_workflow(runnable)
 
     def _install_workflow(self, runnable):
-        if self._kwds.get("shed_install") and self._kwds.get("engine") != "external_galaxy":
+        if self._kwds.get("shed_install") and (self._kwds.get("engine") != "external_galaxy" or self._kwds.get("galaxy_admin_key")):
             install_shed_repos(runnable,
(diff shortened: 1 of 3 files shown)


planemo/galaxy/config.py
```diff
@@ -761,7 +761,7 @@ def install_workflows(self):
                 self._install_workflow(runnable)
 
     def _install_workflow(self, runnable):
-        if self._kwds.get("shed_install"):
+        if self._kwds.get("shed_install") and self._kwds.get("engine") != "external_galaxy":
```
simonbray, line 764: or maybe something like: `(self._kwds.get("engine") != "external_galaxy" and not self._kwds.get("galaxy_admin_key"))`

so it would still be possible to run shed_install on an external galaxy instance, as long as the admin key was specified.

bgruening: > so it would still be possible to run shed_install on an external galaxy instance, as long as the admin key was specified.

I think this is an important use-case, e.g. for a Galaxy test server. Or a training instance that will be setup.

jmchilton:
Thanks for the fixes - these seem great!

Issue #1070, opened 2020-08-25
using planemo test --update_test_data does produce the output files but the test fails  


fubar2:
Has #846 come back?

I see the same reported behaviour with a fresh Galaxy dev checkout and planemo 0.72.0.dev0

When I remove a test output and run test --update_test_data, planemo _**does**_ correctly recreate the missing output but the test fails with:

`Test file (outfile_sample) is missing. If you use planemo try --update_test_data to generate one.`

It is indeed **not** really passing the test if there's a missing output so this needs some thought.

Perhaps planemo should have an unforgiving "test" that does exactly that and fails if it fails....as distinct from a different command to execute the specified test and generate fresh appropriate outputs, then test properly? Once generated, those fresh outputs are then ideal for the *real* pass/fail test. In generating them there has certainly been a test, but it seems odd to report that process as successful - I think I'll want to re-test with the generated output before declaring victory.

Here's the traceback:

`
FAIL: ( pyrevargparse ) > Test-1
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ross/galaxy/test/functional/test_toolbox.py", line 99, in test_tool
    self.do_it(tool_version=tool_version, test_index=test_index)
  File "/home/ross/galaxy/test/functional/test_toolbox.py", line 36, in do_it
    verify_tool(tool_id, self.galaxy_interactor, resource_parameters=resource_parameters, test_index=test_index, tool_version=tool_version, register_job_data=register_job_data)
  File "/home/ross/galaxy/lib/galaxy/tool_util/verify/interactor.py", line 834, in verify_tool
    raise e
  File "/home/ross/galaxy/lib/galaxy/tool_util/verify/interactor.py", line 830, in verify_tool
    job_stdio = _verify_outputs(testdef, test_history, jobs, tool_id, data_list, data_collection_list, galaxy_interactor, quiet=quiet)
  File "/home/ross/galaxy/lib/galaxy/tool_util/verify/interactor.py", line 1005, in _verify_outputs
    raise JobOutputsError(found_exceptions, job_stdio)
galaxy.tool_util.verify.interactor.JobOutputsError: Test file (outfile_sample) is missing. If you use planemo try --update_test_data to generate one.
`

simonbray:
[...]

PR #1079, opened 2020-09-14
Changes to update_test_data testing mode

simonbray:
I tried to improve the 'update test data' mode to make it a bit more user-friendly, see https://github.com/galaxyproject/planemo/issues/1070. Ping @fubar2.

The currently behaviour is that the `--update_test_data` flag just causes the job_output_files dir to be copied back to the test dir. So the test results confusingly do not actually relate to the new updated data at all. Furthermore, if tests fail, the command needs to be rerun to check that they work with the new data.

New behaviour in this PR when the `--update_test_data` flag is used:
* if tests fail, tests are automatically rerun after the test data has been updated, so that the results the user gets at the end relate to the new data, as expected. The user is  informed that `Test data were updated and tests were rerun.`
* if the tests pass, the test data are not updated and the user is explicitly informed that `All x test(s) executed passed. Therefore, no test data were updated.`


bgruening:
:+1: I guess this will also be useful for automatic update PRs.

fubar2:
Nice one, thanks @simonbray - that seems like a logical fix to me - will take a look shortly.

PR #1091, opened 2020-10-05
Add workflow_upload command

mvdbeek:
`planemo workflow_upload` will update or create a github repository with the contents of a directory that contains a workflow.
Every upload will create a new release. Workflow versions are taken from the workflow files.

Also moves from `hub` to `gh` and makes `planemo ci_find_repos` return directories with workflows.

TODO:
- [x] Check if version already exists / handle empty diff ?
- [ ] `--force` that will override existing version
- [x] Handle deleted files (remove everything in work dir, then copy ?)

### DIFF
--- a/planemo/ci.py
+++ b/planemo/ci.py
@@ -11,6 +11,9 @@ import yaml
 from planemo import git
 from planemo import io
 from planemo.shed import SHED_CONFIG_NAME
+from planemo.workflow_lint import DOCKSTORE_REGISTRY_CONF
+
+REPO_METADATA_FILES = (SHED_CONFIG_NAME, DOCKSTORE_REGISTRY_CONF)
 
 
 def filter_paths(ctx, raw_paths, path_type="repo", **kwds):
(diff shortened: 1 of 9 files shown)


planemo/ci.py
```diff
@@ -29,21 +32,16 @@ def filter_paths(ctx, raw_paths, path_type="repo", **kwds):
...
+                diff_path = metadata_file_in_path(diff_dir)
+                if diff_path:
+                    diff_paths.add(diff_path)
+                    break
         else:
             diff_paths = diff_files
 
     unique_paths = set(os.path.relpath(p, cwd) for p in raw_paths)
     if diff_paths is not None:
-        new_unique_paths = []
```
mvdbeek, line 42: I don't think I have understood the purpose here. Was this meant to filter our paths in `diff_paths` that are not within `raw_paths` ?

mvdbeek:
I think this is ready for review!

jmchilton (approved):

jmchilton:
Simply amazing!

bgruening:
Wonderful!

planemo/ci.py
nsoranzo: This was (and should remain) the intersection of `unique_paths` (i.e. the tool or repo specified by path by the user) and `diff_paths` (the list of files or directories changed in Git).
Sorry I missed this at the time, should be fixed by https://github.com/galaxyproject/planemo/pull/1127 .

PR #1093, opened 2020-10-15
Explicit tests for Galaxy 20.09.

jmchilton:


jmchilton:
Release new Galaxy packages today resulted in https://github.com/galaxyproject/galaxy/pull/9684/files breaking Planemo. I'll see what I can do.

PR #1095, opened 2020-11-02
Allow testing dir of workflows

mvdbeek:


planemo/engine/interface.py
```diff
@@ -54,9 +54,10 @@ def cleanup(self):
 
     def run(self, runnable, job_path):
         """Run a job using a compatible artifact (workflow or tool)."""
-        self._check_can_run(runnable)
-        run_response = self._run(runnable, job_path)
-        return run_response
+        # TODO where is path comining from?
+        for runnable in for_path(path):
```
bernt-matthias, line 58: Neither `path` nor `for_path` are defined here. Can you give me a few hints how to solve this?

planemo/commands/cmd_workflow_edit.py
```diff
@@ -23,8 +23,10 @@ def cli(ctx, workflow_identifier, output=None, force=False, **kwds):
 
     kwds["workflows_from_path"] = True
 
+    # TODO this foo seems unimported (also not sure if works when called with id)
+    runnables = for_path(workflow_identifier)
```
bernt-matthias, line 27: Before my merge `for_path` (which misses an import I guess) was called on `workflow_path`, but recently all `workflow_path` were replaced by `workflow_identifier` .. so this is likely broken with my commit.

bernt-matthias (commented):
Just a few comments on my attempt to merge.

planemo/commands/cmd_workflow_edit.py
mvdbeek: Ah, yeah, this was changed even before that. I think I found a less intrusive way to do this now.

planemo/workflow_lint.py
```diff
@@ -76,19 +76,22 @@ def structure(path, lint_context):
...
+        else:
+            all_tests_valid = True
+        for test_case in test_cases:
+            if not _lint_case(path, test_case, lint_context):
+                all_tests_valid = False
 
-    if all_tests_valid:
-        lint_context.valid("Tests appear structurally correct")
+        if all_tests_valid:
+            lint_context.valid("Tests appear structurally correct")
```
simonbray, line 94:
```suggestion
            lint_context.valid(f"Tests appear structurally correct for {runnable.path}")
```

simonbray (approved):

jmchilton (approved):

PR #1099, opened 2020-11-23
Fix ``run_tests.sh`` invocation

nsoranzo:
Also:
- Drop support for Python 3.5
- Add support for Python 3.9
- Rebuild docs


nsoranzo:
Depends on https://github.com/kislyuk/argcomplete/pull/332 and a new argcomplete release on PyPI.

nsoranzo:
Ready for review!

BTW, it looks to me that all test jobs in `.circleci/config.yml` are already run in GitHub workflows, can we drop the file? @jmchilton

mvdbeek (approved):

jmchilton:
There are a bunch of overlapping parameters to get more of these buggy tests working - I'd very much like to expand test coverage (more tests, more versions of Python, more versions of Galaxy, etc..) so I was working on trying to get those circleci tests working. I wouldn't delete the file - I'd like to keep working on it.

jmchilton:
Thanks so much for the fix!

PR #1100, opened 2020-11-26
sort failing tests to top first

bernt-matthias:
because for huge reports it gets annoying to spot the failing tests.

### DIFF
--- a/planemo/galaxy/test/actions.py
+++ b/planemo/galaxy/test/actions.py
@@ -152,7 +152,7 @@ def merge_reports(input_paths, output_path):
     tests = []
     for report in reports:
         tests.extend(report["tests"])
-    tests = sorted(tests, key=lambda k: k['id'])
+    tests = sorted(tests, key=lambda k: (k['data']['status'], k['id']))
     merged_report = {"tests": tests}
     with io.open(output_path, mode="w", encoding='utf-8') as out:
         out.write(unicodify(json.dumps(merged_report)))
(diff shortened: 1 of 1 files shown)


mvdbeek:
Have you tried planemo from the master branch already ?

bernt-matthias:
> Have you tried planemo from the master branch already ?

No. Do you think that something changed in this matter.

bernt-matthias:
Just checked. `merge_test_reports` does not sort failing tests to the top. Even if `test` sorts the output (which would also be handy) this would still be necessary here. 

But maybe sorting would be even more appropriate here: https://github.com/galaxyproject/planemo/blob/3201c415a5d637984620d86b07bb9407d33833b5/planemo/reports/build_report.py#L20

(maybe also for other report type) then also `test` would be covered.

bernt-matthias:
like so. we would just need to decide if we want to treat all outputs and the summary this way. In particular for the summary I do not particularly like it.

mvdbeek:
have you looked at the html test report ?

bernt-matthias:
I see.. nice.. even if I needed a while until I discovered that `Errored Tests` and `Passed Tests` can be clicked to be unfolded.

mvdbeek:
Ah, that's a CSS problem I just haven't had the time to look at. Really should do that though.

mvdbeek:
(It looks better on chrome)

PR #1101, opened 2020-12-02
Some fixes for workflow_job_init and workflow_test_init

simonbray:
- Recognize the input types as specified by gxformat2 here: https://github.com/galaxyproject/gxformat2/blob/master/gxformat2/model.py#L66
- Add parameters
- Add a more helpful error message for unlinted workflows

### DIFF
--- a/planemo/galaxy/workflows.py
+++ b/planemo/galaxy/workflows.py
@@ -153,7 +153,10 @@ def input_labels(workflow_path):
 
 
 def required_input_steps(workflow_path):
-    steps = inputs_normalized(workflow_path=workflow_path)
+    try:
+        steps = inputs_normalized(workflow_path=workflow_path)
+    except Exception:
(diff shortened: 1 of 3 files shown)


bgruening:
Can you please add a test, so we do not regress here anymore? Thanks.

simonbray:
> Can you please add a test, so we do not regress here anymore? Thanks.

There is a test already, but you are right, I need to make it stricter.

Also, these commands don't work yet with workflow IDs, it would be good to get that working.

simonbray:
> Also, these commands don't work yet with workflow IDs, it would be good to get that working.

I will leave this for now.

jmchilton:
Sorry for missing this! Feel free to ping me on these Planemo PRs for potentially quicker turn around.

simonbray:
No worries, thanks for the merge!

PR #1076, opened 2020-09-04
Integrate some features from gxwf 

simonbray:
Start working on integrating some features from gxwf (see https://github.com/simonbray/gxwf):
- some work developing the profile commands, in particular to allow creation of profiles for external galaxies
- allow execution of workflows on an external galaxy instance with only a workflow ID
- allow creation of aliases which can be used instead of workflow (eventually also dataset) IDs
- a list_invocations command
- various other small fixes for things I found along the way

PR #1102, opened 2020-12-03
Some documentation for planemo run

simonbray:
Particularly focussing on the new features in https://github.com/galaxyproject/planemo/pull/1076.

It would be really appreciated if a few people who are informed about planemo could scan through the headings and suggest any other things I could/should include; quite likely there is some functionality I'm not aware of.


docs/_running_external.rst
```diff
@@ -0,0 +1,223 @@
...
+
+
+Using aliases
+===============================================
+
+Once you are dealing with a large number of workflows and datasets, you may
+find that it becomes difficult to keep track of the file paths or IDs
+which you are using for execution, particularly if you are executing workflows
+based on their ID. Planemo offers the option to create aliases, or easily
+remembrable mnemonics, for Galaxy workflows, with the following command:
```
mtekman, line 186:
```suggestion
rememberable mnemonics, for Galaxy workflows, with the following command:
```

docs/_running_external.rst
```diff
@@ -0,0 +1,223 @@
...
+
+    GSE37268_mof3:
+      class: File
+      path: todo_test_data_path.ext
+    Genes:
+      class: File
+      path: todo_test_data_path.ext
+
+
+For each of the specified inputs in the workflow, a entry is created in the
```
anuprulez, line 82: `an entry`

docs/_running_external.rst
simonbray: `memorable` I suppose would be better than either :laughing:

docs/_running_external.rst
```diff
@@ -0,0 +1,223 @@
...
+
+::
+
+    input_dataset:
+      class: File
+      path: todo_test_data_path.ext
+    input_collection:
+      class: Collection
+      collection_type: list
+      elements:
```
bgruening, line 99: can we add here a second element to make it crystal clear its a list

docs/_running_external.rst
```diff
@@ -0,0 +1,223 @@
...
+workflow and the input datasets get newly uploaded at each execution. If you
+want to run the same workflow multiple times, you may prefer to avoid this.
+In the examples given so far, all workflows and datasets are specified by means
+of a local path, but Planemo also allows you to use the IDs created by Galaxy
+as well. These IDs are unique to each Galaxy server, so this approach isn't
+transferrable if you want to run your workflows on multiple servers.
+
+The first step is to ensure all the datasets which are required for the
+workflow are already uploaded. You can either do this by running the workflow
+once in the normal way, as described above, or just manually uploading through
```
bgruening, line 123: Can you please create a planemo issue about this one? I guess a planemo upload_data command might be useful that returns the IDs?

docs/_running_external.rst
```diff
@@ -0,0 +1,223 @@
...
+    input_collection2:
+      class: Collection
+      collection_type: list
+      elements:
+      - class: File
+        identifier: todo_element_name
+        galaxy_id: "457d46215431cc37baf96108ad87f351"
+
+
+In the first example, an existing collection will be used (by specifying its
```
bgruening, line 162: first example might be confusing, maybe saying explicitly int the input_collection2 example  ... and maybe this should be collection1?

docs/_running_external.rst
```diff
@@ -0,0 +1,223 @@
...
+    usegalaxy-eu
+    usegalaxy-org
+    3 configured profiles are available.
+
+
+Once the new ``tutorial_profile`` is created, a workflow can be executed with:
+
+::
+
+    $ planemo run wf3-shed-tools.ga wf3-shed-tools-job.yml --profile tutorial_profile
```
bgruening, line 54: the next logical step would be to define a default_profile ... a profile that is used when no param is used and not profile. In the profile_list command, it could be marked with a * - pretty much like git branches. If you like this idea, could you please create an issue for it.

bgruening (commented):
Great tutorial @simonbray a few comments inline.

I think it would also be useful to show an example planemo conf at some point to demonstrate what happens if a profile or alias is created. The text file I mean.

docs/_running_external.rst
```diff
@@ -0,0 +1,255 @@
...
+collection ID), whereas for ``input_collection2``, a new collection will be created from an
+existing dataset.
+
+Once the job file has been modified, run ``planemo run`` as before. The result
+should be the same, though it should be a bit faster, since the upload step was
+skipped. Instead, the selected datasets get copied to a new history, which
+unlike a new upload, doesn't result in any additional storage being used.
+
+To run the workflow using a workflow ID, replace the workflow file path with
+the workflow ID from the Galaxy server:
```
simonbray, line 175: I am kind of skating over how to get this - if https://github.com/galaxyproject/galaxy/issues/10848 can be added I will write another sentence here.

bgruening: We could for the time being use the ID that is displayed when the workflow editor is open, do we?

bgruening:
> _running_external.rst:248: WARNING: Literal block ends without a blank line; unexpected unindent.

That seems to be a legitimate warning.

docs/_running_external.rst
simonbray: Not sure what you mean here? That it is displayed in the URL? I can't see it displayed on the page anywhere.

bgruening: https://usegalaxy.eu/workflow/editor?id=adb3676c3450a227 --> this is what I see when the workflow editor is open. And I assume this is the workflow-id.

simonbray: Yes, it is. I would prefer if the user could get it some better way than copying out of the URL, that's why I made the issue above. But yes, if this cannot be fixed in the near future, I'll mention this alternative here.

simonbray:
I think this is ready, though it would be great if https://github.com/galaxyproject/planemo/pull/1101 were merged first.

docs/_running_external.rst
bgruening: I guess we can mention this here and go for it until we have a better way.

jmchilton:
Amazing - thanks so much!

PR #1106, opened 2020-12-06
consider exclude in shed update

### LINKED ISSUE #554: planemo uploads to the Tool Shed include unwanted files
It seems planemo is uploading unwanted files to the Tool Shed with individual tool uploads.  In some cases it uploads multiple tools from a demultiplexed git repository into a single tool repository, while in other cases it uploads unwanted files from other tools.

The individual ImageJ2 tools are demultiplexed in here: https://github.com/bgruening/galaxytools/tree/master/tools/image_processing/imagej2
Planemo uploaded each of these tools to the TTS, including every ImageJ2 tool in each individual repo:

```
Repository 'imagej2_adjust_threshold_binary' revision c05d862ac753 (repository tip)
Contents:
imagej2_adjust_threshold_binary.py
imagej2_adjust_threshold_binary.xml
imagej2_adjust_threshold_binary_jython_script.py
imagej2_analyze_particles_binary.py
imagej2_analyze_particles_binary_jython_script.py
imagej2_analyze_skeleton.py
imagej2_analyze_skeleton_jython_script.py
...
```

On the MTS, the meme tool repo contains the fimo tool's Python script:

```
Repository 'meme_meme' revision 487ce3fa1822 (repository tip)
Contents:
all_fasta.loc.sample
fimo_wrapper.py
meme.xml
test-data/
tool_data_table_conf.xml.sample
tool_dependencies.xml
```

There are other similar examples on both the TTS and MTS.

bernt-matthias:
by considering exclude when

- build and downloading tar balls
- diff

shed update should now consider exclude in .shed.yml

I checked this (https://github.com/galaxyproject/planemo/pull/1106/commits/a1e2cae620bf195c70ea97ae09f3ea0db8488861) in the following way on the TTS. 

- added a file (exclude.txt) to a repo (https://testtoolshed.g2.bx.psu.edu/repos/mbernt/fail)
- shed_update
- added the exclude to shed.yml
- shed_update

Afterwards the file was gone again (on the TTS).

Would be great to have for updating the OpenMS tools (where we have quite a few extra files in the TS that shouldn't be there, i.e. we would like to remove them from the TS and not consider them for the question if a repo should be updated). 

TODO(?):

- [ ] I guess we could/should do the same for include, but I'm not sure what the default '**' does
- [ ] not sure if the use of exclude in diff, os.walk, and tar is equivalent (which would be nice) .. also it should be consistent with the docs (https://planemo.readthedocs.io/en/latest/standards/docs/best_practices/shed_yml.html#shed-upload-includes-excludes) and I think it is not yet
- [ ] which other planemo commands no not consider exclude/include and should

may fix: https://github.com/galaxyproject/planemo/issues/554

### DIFF
--- a/planemo/commands/cmd_shed_build.py
+++ b/planemo/commands/cmd_shed_build.py
@@ -20,7 +20,8 @@ def cli(ctx, path, **kwds):
     (which you could upload to the Tool Shed manually).
     """
     def build(realized_repository):
-        tarpath = shed.build_tarball(realized_repository.path)
+        exclude = shed._shed_config_excludes(realized_repository.config)
+        tarpath = shed.build_tarball(realized_repository.path, exclude)
         outpath = realized_repository.real_path + ".tar.gz"
         shutil.move(tarpath, outpath)
         print("Created: %s" % (outpath))
--- a/planemo/shed/__init__.py
+++ b/planemo/shed/__init__.py
@@ -289,7 +289,7 @@ def report_non_existent_repository(realized_repository):
 def upload_repository(ctx, realized_repository, **kwds):
     """Upload a tool directory as a tarball to a tool shed."""
     path = realized_repository.path
-    exclude = realized_repository.config.get("exclude", [])
+    exclude = _shed_config_excludes(realized_repository.config)
     tar_path = kwds.get("tar")
     if not tar_path:
         tar_path = build_tarball(path, exclude, **kwds)
@@ -707,7 +707,7 @@ def create_repository_for(ctx, tsi, name, repo_config):
 
 
 def download_tarball(ctx, shed_context, realized_repository, **kwds):
-    exclude = realized_repository.config.get("exclude", [])
+    exclude = _shed_config_excludes(realized_repository.config)
     repo_id = realized_repository.find_repository_id(ctx, shed_context)
     if repo_id is None:
         message = "Unable to find repository id, cannot download."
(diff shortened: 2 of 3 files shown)


jmchilton:
This looks good to me, is it ready to be merged?

bernt-matthias:
> This looks good to me, is it ready to be merged?

Wondering about the questions in the TODOs. In particular question 2.

Is there anything that deserves a test in this changes? If yes ideas?

nsoranzo:
`build_tarball` is also used in `planemo/commands/cmd_shed_build.py` , so that needs to be also updated or `exclude` made optional.

bernt-matthias:
> `build_tarball` is also used in `planemo/commands/cmd_shed_build.py` , so that needs to be also updated or `exclude` made optional.

No problem, but I'm still quite confused by the code. It seems that `RawRepositoryDirectory` already considers excludes (https://github.com/galaxyproject/planemo/blob/38227c1c17137de3bb085f941f7bfd3e9a08aa77/planemo/shed/__init__.py#L1025) partially (for instance not in _realized_files)  but RealizedRepositry not (btw. do we want to fix the typo).

Both tools essentially start by calling `shed.for_each_repository` -> `_realize_effective_repositories` -> there `_find_raw_repositories` constructs `RawRepositoryDirectory`s which are then turned into `RealizedRepositry`. So I guess the `excludes` could be ignored in the calls to `build_tarball` from the commands.

Is this correct?

jmchilton:
I don't remember all the details - but the idea is that the raw directory structures are materialized into effective repository structures with includes and excludes handled and tools and tool-data demultiplexed. So should any of this be needed? Maybe exclude is only used for automatic demultiplexing though?

jmchilton:
Closing this out because the agent thinks it no longer applies. 

> Issue #554 is already fixed on master — per-repo tarballs from the demultiplexing scenario are identical on both branches, cat2.xml already absent. The PR's stated motivation doesn't survive contact with current master.

If this is wrong - feel free to reopen a deconflicted PR and I will fight my agent over it.

PR #1110, opened 2020-12-14
docs: fix link from a redirect loop to a section

martenson:
if only I knew how to link relatively in rst


nsoranzo:
You can use [:ref:](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-ref), see e.g. how https://github.com/galaxyproject/galaxy/blob/dev/doc/source/admin/reports.rst#L61 refers to the label added in https://github.com/galaxyproject/galaxy/blob/dev/doc/source/admin/nginx.md#L353 .

PR #1111, opened 2020-12-15
raise report level for some shed linting

### LINKED ISSUE #1112: Make CI check for validity of repository name
We just had a case (repo name `apCLMS_to_RamClustR_converter`) where the CI passes [fine](https://github.com/RECETOX/galaxytools/runs/1549734848) but the toolshed upload [fails](https://github.com/RECETOX/galaxytools/runs/1552741248?check_suite_focus=true) (**silently**)

```
Unexpected HTTP status code: 400: {"err_msg": "Repository names must contain only lower-case letters, numbers and underscore.", "err_code": 400008}
```

bernt-matthias:
I think these are all cases which lead to errors in deploying.

fixes #1112

### DIFF
--- a/planemo/shed_lint.py
+++ b/planemo/shed_lint.py
@@ -291,13 +291,13 @@ def lint_shed_yaml(realized_repository, lint_ctx):
     path = realized_repository.real_path
     shed_yaml = os.path.join(path, ".shed.yml")
     if not os.path.exists(shed_yaml):
-        lint_ctx.info("No .shed.yml file found, skipping.")
+        lint_ctx.error("No .shed.yml file found, skipping.")
         return
     try:
         with open(shed_yaml, "r") as fh:
             yaml.safe_load(fh)
     except Exception as e:
-        lint_ctx.warn("Failed to parse .shed.yml file [%s]" % unicodify(e))
+        lint_ctx.error("Failed to parse .shed.yml file [%s]" % unicodify(e))
         return
     lint_ctx.info(".shed.yml found and appears to be valid YAML.")
     _lint_shed_contents(lint_ctx, realized_repository)
@@ -311,7 +311,9 @@ def _lint_shed_contents(lint_ctx, realized_repository):
         if value is not None:
             msg = func(value, *args)
             if msg:
-                lint_ctx.warn(msg)
+                lint_ctx.error(msg)
+        else:
+            lint_ctx.error("Repository does not define: %s" % key)
 
     _lint_if_present("owner", validate_repo_owner)
     _lint_if_present("name", validate_repo_name)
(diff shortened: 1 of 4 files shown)


planemo/shed_lint.py
```diff
@@ -311,7 +311,9 @@ def _lint_if_present(key, func, *args):
         if value is not None:
             msg = func(value, *args)
             if msg:
-                lint_ctx.warn(msg)
+                lint_ctx.error(msg)
+        else:
+            lint_ctx.error("Repository does not define: %s" % key)
```
nsoranzo, line 316: I think this is going against the original intent of the method `_lint_if_present` ? E.g. `type` is `unrestricted` by default, so probably no need to give an error if it's not present.

bernt-matthias: Maybe we can rename the function? My aim it to error in cases where `shed_lint` will error ... and I guess this is the case here.

bernt-matthias: OK. I see .. the type does not need to be defined in `.shed.yml`.

bernt-matthias: Is this also the case for `name`, `owner`, and `categories`?

nsoranzo: `name` and `owner` can be specified on the command-line with e.g. `shed_update`. `name` defaults to the directory name if it's not specified in other ways. Categories I think are optional? I suppose these can all be warnings if not present.

bernt-matthias: I think `name` and `owner` should also be in the realized repo if given on the command line: https://github.com/galaxyproject/planemo/blob/157cace58d4761cb42a8003d47221e55dce95da9/planemo/shed/__init__.py#L1022 (but I do not understand the code for setting `name` ..)

planemo/shed_lint.py
```diff
@@ -290,14 +290,14 @@ def lint_repository_dependencies(realized_repository, lint_ctx):
 def lint_shed_yaml(realized_repository, lint_ctx):
     path = realized_repository.real_path
     shed_yaml = os.path.join(path, ".shed.yml")
-    if not os.path.exists(shed_yaml):
-        lint_ctx.info("No .shed.yml file found, skipping.")
+    if not os.path.exists(shed_yaml) and realized_repository.repository_type != REPO_TYPE_UNRESTRICTED:
```
nsoranzo, line 293: Not sure I understand this change?

bernt-matthias: Was an attempt to fix the testing error for https://github.com/galaxyproject/planemo/tree/master/tests/data/repos/suite_1 which fails because there is no `.shed.yml` .. which is OK in this case .. or isn't it?

jmchilton (approved):

PR #1116, opened 2020-12-16
Fix for workflow_lint to work with params/collections and add it to docs

simonbray:
@bernt-matthias here is the fix for `workflow_lint`.

I think I'll keep this PR open for now and continue adding any other necessary fixes, rather than open lots of small PRs.

### DIFF
--- a/docs/_running_intro.rst
+++ b/docs/_running_intro.rst
@@ -1,3 +1,19 @@
+.. note::
+
+    Some features described in this tutorial are currently available only in
+    the development version of Planemo and may not work correctly when installing
+    the latest version from PyPI or bioconda. To install the development versions
+    of Planemo and galaxy-tool-util required, run the following commands:
+
+    ::
+
+        $ # install dev version of planemo
+        $ pip install https://github.com/galaxyproject/planemo/archive/master.tar.gz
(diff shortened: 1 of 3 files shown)


jmchilton:
Thanks so much. We should just merge this - it looks solid! Lots of little PRs are good IMO.

docs/_running_intro.rst
```diff
@@ -1,3 +1,19 @@
...
+    the latest version from PyPI or bioconda. To install the development versions
+    of Planemo and galaxy-tool-util required, run the following commands:
+
+    ::
+
+        $ # install dev version of planemo
+        $ pip install https://github.com/galaxyproject/planemo/archive/master.tar.gz
+        $ # install dev version of galaxy-tool-util
+        $ git clone https://github.com/galaxyproject/galaxy.git
+        $ pip install -e galaxy/packages/tool_util
```
jmchilton, line 14: I published the latest release to PyPI if that helps at all. https://pypi.org/project/galaxy-tool-util/21.1.0.dev4/

simonbray: Yes it does, thanks a lot! You can go ahead and merge this if you like.

jmchilton:
Thanks so much - keep up the awesomeness!

PR #1123, opened 2020-12-29
More cohesive abstraction for building runnables from CLI args.

jmchilton:
I want to setup this abstraction so I can use to build URIs for testing remote resources, etc.. so more coming in subsequent PRs.


bgruening:
Thanks John.

PR #1128, opened 2021-01-04
Bump galaxy-tool-util version in requirements

simonbray:
Specifically, so that that these two little PRs are available: https://github.com/galaxyproject/galaxy/pull/10334, https://github.com/galaxyproject/galaxy/pull/10886

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ cwltool>=1.0.20191225192155
 docutils
 ephemeris>=0.10.3
 galaxy-containers
-galaxy-tool-util>=20.9.0.dev2
+galaxy-tool-util>=21.1.0.dev4
 galaxy-util>=20.5.0
 glob2
 gxformat2>=0.12.0


bgruening (approved):

requirements.txt
```diff
@@ -7,7 +7,7 @@ cwltool>=1.0.20191225192155
 docutils
 ephemeris>=0.10.3
 galaxy-containers
-galaxy-tool-util>=20.9.0.dev2
+galaxy-tool-util>=21.1.0.dev4
```
nsoranzo, line 10: This could be an issue when updating the conda package for planemo, since this is a pre-release and there is no 21.1.x proper release.

bgruening: We should be able to get a conda package for this as well - if needed.

simonbray: True, but I didn't have the impression that having a up-to-date conda package was a high priority (the currently specified version is also a pre-release).

simonbray: We can also wait for the official 21.01, it doesn't really bother me.

nsoranzo: > the currently specified version is also a pre-release

Yes, but there is 20.9.x non pre-release package to satisfy the `>=` requirement. Any way it's not an issue until this ends in a Planemo release and it's good to have this merged at some point.

bgruening: if you have concerns, we could also create a dev branch and merge those PRs into the dev branch. I do think its worth to start working on new features during Galaxy release phase.

PR #1130, opened 2021-01-17
ci_find*: include sub and parent dirs

### LINKED ISSUE #1129: ci_find_tools misses changed tools if tool XML is unchanged
`ci_find_tools --changed_in_commit_range RANGE` seems not find changed tools if the main xml file is not changed (e.g. if only scripts or test-data is changed). 

Not sure if this is the desired behavior, but the finding such tools would be needed in the IUC PR workflow in the linting step: https://github.com/galaxyproject/tools-iuc/blob/4daa375d022673d2437d609b1865b78c64b04415/.github/workflows/pr.yaml#L156

I guess for each non-xml file one would need to traverse the tree up to the repo root and take the first tool xml that is found in one of these paths.

bernt-matthias:
fixes https://github.com/galaxyproject/planemo/issues/1129  and more: 

include also tools/repos with:

- change of test data
- change of supplementary scripts (this already was the case for ci_find_repos)
- change of macros where the tools are in sub-directories (this failed for ci_find_repos if each tool has an individual shed file)
- change of a single tool in a collection where the shed file is in the parent

currently this will "fail" if the working dir of the planemo run is not the root of the repository: then changing for instance the README of the repository would lead to the inclusion of the all tools/repos in the repository. We could check if its the root by checking if there is a `.git` directory?

### DIFF
--- a/planemo/ci.py
+++ b/planemo/ci.py
@@ -3,6 +3,7 @@
 from __future__ import print_function
 
 import copy
+import glob
 import math
 import os
 
@@ -11,6 +12,10 @@ import yaml
 from planemo import git
 from planemo import io
 from planemo.shed import REPO_METADATA_FILES
+from planemo.tools import (
+    is_tool_load_error,
+    yield_tool_sources_on_paths
+)
 
 
 def filter_paths(ctx, raw_paths, path_type="repo", **kwds):
@@ -26,14 +31,9 @@ def filter_paths(ctx, raw_paths, path_type="repo", **kwds):
     if changed_in_commit_range is not None:
         diff_files = git.diff(ctx, cwd, changed_in_commit_range)
         if path_type == "repo":
-            diff_dirs = set(os.path.dirname(p) for p in diff_files)
-            diff_paths = set()
-            for diff_dir in diff_dirs:
-                diff_path = metadata_file_in_path(diff_dir)
-                if diff_path:
-                    diff_paths.add(diff_path)
+            diff_paths = changed_repos(diff_files)
         else:
-            diff_paths = diff_files
+            diff_paths = changed_tools(diff_files, ctx, cwd)
 
     unique_paths = set(os.path.relpath(p, cwd) for p in raw_paths)
     if diff_paths is not None:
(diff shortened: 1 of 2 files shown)


mvdbeek:
> currently this will "fail" if the working dir of the planemo run is not the root of the repository: then changing for instance the README of the repository would lead to the inclusion of the all tools/repos in the repository. We could check if its the root by checking if there is a `.git` directory?


To clarify, the problem here is if planemo's working dir is e.g `tools-iuc/tools/minimap2`, and not `tools-iuc/` ?
I think that's fine, planemo ci_find* is really meant for running within a CI system, and I think it's reasonable to expect that the working directory is the repository root.

On the other hand I don't think the presence of `.git` is a good proxy to check for this.

bernt-matthias:
> the problem here is if planemo's working dir is e.g `tools-iuc/tools/minimap2`, and not `tools-iuc/` ?

This might also be a problem. But I thought more of the case when the working dir is the parent of `tools-iuc/`.

> I think that's fine, planemo ci_find* is really meant for running within a CI system, and I think it's reasonable to expect that the working directory is the repository root.

Then everything should be fine.

planemo/ci.py
```diff
@@ -55,6 +55,38 @@ def filter_paths(ctx, raw_paths, path_type="repo", **kwds):
     return chunked_paths
 
 
+def changed_repos(diff_files):
+    diff_dirs = set(os.path.dirname(p) for p in diff_files)
+    diff_paths = set()
+    for diff_dir in diff_dirs:
+        new_diff_paths = set()
+        while diff_dir != "" and len(new_diff_paths) == 0:
```
bernt-matthias, line 63: Maybe a bit more explanation on the potential problem: 

With `diff_dir != ""` I ensure that  `glob.glob(diff_dir + "/**/"` is not called for the working dir. This is needed because otherwise a change of a file in the repository root would lead to the inclusion of all tools / repos.

Hence calling from outside of the repo this would not work .. BUT this is irrelevant, because `git diff` will fail before that.

Calling from subdirs would lead to ignoring changes in this dir.


Maybe we can assert the assumption by comparing the output of `git rev-parse --show-toplevel` with the working dir.

mvdbeek: That said if this lets you make progress on the planemo action I'm happy to merge and iterate on it if there are issues.

bernt-matthias: That's fine. I can easily use test cases that include changes in the tool xml. 

Just need to think of something different than `--recursive`, because this does not capture the essence. Maybe `--extended_git_diff`.

jmchilton:
From Claude:

> PR 1130 now does what the 2021 review actually converged on: new behavior opt-in behind --extended_git_diff, default path provably identical to master, three defects fixed with tests, and a tests/test_ci.py that didn't exist before.

jmchilton (approved):

mvdbeek:
Nice, i think that'll catch a few drift issues!

PR #1133, opened 2021-01-22
Push to origin as gh fork renames original origin to upstream

mvdbeek:
See https://github.com/cli/cli/issues/582


nsoranzo:
Ops, just noticed the wip, sorry! Do you want this to be reverted in the mean time?

PR #1145, opened 2021-02-24
Fix virtualenv mess maybe

mvdbeek:


jmchilton:
I've spent so many hours banging my head futilely at this issue - I will be so grateful if this works!

mvdbeek:
I did see some really odd behavior locally, and that fixes it, but I admit I'm flying blind here :/. Ideally we'd add some testing with conda, brew and pyvenv to be sure

planemo/virtualenv.py
```diff
@@ -3,47 +3,28 @@
...
-        ]
-    else:
-        base_command = [
-            planemo_path, "virtualenv",
-        ]
-
-    command = base_command
-
     # Create a virtualenv with the selected python version.
     # default to 2.7
```
nsoranzo, line 17:
```suggestion
```

planemo/virtualenv.py
```diff
@@ -3,47 +3,28 @@
...
     if galaxy_python_version is None:
         galaxy_python_version = DEFAULT_PYTHON_VERSION
+    pyvenv = which('pyvenv')
     python = which("python%s" % galaxy_python_version)
     if python:
         python = os.path.abspath(python)
-        command.extend(["-p", python])
-    command.append(virtualenv_path)
+    else:
+        python = 'python'
```
nsoranzo, line 25:
```suggestion
        python = 'python3'
```

planemo/virtualenv.py
```diff
@@ -3,47 +3,28 @@
...
-import sys
 
-import virtualenv
 from galaxy.util.commands import which
 
-
 DEFAULT_PYTHON_VERSION = os.environ.get("PLANEMO_DEFAULT_PYTHON_VERSION", "3")
 
 
-def create_and_exit(virtualenv_path, **kwds):
```
nsoranzo, line 15: This function is still used in `planemo/commands/cmd_virtualenv.py` , did you intend to remove that too?

planemo/virtualenv.py
```diff
@@ -3,47 +3,28 @@
...
-            planemo_path, "virtualenv",
-        ]
-
-    command = base_command
-
     # Create a virtualenv with the selected python version.
     # default to 2.7
     if galaxy_python_version is None:
         galaxy_python_version = DEFAULT_PYTHON_VERSION
+    pyvenv = which('pyvenv')
```
nsoranzo, line 20: pyvenv is deprecated since Python 3.6, which is the earliest version supported by Planemo.

planemo/virtualenv.py
mvdbeek: I think there a some distribution that don't create a python3 binary/symlink.

mvdbeek: But maybe we can use https://docs.python.org/3/library/sys.html#sys.executable here, that is probably the least surprising default and would give us a python version that is compatible with current planemo, which should be on par with what Galaxy supports.

nsoranzo (approved):

PR #1151, opened 2021-03-19
Extend autoupdate subcommand to workflows

simonbray:


jmchilton:
Is this ready for review? Can you rebase it with the latest head - I think other PRs are green right now.

simonbray:
I'll have another look at it, I think it should be almost ready though.

The reason why I gave up working on it was because I wasn't sure how to deal with subworkflows - I don't know how updating a subworkflow to a newer version should work in the context of a local, throwaway Galaxy instance. Eventually hopefully it will be possible to publish a workflow with the IWC and then point subworkflows to it using a TRS id or similar.

But I suppose that in the meantime this is still useful, even if it is only updating tools and not subworkflows.

simonbray:
Cool, thank you for merging @mvdbeek

mvdbeek:
This is very cool!

Issue #1149, opened 2021-03-12
Planemo run should return a short parsable report


simonbray:
At least for external galaxies, containing history_id, invocation_id, etc. If the user wants to run multiple workflows after one another, then this information can be parsed and used in further analysis steps.

PR #1153, opened 2021-03-22
Add JSON report for planemo run invocations

simonbray:
Fixes https://github.com/galaxyproject/planemo/issues/1149.

We could make this way more extensive, e.g. include all job ids, all invocation step ids, etc... For @wm75 and me this would be sufficient at the moment:

```
{
    "error_message": "",
    "history_id": "e96c6ce43a3d239a",
    "history_state": "ok",
    "invocation_id": "b735ed9e5e005602",
    "invocation_state": "scheduled",
    "workflow_id": "8a72a9a90df70b20"
}
```

Maybe we can extend it later if there's interest.


mvdbeek (approved):
I think this is a cool start. Maybe you could model the run report after the test report functionality that generates the report from a template (https://github.com/galaxyproject/planemo/blob/1b61aa1f75b6fad5ba14efc42e6746d1ade0f09a/planemo/commands/cmd_test_reports.py#L1) ?

simonbray:
> I think this is a cool start. Maybe you could model the run report after the test report functionality that generates the report from a template (
> 
> https://github.com/galaxyproject/planemo/blob/1b61aa1f75b6fad5ba14efc42e6746d1ade0f09a/planemo/commands/cmd_test_reports.py#L1
> ) ?

I don't think we need something human-readable actually - we just want to be able to get the IDs. Wolfgang is parsing `planemo -v` atm, which is not really ideal.

(We don't have to use YAML btw, JSON is also fine if you prefer.)

mvdbeek:
I don't think the report itself needs to be human readable (as opposed to what the docs say), I think what's cool is that we can start from the same JSON and produce different kind of reports (markdown, HTML, xunit, plain text, yaml).

bgruening (approved):
Cool!

jmchilton:
I was just hacking on that invocation test report for https://github.com/galaxyproject/planemo/pull/1152/files#diff-27ed15e8e508ef68adea5c98c2025fc3e70937c5d14affdaeaa59b56ba25a6dcR386. It would be nice if this exact JSON is something we could put in the test result JSON somewhere - maybe as a subset of the "data" key or a an "invocation_details" key below "data".

I guess that is just an implementation detail though.

bgruening:
@jmchilton do you mean the json structure should look different so it can be reused?

jmchilton:
Yeah - let me be more concrete. I think it should be called ``execution_report`` instead of ``invocation_report`` since ``run`` could be done on tools also and in non-Galaxy contexts.

Move:

```python
        invocation_details = {
            'invocation_id': run_result._invocation_id,
            'history_id': run_result._history_id,
            'workflow_id': run_result._workflow_id,
            'invocation_state': run_result.invocation_state,
            'history_state': run_result.history_state,
            'error_message': run_result.error_message,
        }
```

Into an ``execution_details()`` method on the ``GalaxyWorkflowRunResponse`` object and with an empty abstract implementation placed in ``RunResponse``.

Then in ``def structured_test_data(self, run_response):`` of ``planemo/runnable.py`` also place the details in the ``data_dict``.

``data_dict['execution_details'] = run_response.execution_details()``

That way tooling that is developed around ``planemo run`` for workflows can also be used from the JSON for test cases.

simonbray:
Okay, I'll have a look.

> an execution_details() method on the GalaxyWorkflowRunResponse object and with an empty abstract implementation placed in RunResponse.

Aren't we duplicating a lot of stuff here? All this data comes from the GalaxyWorkflowRunResponse in the first place.

simonbray:
There seems to be a step missing, structured_test_data is a method of the TestCase class which is not used when running runnables without testing them.

I can probably work get it to work (maybe add a structured_test_data method to the RunResponse class as well?) but otherwise maybe you can clarify what you meant @jmchilton?

jmchilton:
I think what I was thinking was this - https://github.com/galaxyproject/planemo/commit/b8cebeb6704463448eec995e6397db46556406f2. But now I see there is a thing called invocation_details that is a lot like what I was proposing with execution_details 😆 - of course @mvdbeek did it first and better.

His general idea of making this into a report option would be doable if we ensure this is all in there in this way (which it may already be though - I'm unsure).

simonbray:
I would be happy to make the simplest change possible:
```
    invocation_report = kwds.get("invocation_report", None)
    if invocation_report:
        with open(invocation_report, "w") as f:
            json.dump(run_result.invocation_details, f, indent=4, sort_keys=True)
```
and then modify the `collect_invocation_details()` method a little bit. But I thought your and @mvdbeek's point was more about making the output compatible with the various report generation options which planemo provides.

But maybe this can be done in a later PR.

simonbray:
I think this might be ready for a review. It produces an output like this:

image

mvdbeek (approved):
This looks fantastic, thanks a lot @simonbray!

PR #1156, opened 2021-03-27
Make simultaneous file upload configurable for planemo run / test

simonbray:
So that if we are running a tool or workflow with multiple input datasets that need to be uploaded, we can choose to have all uploads started at once, rather than waiting for each to complete before the next one starts.

In my opinion this simultaneous uploading could even be the default (it makes more sense for running against an external galaxy at least) but it seems it was considered a bad idea, as planemo explicitly overrides the default behaviour of the Galaxy StagingInterface here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/client/staging.py#L57. Do you remember why you thought this was necessary @jmchilton?

### DIFF
--- a/planemo/galaxy/activity.py
+++ b/planemo/galaxy/activity.py
@@ -81,11 +81,12 @@ def log_contents_str(config):
 
 class PlanemoStagingInterface(StagingInterace):
 
-    def __init__(self, ctx, runnable, user_gi, version_major):
+    def __init__(self, ctx, runnable, user_gi, version_major, simultaneous_uploads):
         self._ctx = ctx
         self._user_gi = user_gi
         self._runnable = runnable
         self._version_major = version_major
+        self._simultaneous_uploads = simultaneous_uploads
 
     def _post(self, api_path, payload, files_attached=False):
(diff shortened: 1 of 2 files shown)


jmchilton:
It was off by default because Planemo runs against an sqlite database by default that used to fail frequently. I think things have gotten more stable and this definitely doesn't need to be one at a time against a production, external Galaxy. Seems like a good enhancement.

bernt-matthias:
Hey @simonbray .. I just discovered this option and wonder how to make use of it. It appears to me that it does not make a difference in a standard `galaxy test tool.xml` call?

simonbray:
> Hey @simonbray .. I just discovered this option and wonder how to make use of it. It appears to me that it does not make a difference in a standard `galaxy test tool.xml` call?

`planemo test` runs tool tests by default through Galaxy's `./run_tests.sh` script afaik, so it would not touch this code at all. If you have a local Galaxy server running separately and run `planemo test` against it, hopefully you should see the difference.

I'm not sure if this option is useful for standard tool testing though? It was definitely meant more for external galaxies, see the discussion above.

PR #1157, opened 2021-03-29
Create --download_outputs flag for planemo run

simonbray:
When running using `--external_galaxy`, downloading the outputs is probably superfluous, so this PR adds a `--download_outputs` flag which needs to be specified to perform downloads when using `planemo run`.

For `planemo test` the current behaviour is unchanged (files need to be downloaded to be compared against test data.)

ping @wm75


wm75:
@simonbray seems to work as intended!

bgruening (approved):

PR #1165, opened 2021-04-29
print debug messages for galaxy.tools.actions

bernt-matthias:
in order to see output filter errors

also set level to WARN for urllib3 to ignore messages like

```
2021-04-29 12:00:19,558 DEBUG [urllib3.connectionpool] http://127.0.0.1:9925 "GET /api/jobs/54f2a3a23292eb07 HTTP/1.1" 200 None
2021-04-29 12:00:19,821 DEBUG [urllib3.connectionpool] Starting new HTTP connection (1): 127.0.0.1:9925
```

which are flooding the logs

I still see 

```
INFO:     127.0.0.1:35964 - "POST /api/tools HTTP/1.1" 200 OK
```

but have no idea where this is coming from. Is there a way to address logging for all modules but a certain set (i.e. set warn for all modules but galaxy and planemo)?

### DIFF
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -513,7 +513,7 @@ def local_galaxy_config(ctx, runnables, for_tests=False, **kwds):
         env["GALAXY_TEST_LOGGING_CONFIG"] = config_join("logging.ini")
         env["GALAXY_DEVELOPMENT_ENVIRONMENT"] = "1"
         # disable all access log messages from uvicorn
-        env["GALAXY_UVICORN_ACCESS_LOG"] = "False"
+        env["GALAXY_TEST_DISABLE_ACCESS_LOG"] = "False"
         # Following are needed in 18.01 to prevent Galaxy from changing log and pid.
         # https://github.com/galaxyproject/planemo/issues/788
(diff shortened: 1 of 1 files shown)


mvdbeek:
> but have no idea where this is coming from.

This is uvicorn's logging ... I looked into how we can turn this off, but I didn't see something that would work at first glance.

bernt-matthias:
Wondering why the urllib messages were still shown (https://github.com/galaxyproject/galaxy/pull/11266)?

bernt-matthias:
> This is uvicorn's access logging ... I looked into how we can turn this off, but I didn't see something that would work at first glance.

I think I found a way: https://github.com/galaxyproject/galaxy/pull/11908/commits/e9b08f300a8ee1db2eb881593de8dcdb587a554c

If we like we can also have more control over the uvicorns logging config https://github.com/encode/uvicorn/blob/bf1c64e2c141971c546671c7dc91b8ccf0afeb7d/uvicorn/config.py#L131 ... e.g. the formater.

mvdbeek:
Nice! Thanks a lot @bernt-matthias!

PR #1174, opened 2021-06-07
Create more useful output for failed invocations

mvdbeek:
This should make debugging invocations with failed steps much easier. The core of the change is the try/except AssertionError, which is raised when the step that produces a dataset collection is in error (https://github.com/galaxyproject/galaxy/blob/release_21.01//lib/galaxy/tool_util/cwl/util.py#L447). I'll also remove that on the Galaxy side. Without this change no useful output is produced except that the test failed. With the change we get the regular test structure written out.

### DIFF
--- a/tests/test_test_engines.py
+++ b/tests/test_test_engines.py
@@ -1,3 +1,4 @@
+import json
 import os
 from tempfile import NamedTemporaryFile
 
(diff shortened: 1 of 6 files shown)


tests/test_test_engines.py
```diff
@@ -169,3 +170,39 @@ def test_galaxy_workflow_non_data_inputs():
...
+def test_galaxy_workflow_step_failed():
+    ctx = t_context()
+    test_artifact = os.path.join(TEST_DATA_DIR, "wf_failed_step.ga")
+    runnables = for_paths([test_artifact])
+    with NamedTemporaryFile(prefix="result_json") as json_out:
+        kwds = {
+            "engine": "galaxy",
+            "no_dependency_resolution": True,
+            "paste_test_data_paths": False,
+            "extra_tools": ['$GALAXY_FUNCTIONAL_TEST_TOOLS'],
```
jmchilton, line 186: Love this, thanks so much!

jmchilton (approved):

PR #1179, opened 2021-07-12
Ignore failure to download output datasets

mvdbeek:
If we raise an exception here we won't get a full report with what
really is the issue.
Also uses output_to_cwl_json for hdas, since it produces the same
output.

Should produce a more meaningful test report for https://github.com/galaxyproject/iwc/pull/47

### DIFF
--- a/planemo/galaxy/activity.py
+++ b/planemo/galaxy/activity.py
@@ -1,6 +1,5 @@
 """Module provides generic interface to running Galaxy tools and workflows."""
 
-import json
 import os
 import sys
 import tempfile
@@ -15,7 +14,6 @@ from galaxy.tool_util.client.staging import (
 )
 from galaxy.tool_util.cwl.util import (
     invocation_to_output,
-    output_properties,
     output_to_cwl_json,
     tool_response_to_output,
 )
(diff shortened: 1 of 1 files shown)


planemo/galaxy/activity.py
```diff
@@ -456,7 +446,15 @@ def _history_content_download(self, history_id, dataset_id, to_path, filename=No
...
-        r.raise_for_status()
+        try:
+            r.raise_for_status()
+        except HTTPError as e:
+            # When a job fails abruptly the object store may not contain a dataset,
+            # and that results in an internal server error on the Galaxy side.
+            # We don't want this to break the rest of the test report.
+            # Should probably find a way to propagate this back into the report.
+            ctx.vlog(f"Failed to download history content at URL {url}, exception: {e}")
+            return
```
jmchilton, line 457: The job needs to fail right? You can run test assertions on this.

mvdbeek: If we're reaching this point the job has already failed (most likely), and the test assertions would run on an empty file, so they should still produce the failure. Checkout line 380 and 363, where we're also skipping outputs.

PR #1182, opened 2021-07-21
Fix --shed_install for gxformat2 workflows

mvdbeek:


nsoranzo:
Do we need to add a test for this? I couldn't find a Format 2 workflow with `tool_shed_repository` in `tests/` .

planemo/galaxy/workflows.py
```diff
@@ -32,9 +32,15 @@ def load_shed_repos(runnable):
         # require explicit annotation like this... I think?
         with open(path, "r") as f:
             workflow = yaml.safe_load(f)
-
-        tools = workflow.get("tools", [])
```
jmchilton, line 36: I liked the idea of the workflows also serving as ephemeris installable entities - so the workflow could just have a list of repos to install. I guess that never made it into the schema though 🤔.

mvdbeek: We should make the tool ids resolvable, then we don't need this duplication (or the tool_shed_repositories entry in general) and we could finally get away from the repository concept 🙏.

PR #1187, opened 2021-09-21
Select refgenie config based on Galaxy version

mvdbeek:
if available. Else default to current version (0.4).

Fixes https://github.com/galaxyproject/tools-iuc/runs/3654913135?check_suite_focus=true


bernt-matthias (approved):

jmchilton (approved):

jmchilton:
Thanks for working through all this complexity. I wish we were deriving more benefit from refgenie.

PR #1190, opened 2021-09-27
test: check if extra files were created in Galaxy's file_path

### LINKED ISSUE #1189: Fail test if tool writes to input dir
Would be nice if `planemo test` would fail for such cases automatically. Like so https://github.com/galaxyproject/tools-iuc/pull/3986#issuecomment-926729276

bernt-matthias:
This is an attempt to automatize https://github.com/galaxyproject/tools-iuc/pull/3986#issuecomment-926729276

fixes https://github.com/galaxyproject/planemo/issues/1189

Seems difficult / impossible to find out which tool / test generates the extra files, so I just set all tests for failure and add some note to the stderr .. (but IUC style tests run 1 `planemo test` per repo)

I guess this will create quite a bit of noise in the IUC style repos weekly CI tests .. but maybe its a good thing.

### DIFF
--- a/planemo/galaxy/test/actions.py
+++ b/planemo/galaxy/test/actions.py
@@ -128,17 +128,18 @@ def run_in_config(ctx, config, run=run_galaxy_command, test_data_target_dir=None
     # check if tools created extra files in the file_path all files in the files path
     # - end with '.dat' or
     # - are contained in a directory `dataset_.*_files`
-    for (path, dirs, files) in os.walk(config.env["GALAXY_CONFIG_OVERRIDE_FILE_PATH"]):
-        for name in files:
-            if not (name.endswith(".dat") or re.match('.*/dataset_.*_files$', path)):
-                extra_files.append(os.path.join(path, name))
-    if len(extra_files) > 0:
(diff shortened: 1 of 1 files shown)


jmchilton:
The issue has been closed be cause a work around was found. I'm going to close this out but I think the idea is still good - I would just rather it be solved in some way at the Galaxy level instead of having Planemo break into Galaxy's abstractions. The agent prefers just dropping this outright because Docker is the right approach and pushed back on this - I'm a little stubborn and don't agree with their assessment exactly but admit it makes a good point why not.

> Would a Galaxy config option be the right home for detecting that? The layering argument favors it — Galaxy owns the object store and knows exactly which files it created, whereas planemo has to guess with .dat/dataset_*_files heuristics that are already wrong-ish (planemo sets object_store_store_by="uuid", and Galaxy's own comment two lines above the RO mount calls default_file_path "an estimate given object stores anyway"). But the cost is a production config flag, a job-lifecycle hook, and docs, for a check no production admin wants and that exists solely to serve tool-developer CI. I don't think Galaxy maintainers should carry that.

Still like an API endpoint to ask if the object store is dirty or something like that would be a good compromise to keep the abstractions in Galaxy but not pollute the job running path.

PR #1196, opened 2021-10-10
Exclude click version 8.0.2

mvdbeek:
As an addition/alternative to https://github.com/galaxyproject/galaxy/pull/12682. I'm gonna assume it'll be fixed upstream (https://github.com/pallets/click/issues/2088 is the issue, I think).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 allure-python-commons
 BeautifulSoup4
 bioblend>=0.14.0
-Click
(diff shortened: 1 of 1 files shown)


bgruening:
I like that more. But we need a new release isn't it to make Galaxy tests pass?

mvdbeek:
New release is on https://pypi.org/project/planemo/ and the tests are passing again.

PR #1198, opened 2021-10-20
Allways use random id_secret for testing

bernt-matthias:
Fix https://github.com/galaxyproject/planemo-ci-action/issues/18 

As far as I can see this is then set as `'GALAXY_CONFIG_OVERRIDE_ID_SECRET'` in the env over here: 
https://github.com/galaxyproject/planemo/blob/2c3c59402506fef7232e8fe8e2259533b3ff3318/planemo/galaxy/config.py#L514

- The first two commits implement setting `--id_secret` on the CLI, ie. the user needs to generate the secret on its own. 
- Then I thought it might be more user friendly to generate the random hash automatically. 
- Finally as suggested by @nsoranzo changed to always setting a random secret

I'm not 100% sure if Galaxy actually uses the env variable .. any idea how to test this?

### DIFF
--- a/planemo/commands/cmd_test.py
+++ b/planemo/commands/cmd_test.py
@@ -1,7 +1,5 @@
 """Module describing the planemo ``test`` command."""
 import click
-import hashlib
-import time
 
 from planemo import options
 from planemo.cli import command_function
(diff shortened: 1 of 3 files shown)


nsoranzo:
I don't think we need an option for this, we should just always generate a random `id_secret`.

bernt-matthias:
> I don't think we need an option for this, we should just always generate a random `id_secret`.

Good idea. Then we also need no change to the planemo-ci-action.

jmchilton (approved):

bernt-matthias:
Thanks for your approval @jmchilton . After rebasing also tests seem to pass now.

bgruening:
Thanks @bernt-matthias!

PR #1199, opened 2021-10-22
Enable Galaxy ``outputs_to_working_directory`` by default

bernt-matthias:
most importantly this will set the galaxy dir and its files dir to read only (for dockerized testing). which allows planemo test to find cases of tools that write into galaxy's files dir.

see also here https://github.com/galaxyproject/galaxy/blob/5dc9111f45b4ed5012575b31e199a591cfad6d7a/lib/galaxy/config/sample/job_conf.xml.sample_advanced#L483

related: https://github.com/galaxyproject/planemo/pull/1190


nsoranzo:
@bernt-matthias Can you rebase this?

PR #1201, opened 2021-11-02
Replace Galaxy interactor galaxy_requests_post with make_post_request from BioBlend

simonbray:
Planemo tests are currently failing here (https://github.com/bgruening/galaxytools/pull/1173) with

```
planemo test --galaxy_python_version 3.7 --no_conda_auto_init --galaxy_source https://github.com/galaxyproject/galaxy --galaxy_branch release_21.05 /tmp/tmp.Uk4Nu5LXyy
Problem loading command test, exception cannot import name 'galaxy_requests_post' from 'galaxy.tool_util.verify.interactor' (/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/galaxy/tool_util/verify/interactor.py)
```
It seems to be because the `galaxy_requests_post` method was removed in Galaxy. Hopefully replacing it with BioBlend's `make_post_request` method will fix it.

### DIFF
--- a/planemo/galaxy/activity.py
+++ b/planemo/galaxy/activity.py
@@ -18,7 +18,6 @@ from galaxy.tool_util.cwl.util import (
     tool_response_to_output,
 )
 from galaxy.tool_util.parser import get_tool_source
-from galaxy.tool_util.verify.interactor import galaxy_requests_post
 from galaxy.util import (
     safe_makedirs,
     unicodify,
@@ -91,7 +90,7 @@ class PlanemoStagingInterface(StagingInterace):
     def _post(self, api_path, payload, files_attached=False):
         params = dict(key=self._user_gi.key)
         url = urljoin(self._user_gi.url, "api/" + api_path)
-        return galaxy_requests_post(url, data=payload, params=params, as_json=True).json()
+        return self._user_gi.make_post_request(url, payload=payload, params=params).json()
 
     def _attach_file(self, path):
         return attach_file(path)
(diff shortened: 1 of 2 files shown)


mvdbeek (approved):
That seems like a good idea. https://github.com/mvdbeek/galaxy/commit/b7ab45dd388913888c5e9cd3b14aaefbb2daf45c#diff-06835be1d1943ebce8373e2cd1f8e86c7b7813b720398850c978b734dd29e2da is where galaxy_requests_post was dropped.

davelopez:
Wow, sorry I didn't find any other reference in the codebase and didn't consider this method may be used in external projects.

Thanks for fixing it @simonbray!

simonbray:
@davelopez no worries, I think it isn't realistic to check all dependent projects every time you refactor some code.

mvdbeek:
We should IMO pin requirements for releases of planemo. That is a common technique when you release applications (as opposed to libraries, where you should not pin). I think planemo's use as a library is not that common, so I don't think there are many downsides.

martenson:
thanks for fixing this @simonbray !
@davelopez when dropping an exposed method you can do sanity check by entering something like this in a github search: `org:galaxyproject galaxy_requests_post`

martenson (approved):

mvdbeek:
Thanks everyone, I'll cut a new release in the 0.74 version range. If/when we merge https://github.com/galaxyproject/planemo/pull/1203 we should probably jump to 0.75

mvdbeek:
New release is on PyPI: https://pypi.org/project/planemo/0.74.9/

PR #1140, opened 2021-02-17
Add rerun subcommand for rerunning jobs

simonbray:
Implements https://github.com/galaxyproject/planemo/issues/1138.

PR #1203, opened 2021-11-03
Add poetry setup to produce locked releases

mvdbeek:
That's the first time I set up poetry for a project, I'd appreciate your insights @nsoranzo @jmchilton


poetry.lock
```diff
@@ -0,0 +1,2615 @@
+[[package]]
```
nsoranzo, line 1: Not sure we want to track this in the repository?

Makefile
```diff
@@ -56,6 +56,9 @@ submodule:
 install: submodule ## install into Python envirnoment
 	python setup.py install && cd cwl-runner && python setup.py install
 
+dependencies-update:
+	poetry update && poetry export --without-hashes  -f requirements.txt  --output requirements.txt && poetry export --dev --without-hashes  -f requirements.txt  --output dev-requirements.txt
```
nsoranzo, line 60: Do these commands need to be chained with `&&` or can they be on separate lines?

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+readme = "*"
+pyflakes = "*"
+flake8 = "*"
+flake8-import-order = "*"
+wheel = "*"
+twine = "*"
+
+[build-system]
+requires = ["poetry-core>=1.0.0", "setuptools"]
+build-backend = "poetry.core.masonry.api"
```
nsoranzo, line 51: I don't think we want to force the use of `poetry` to install planemo.

```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+name = "planemo"
+version = "0.75.0"
+description = "Command-line utilities to assist in building tools for the Galaxy project (http://galaxyproject.org/)."
+authors = ["Galaxy Project and Community"]
+license = "MIT"
+
+[tool.poetry.dependencies]
+python = "^3.6"
+allure-python-commons = "*"
+BeautifulSoup4 = "*"
```
nsoranzo, line 11: That's not needed any more.

```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+bioblend = "*"
+click = "*"
+configparser = "*"
```
nsoranzo, line 14:
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+cwltool = "*"
+docutils = "*"
```
nsoranzo, line 16:
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+ephemeris = "*"
+galaxy-containers = "*"
+galaxy-tool-util = "*"
+galaxy-util = "*"
+glob2 = "*"
+gxformat2 = "*"
+jinja2 = "*"
+lxml = "*"
+oyaml = "*"
+pyaml = "*"
```
nsoranzo, line 26:
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+toil = "*"
+flask = "*"
+Werkzeug = "*"
+tox = "*"
+pytest = "*"
+coverage = "*"
+sphinx = "*"
+sphinx_rtd_theme = "*"
+recommonmark = "*"
+readme = "*"
```
nsoranzo, line 42: Obsolete and not used directly in planemo, its successor is installed by twine.
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+pyflakes = "*"
```
nsoranzo, line 43: Dep of flake8
```suggestion
```

pyproject.toml
mvdbeek: I do think we need `requires = ["setuptools"]`, otherwise it seems `pip install -e` fails

mvdbeek: I see, that's only needed if we have the `[build-system]` section

mvdbeek:
Yay, all green. What do you think about this @jmchilton ?

dev-requirements.txt
```diff
@@ -1,28 +1,130 @@
...
+decorator==4.4.2; python_version >= "3.6" and python_full_version < "3.0.0" and python_version < "4" or python_version >= "3.6" and python_version < "4" and python_full_version >= "3.2.0"
+dill==0.3.4; python_version >= "3.6" and python_full_version < "3.0.0" or python_full_version >= "3.1.0" and python_version >= "3.6"
+distlib==0.3.3; python_version >= "2.7" and python_full_version < "3.0.0" or python_full_version >= "3.5.0"
+docker==4.3.1; python_version >= "3.6" and python_full_version < "3.0.0" or python_full_version >= "3.5.0" and python_version >= "3.6"
+docutils==0.17.1; python_version >= "3.6" and python_full_version < "3.0.0" or python_full_version >= "3.5.0" and python_version >= "3.6"
+enlighten==1.10.1; python_version >= "3.6"
+ephemeris==0.10.7
+filelock==3.3.2; python_version >= "3.6" and python_full_version < "3.0.0" or python_full_version >= "3.5.0" and python_version >= "3.6"
+flake8-import-order==0.18.1
+flake8==2.5.5
```
nsoranzo, line 37: This seems quite an old version. When running poetry locally, it got resolved to the latest (4.0.1). How did you install poetry? It needs to be in its own virtualenv to prevent mixing its own dependencies with the one of the project.

I'll update this with my resolved version of (*)requirements.txt, if you don't mind.

mvdbeek: I just followed the instructions in https://python-poetry.org/docs/, but of course that'd be great.

pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+name = "planemo"
+version = "0.75.0"
+description = "Command-line utilities to assist in building tools for the Galaxy project (http://galaxyproject.org/)."
+authors = ["Galaxy Project and Community"]
+license = "MIT"
+
+[tool.poetry.dependencies]
+python = "^3.6"
+allure-python-commons = "*"
+bioblend = ">=0.14.0"
```
nsoranzo, line 11:
```suggestion
bioblend = ">=0.16.0"
```

xref. https://github.com/galaxyproject/planemo/pull/1140#pullrequestreview-797695008

jmchilton:
This is a project meant to be used as a library - I don't think it is appropriate to pin all the requirements so exactly like this. It makes it impossible to work with other applications and libraries.

Can you produce a second requirements file for development? Maybe publish a planemo-app to PyPI that depends on Planemo and the exact requirements?

mvdbeek:
I think it would make sense to have planemo be the primary thing we expect our users to install (given all the documentation and training that tells users to install planemo), so maybe we could instead create an unpinned planemo-lib if you want to build apps with planemo ?

jmchilton:
I disagree with this philosophy but I understand the motivation and clearly there is a problem with the current approach of Planemo.

mvdbeek:
@jmchilton I've updated the PR to build a `planemo-lib` package that is unpinned, and a `planemo` package that is pinned. Do you think this is acceptable ?
I'm planning to do the 22.01 release now, and I would feel better about releasing the new packages if we pinned the dependencies.

mvdbeek:
@jmchilton If this isn't acceptable we could mabe pin at least the galaxy-* packages to `< galaxy-<next>` ? I think that would also largely address my concern

nsoranzo:
Superseded by #1230.

PR #1205, opened 2021-11-11
Use WorkflowId rather than StoredWorkflowId when autoupdating subworkflows

simonbray:
I tried autoupdating a couple of workflows containing subworkflows today against usegalaxy.eu and it failed, because a GET to `api/workflows/workflow_id` returns subworkflow WorkflowIds rather than StoredWorkflowIds. When submitting that ID to the API to get the subworkflow details, the `instance` param needs to be set to True.

I was a bit surprised why the test doesn't catch this, because it contains a subworkflow. I think the reason is that on fresh Galaxy instances the Workflow and StoredWorkflow ID is the same for the first workflow uploaded.

### DIFF
--- a/planemo/autoupdate.py
+++ b/planemo/autoupdate.py
@@ -199,15 +199,15 @@ def autoupdate_tool(ctx, tool_path, modified_files=set(), **kwds):
         return perform_required_update(ctx, xml_files, tool_path, requirements, tokens, xml_to_update, wrapper_version_token, **kwds)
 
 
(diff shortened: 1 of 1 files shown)


mvdbeek:
> on fresh Galaxy instances the Workflow and StoredWorkflow ID is the same for the first workflow uploaded.

Yep, that's the problem

mvdbeek (approved):

bgruening:
:shipit:

Issue #1141, opened 2021-02-22
Generate test case from invocation


mvdbeek:
Not sure if that's 100% a planemo task, but we have everything to generate comprehensive test case templates using a previously-run invocation. This might be faster than writing test cases by hand ... although that's actually not too bad either.

PR #1209, opened 2021-12-08
[WIP] Generate workflow test from invocation id

simonbray:
Implementation of #1141:

```
planemo workflow_test_init eed6c2f8d951be9e --from_invocation --profile usegalaxy-eu
```
to create the GA file, tests file and download the test data. I found the process of writing tests for IWC workflows recently quite tricky, I think this would have helped a bit.

### DIFF
--- a/planemo/options.py
+++ b/planemo/options.py
@@ -727,6 +727,16 @@ def split_job_and_test():
     return click.option("--split_test/--no_split_test", default=False, help="Write workflow job and test definitions to separate files.")
 
 
+def from_invocation():
+    return planemo_option(
+        "--from_invocation/--not_from_invocation",
+        is_flag=True,
+        default=False,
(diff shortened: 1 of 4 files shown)


planemo/options.py
```diff
@@ -727,6 +727,16 @@ def split_job_and_test():
     return click.option("--split_test/--no_split_test", default=False, help="Write workflow job and test definitions to separate files.")
 
 
+def from_invocation():
+    return planemo_option(
+        "--from_invocation/--not_from_invocation",
```
jmchilton, line 732: I think ``--from_path`` or ``--from_uri`` instead of ``--not_from_invocation`` feels more natural 😄 .

jmchilton:
Thanks so much for cleaning up the old tests! This is really awesome!

jmchilton:
So awesome - thank you so much.

PR #1212, opened 2022-01-12
Document the use of mandatory macro parameters and add named macro tokens

bernt-matthias:
learned while working on https://github.com/galaxyproject/galaxy/pull/13152

### DIFF
--- a/docs/_writing_macros.rst
+++ b/docs/_writing_macros.rst
@@ -248,6 +248,60 @@ The following expand declarations have replaced the original conditional element
           </options>
         </expand>
 
+From Galaxy release 22.01 named yields are supported. That is, if the macro contains
+``<yield name="xyz"/>`` it is replaced by the content of the ``token` tag with the same name.
+Token tags need to be direct children of the ``expand`` element.
+
+.. code-block:: xml
+
+    <xml name="named_yields_example">
+      <conditional>
+        <param type="select">
+          <option value="a">A</option>
+          <option value="b">B</option>
+          <yield name="more_options"/>
+        </param>
+        <when value="a">
+          <param type="select">
+            <yield />
+          </param>
+        </when>
+        <when value="b">
+          <param ... />
+        </when>
+        <yield name="more_whens">
+      </conditional>
+    </xml>
+
+.. code-block:: xml
+
+        <expand macro="named_yields_example">
+          <token name="more_options">
+            <option value="c">C</option>
+          </token>
+          <token name="more_whens">
+            <when value="c">
+              <param type="select">
+                <yield />
+              </param>
+            </when>
+          </token>
+          <options from_data_table="tophat2_indexes">
+            <filter type="sort_by" column="2"/>
+            <validator type="no_options" message="No genomes are available for the selected input dataset"/>
+          </options>
+        </expand>
+
+Named yields are replaced in the order of the tokens defined in the ``expand``
+tag. Unamed yields are replaced after all named tokens have been replaced. 
+Note that, if there are named yields that have no corresponding token, then
+they are treated like unnamed yields.
+
+Further, note that the order of the replacements offers some possibilities
+to achieve recursion-like replacements, since a token may contain further named
+or unnamed ``yield`` tags.
+
+
 -----------------------------------------
 Parameterizing XML Macros (with tokens)
 -----------------------------------------
--- a/docs/_writing_macros.rst
+++ b/docs/_writing_macros.rst
@@ -348,7 +348,7 @@ named token ``more_whens``).
 Parameterizing XML Macros (with tokens)
 -----------------------------------------
 
-In addition to using ``yield`` blocks, there is another way to parameterize
+In addition to using ``yield`` blocks, there is another way to parametrize
 macros by specifying:
 
 - required parameters as comma-separated list of parameter names using the
--- a/docs/_writing_macros.rst
+++ b/docs/_writing_macros.rst
@@ -249,8 +249,8 @@ The following expand declarations have replaced the original conditional element
    </expand>
 
 From Galaxy release 22.01 named yields are supported. That is, if the macro contains
-``<yield name="xyz"/>`` it is replaced by the content of the ``token`` tag with the same name.
-Token tags need to be direct children of the ``expand`` element. This is useful if different
+``<yield name="xyz"/>`` it is replaced by the content of the ``token`` element with the same name.
+Token elements need to be direct children of the ``expand`` element. This is useful if different
 parts of the macro should be parametrized. 
 
 In the following example two named yield and one unnamed yield are used to
@@ -302,7 +302,7 @@ and expanding the macro in the following way:
 we get the following expanded definition:
 
 .. code-block:: xml
-   :emphasize-lines: 6,9-14,17-24
+   :emphasize-lines: 6,10-13,17-24
 
    <xml name="named_yields_example">
      <conditional>
@@ -332,11 +332,11 @@ we get the following expanded definition:
    </xml>
 
 Named yields are replaced in the order of the tokens defined in the ``expand``
-tag. Unamed yields are replaced after all named tokens have been replaced (by
+tag. Unnamed yields are replaced after all named tokens have been replaced (by
 the non-token child elements of the expand tag). If there are named
 yields that have no corresponding token, then they are treated like unnamed
-yields. Note that, unnamed and named tokens can be used multiple times in a
-macro, then each occurence is preplaced by the corresponding content definedin
+yields. Note that unnamed and named tokens can be used multiple times in a
+macro, then each occurrence is replaced by the corresponding content defined in
 the ``expand``.
 
 Further, note that the order of the replacements offers some possibilities
@@ -396,7 +396,7 @@ is invalid in attribute names). Luckily the delimiting character(s) can be chang
 
    <macros>
      <xml name="color" tokens="attr,attr_value" token_quote="__" token_label="label">
-         <param __VAR__="__ATTR_VALUE__" label="__LABEL__"/>
+         <param __ATTR__="__ATTR_VALUE__" label="__LABEL__"/>
      </xml>
    </macros>
 
--- a/docs/_writing_macros.rst
+++ b/docs/_writing_macros.rst
@@ -19,56 +19,56 @@ block associated with them:
 
 .. code-block:: xml
 
-            <actions>
-              <conditional name="refGenomeSource.genomeSource">
-                <when value="indexed">
-                  <action type="metadata" name="dbkey">
-                    <option type="from_data_table" name="tophat2_indexes" column="1" offset="0">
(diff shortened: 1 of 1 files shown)


docs/_writing_macros.rst
```diff
@@ -253,27 +253,35 @@ Parameterizing XML Macros (with tokens)
 -----------------------------------------
 
 In addition to using ``yield`` blocks, there is another way to parameterize
-macros through the use of specifying ``token_xyz`` attributes on the macro
+macros through the use of specifying 
+
+- required parameters as comma separated list of parameter names using the
```
nsoranzo, line 258:
```suggestion
- required parameters as comma-separated list of parameter names using the
```

docs/_writing_macros.rst
```diff
@@ -253,27 +253,35 @@ Parameterizing XML Macros (with tokens)
...
+``tokens`` attribute of the macro, e.g. ``tokens="foo,bar"`` and then using
+@FOO@ and @BAR@ anywhere in the macro 
+- optional parameters as ``token_xyz`` attributes on the macro
```
nsoranzo, line 261:
```suggestion
- optional parameters as ``token_xyz="default_value"`` attributes of the ``xml`` 
```

docs/_writing_macros.rst
```diff
@@ -253,27 +253,35 @@ Parameterizing XML Macros (with tokens)
...
 
 In addition to using ``yield`` blocks, there is another way to parameterize
-macros through the use of specifying ``token_xyz`` attributes on the macro
+macros through the use of specifying 
+
+- required parameters as comma separated list of parameter names using the
+``tokens`` attribute of the macro, e.g. ``tokens="foo,bar"`` and then using
+@FOO@ and @BAR@ anywhere in the macro 
+- optional parameters as ``token_xyz`` attributes on the macro 
 definition, and then using ``@XYZ@`` anywhere within the XML.
```
nsoranzo, line 262:
```suggestion
  element, and then using ``@XYZ@`` anywhere inside the macro definition.
```

nsoranzo (approved):

docs/_writing_macros.rst
```diff
@@ -248,32 +248,94 @@ The following expand declarations have replaced the original conditional element
...
+          <option value="b">B</option>
+          <yield name="more_options"/>
+        </param>
+        <when value="a">
+          <param type="select">
+            <yield />
+          </param>
+        </when>
+        <when value="b">
+          <param ... />
```
mvdbeek, line 270: I think it helps to have this be somewhat valid XML.

docs/_writing_macros.rst
```diff
@@ -248,32 +248,94 @@ The following expand declarations have replaced the original conditional element
           </options>
         </expand>
 
+From Galaxy release 22.01 named yields are supported. That is, if the macro contains
+``<yield name="xyz"/>`` it is replaced by the content of the ``token` tag with the same name.
+Token tags need to be direct children of the ``expand`` element.
+
```
mvdbeek, line 254: Can we have the example be something like:
Given the following macro:
...
and expanding the macro in the following way:
...
we get the following expanded definition:
...
?

docs/_writing_macros.rst
```diff
@@ -248,32 +248,94 @@ The following expand declarations have replaced the original conditional element
...
+        <when value="b">
+          <param ... />
+        </when>
+        <yield name="more_whens">
+      </conditional>
+    </xml>
+
+.. code-block:: xml
+
+        <expand macro="named_yields_example">
```
mvdbeek, line 278: The indentiation looks weird to me, why not start on the left side ?

docs/_writing_macros.rst
```diff
@@ -248,32 +248,94 @@ The following expand declarations have replaced the original conditional element
...
+            <when value="c">
+              <param type="select">
+                <yield />
+              </param>
+            </when>
+          </token>
+          <options from_data_table="tophat2_indexes">
+            <filter type="sort_by" column="2"/>
+            <validator type="no_options" message="No genomes are available for the selected input dataset"/>
+          </options>
```
mvdbeek, line 292:
```suggestion
```
This is unrelated to the macro example, right ?

docs/_writing_macros.rst
```diff
@@ -248,32 +248,94 @@ The following expand declarations have replaced the original conditional element
           </options>
         </expand>
 
+From Galaxy release 22.01 named yields are supported. That is, if the macro contains
+``<yield name="xyz"/>`` it is replaced by the content of the ``token` tag with the same name.
+Token tags need to be direct children of the ``expand`` element.
```
mvdbeek, line 253:
```suggestion
Token tags need to be direct children of the ``expand`` element.
This is useful when you would like to compose options of a select list, as shown in the example below.
```

docs/_writing_macros.rst
bernt-matthias: The unnamed yield is replaced by this.

mvdbeek: Oh, I have not understood this. Why is that ? How do I read this as a tool author ?

bernt-matthias: Essentially as before: The state was that the content of the expand tag is inserted at the place(s) of the yield tags in the macro. 

In order to keep backward compatibility I implemented it such still all content of the expand minus the token tags and their content is used for the replacement.

mvdbeek: Ah, I see, yeah, that makes sense.

bernt-matthias:
I hope that I have addressed all comment. 

Just had the idea that one could allow content to `yield` tags and interpret it as default if the corresponding token is not given. 
Seems that writing docs is not only annoying but may result in better ideas and helps to reason about code.

bernt-matthias:
Hey @nsoranzo there were quite a few commits after your approval. Are you still OK with merging?

docs/_writing_macros.rst
```diff
@@ -163,124 +163,244 @@ For instance, previously the tophat wrapper contained the following definition:
...
+
+   <expand macro="refGenomeSourceConditional">
+     <options from_data_table="tophat2_indexes">
+       <filter type="sort_by" column="2"/>
+       <validator type="no_options" message="No genomes are available for the selected input dataset"/>
+     </options>
+   </expand>
+
+From Galaxy release 22.01 named yields are supported. That is, if the macro contains
+``<yield name="xyz"/>`` it is replaced by the content of the ``token`` tag with the same name.
```
nsoranzo, line 252:
```suggestion
``<yield name="xyz"/>`` it is replaced by the content of the ``token`` element with the same name.
```

docs/_writing_macros.rst
```diff
@@ -163,124 +163,244 @@ For instance, previously the tophat wrapper contained the following definition:
...
+Token tags need to be direct children of the ``expand`` element. This is useful if different
```
nsoranzo, line 253:
```suggestion
Token elements need to be direct children of the ``expand`` element. This is useful if different
```

docs/_writing_macros.rst
```diff
@@ -163,124 +163,244 @@ For instance, previously the tophat wrapper contained the following definition:
...
+     <options from_data_table="tophat2_indexes">
+       <filter type="sort_by" column="2"/>
+       <validator type="no_options" message="No genomes are available for the selected input dataset"/>
+     </options>
+   </expand>
+
+we get the following expanded definition:
+
+.. code-block:: xml
+   :emphasize-lines: 6,9-14,17-24
```
nsoranzo, line 305:
```suggestion
   :emphasize-lines: 6,10-13,17-24
```

docs/_writing_macros.rst
```diff
@@ -163,124 +163,244 @@ For instance, previously the tophat wrapper contained the following definition:
...
+             <filter type="sort_by" column="2"/>
+             <validator type="no_options" message="No genomes are available for the selected input dataset"/>
+           </options>
+         </param>
+       </when>
+     </conditional>
+   </xml>
+
+Named yields are replaced in the order of the tokens defined in the ``expand``
+tag. Unamed yields are replaced after all named tokens have been replaced (by
```
nsoranzo, line 335:
```suggestion
tag. Unnamed yields are replaced after all named tokens have been replaced (by
```

docs/_writing_macros.rst
```diff
@@ -163,124 +163,244 @@ For instance, previously the tophat wrapper contained the following definition:
...
+the non-token child elements of the expand tag). If there are named
+yields that have no corresponding token, then they are treated like unnamed
+yields. Note that, unnamed and named tokens can be used multiple times in a
+macro, then each occurence is preplaced by the corresponding content definedin
```
nsoranzo, line 339:
```suggestion
macro, then each occurrence is replaced by the corresponding content defined in
```

docs/_writing_macros.rst
```diff
@@ -163,124 +163,244 @@ For instance, previously the tophat wrapper contained the following definition:
...
+         </param>
+       </when>
+     </conditional>
+   </xml>
+
+Named yields are replaced in the order of the tokens defined in the ``expand``
+tag. Unamed yields are replaced after all named tokens have been replaced (by
+the non-token child elements of the expand tag). If there are named
+yields that have no corresponding token, then they are treated like unnamed
+yields. Note that, unnamed and named tokens can be used multiple times in a
```
nsoranzo, line 338:
```suggestion
yields. Note that unnamed and named tokens can be used multiple times in a
```

docs/_writing_macros.rst
```diff
@@ -163,124 +163,244 @@ For instance, previously the tophat wrapper contained the following definition:
...
+Further, note that the order of the replacements offers some possibilities
+to achieve recursion-like replacements, since a token may contain further named
+or unnamed ``yield`` tags (see for instance the ``yield`` tag contained in the
+named token ``more_whens``).
 
 -----------------------------------------
 Parameterizing XML Macros (with tokens)
 -----------------------------------------
 
 In addition to using ``yield`` blocks, there is another way to parameterize
```
nsoranzo, line 351:
```suggestion
In addition to using ``yield`` blocks, there is another way to parametrize
```

docs/_writing_macros.rst
```diff
@@ -163,124 +163,244 @@ For instance, previously the tophat wrapper contained the following definition:
...
 
 .. code-block:: xml
 
-    <inputs>
-        <param name="myvar" type="color" label="Pick a color" value="#ff0000" />
-        <param name="c2" type="color" label="Choose a different color" value="#0000ff" />
-    </inputs>
+   <macros>
+     <xml name="color" tokens="attr,attr_value" token_quote="__" token_label="label">
+         <param __VAR__="__ATTR_VALUE__" label="__LABEL__"/>
```
nsoranzo, line 399:
```suggestion
         <param __ATTR__="__ATTR_VALUE__" label="__LABEL__"/>
```

nsoranzo (approved):

PR #1224, opened 2022-04-01
Update `best_practice_search()` for changes in galaxy-tool-util

nsoranzo:
In https://github.com/galaxyproject/galaxy/pull/13532 I've removed the `channels_override` parameter from the `best_search_result()` method in galaxy-tool-util.

### DIFF
--- a/planemo/conda.py
+++ b/planemo/conda.py
@@ -8,6 +8,7 @@ from __future__ import absolute_import
 import collections
 import os
 import threading
+from copy import deepcopy
 
 from galaxy.tool_util.deps import conda_util
 from galaxy.util import unicodify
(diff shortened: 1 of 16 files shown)


nsoranzo:
According to my git bisecting, the failure of tests/test_external_galaxy_commands.py in unit-quick is due to https://github.com/galaxyproject/galaxy/commit/69f57b39122b8ea52d7ae2c356bc1a752ec7f8a5 , not sure which dependency update and exactly why yet.

nsoranzo:
The unit-quick tests now pass, but the build doesn't finish until it times out after 90'. Maybe a Galaxy process that doesn't stop?

nsoranzo:
Found the issue with tests not terminating!
I see this when testing in my old local virtualenv: `/opt/planemo/tests/shed_app.py:184: UserWarning: The 'environ['werkzeug.server.shutdown']' function is deprecated and will be removed in Werkzeug 2.1.`
The CI tests install Werkzeug 2.1.1 , where this functionality has been removed. Working on a fix now.

jmchilton:
Thank you so much @nsoranzo !

PR #1227, opened 2022-04-19
Make `--channels` also affect mulled channels and update/extend howto use bioconda artifacts

bernt-matthias:
Adapted `--channels` such that it also affects Galaxy's mulled channels config. 

Update docs due to the recent change in bioconda CI (packages and images are now available only as download)

Also document how to set conda channels for mulled container building (see also https://github.com/galaxyproject/galaxy/pull/13755)

### DIFF
--- a/docs/writing_how_do_i.rst
+++ b/docs/writing_how_do_i.rst
@@ -333,20 +333,31 @@ file.
 ----------------------------------------------------------------------------
 
 First, obtain the artifacts of the PR by adding this comment:
-``@BiocondaBot please fetch artifacts``. In the reply one finds the links to the
-built package and docker image. 
+``@BiocondaBot please fetch artifacts``. In the reply one finds the links a zip file containing
+the built package and docker image. Download this zip and extract it. For the following let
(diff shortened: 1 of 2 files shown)


docs/writing_how_do_i.rst
```diff
@@ -333,20 +333,31 @@ file.
...
-     $ planemo test ... --biocontainers --no_conda_auto_init ...
+     $ gzip -dc IMAGE_ZIP | docker load
+
+and a planemo test can then simply use this image::
+
+     $ planemo test ... --biocontainers --no_dependency_resolution --no_conda_auto_init ...
+
+For the later case it suffices to call planemo as follows::
+
+     $ DEFAULT_MULLED_CONDA_CHANNELS="file://PACKAGES_DIR,conda-forge,bioconda" planemo test ... --biocontainers --no_dependency_resolution --no_conda_auto_init ...
```
bernt-matthias, line 360: I would really like if this could be realized with `--conda_channels` (or a new CLI arg like `--mulled_conda_channels`), but I could not find out how.

mvdbeek (approved):

PR #1228, opened 2022-04-20
Drop unused Python dependencies and upgrade syntax to Python 3.6

nsoranzo:
using `ack --type=python -f | xargs pyupgrade --py36-plus`

Also:
- Remove `six` code and requirement
- More f-strings


nsoranzo:
Also, do we want to adopt `black` for code formatting as for Galaxy?
Probably in a follow up PR.

mvdbeek (approved):

PR #1230, opened 2022-05-06
Pin planemo to last known working major galaxy version

mvdbeek:
As an alternative to https://github.com/galaxyproject/planemo/pull/1203.
I'd then proceed to create a new planemo release, release the 22.01 galaxy packages, and bump the pin.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,9 +6,9 @@ configparser
 cwltool>=1.0.20191225192155
 docutils
 ephemeris>=0.10.3
-galaxy-containers
-galaxy-tool-util>=21.1.0.dev4
-galaxy-util>=20.5.0
+galaxy-containers<22.01
+galaxy-tool-util<22.01,>=21.1.0.dev4
+galaxy-util<22.01,>=20.5.0
 glob2
 gxformat2>=0.12.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -6,9 +6,9 @@ configparser
 cwltool>=1.0.20191225192155
 docutils
 ephemeris>=0.10.3
-galaxy-containers
-galaxy-tool-util>=21.1.0.dev4
-galaxy-util>=20.5.0
+galaxy-containers<22.01
+galaxy-tool-util<22.01,>=21.1.0.dev4
+galaxy-util<22.01,>=20.5.0
```
nsoranzo, line 11:
```suggestion
galaxy-tool-util>=21.1.0.dev4,<22.01
galaxy-util>=20.5.0,<22.01
```

I think this is the "standard" way to write the pins.

mvdbeek: That's what I did first, but pip always lists it the other way round (in the pip install output), so I thought that is the standard ...

mvdbeek:
Alright, this is green too @nsoranzo. I've had @jmchilton's OK for this approach

PR #1231, opened 2022-05-08
Format code with black and isort

nsoranzo:


nsoranzo:
I've traced the test failures to commit https://github.com/galaxyproject/galaxy/commit/02ac5b3c9e1b4c6e5344889314ca8fc8948dc071 , but the issue is actually caused by BioBlend and has been already fixed upstream in https://github.com/galaxyproject/bioblend/commit/95a49936aa74e6368dba852aca09e43544eb98de , so I need to make a new BioBlend release and update the minimum version in the Planemo requirements.

nsoranzo:
Green!
@mvdbeek This also contains fixes that should help with your PRs.

PR #1234, opened 2022-05-11
Add Python 3.10 to the build matrix

bgruening:
The idea is to test the lower and upper bound.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.7']
+        python-version: ['3.7', '3.10']
         tox-action:
           - lint
           - lint_docs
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.7']
+        python-version: ['3.7', '3.10']
```
mvdbeek, line 20: That double the long running test time. can we skip the 3.10 tests on everything but `unit-quick` ?

mvdbeek:
Just a note, I've disabled new branch creation to avoid all the extra CI here.
Screenshot 2023-06-21 at 12 11 22

nsoranzo:
> Just a note, I've disabled new branch creation to avoid all the extra CI here. 

Thanks a lot, that works much better than reminding people! I can close and reopen from my fork.

nsoranzo:
> Just a note, I've disabled new branch creation to avoid all the extra CI here.

Thanks for the screenshot, I've done the same for the BioBlend repo!

Issue #999, opened 2020-02-11
Option to limit total test time / individual test time


mvdbeek:
Should help preventing CI timeouts in https://github.com/galaxyproject/tools-iuc/pull/2861

bernt-matthias:
I guess one could use something along the lines of https://stackoverflow.com/questions/366682/how-to-limit-execution-time-of-a-function-call-in-python and apply it here https://github.com/galaxyproject/planemo/blob/3201c415a5d637984620d86b07bb9407d33833b5/planemo/engine/interface.py#L99 ? Just need to find out what a useful response would be. 

Would not cover the else branch in https://github.com/galaxyproject/planemo/blob/master/planemo/engine/test.py

bernt-matthias:
Global timeout could also be achieved with `timeout 500s planemo test ...`

mvdbeek:
This is exposed both in the test interaction and as a global variable (`GALAXY_TEST_DEFAULT_WAIT`, https://github.com/mvdbeek/galaxy/blob/06836c81ce4d6614649c2b73c8cb0f1f348aebaf/lib/galaxy/tool_util/verify/interactor.py#L40), you just need to pass it into the test interactor (https://github.com/mvdbeek/galaxy/blob/06836c81ce4d6614649c2b73c8cb0f1f348aebaf/lib/galaxy/tool_util/verify/interactor.py#L913).

bernt-matthias:
Since this only applies to Galaxy it might be the easiest to run `GALAXY_TEST_DEFAULT_WAIT=XYZ planemo test`?

PR #1236, opened 2022-05-19
add option to limit test run time time

bernt-matthias:
fixes https://github.com/galaxyproject/planemo/issues/999

not sure if we need an option in planemo / if setting the env variable in planemo ci action would be sufficient (but a bit more complex in bash, since it must be checked against zero)

### DIFF
--- a/planemo/options.py
+++ b/planemo/options.py
@@ -1591,6 +1591,12 @@ def test_options():
                 "to disable completely."
             ),
         ),
+        planemo_option(
+            "--test_default_wait",
+            type=int,
+            help="Maximum runtime of a single test in seconds.",
+            default=0,
+        ),
     )
 
 
--- a/planemo/options.py
+++ b/planemo/options.py
@@ -1592,7 +1592,7 @@ def test_options():
             ),
         ),
         planemo_option(
-            "--test_default_wait",
+            "--test_timeout",
             type=int,
             help="Maximum runtime of a single test in seconds.",
             default=0,
--- a/tests/test_cmd_test.py
+++ b/tests/test_cmd_test.py
@@ -11,6 +11,7 @@ from .test_utils import (
     skip_if_environ,
     skip_unless_module,
     TEST_DATA_DIR,
+    TEST_TOOLS_DIR,
 )
 
 DATA_MANAGER_TEST_PATH = "data_manager/data_manager_fetch_genome_dbkeys_all_fasta/data_manager/data_manager_fetch_genome_all_fasta_dbkeys.xml"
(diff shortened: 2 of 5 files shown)


planemo/options.py
```diff
@@ -1591,6 +1591,12 @@ def test_options():
                 "to disable completely."
             ),
         ),
+        planemo_option(
+            "--test_default_wait",
```
mvdbeek, line 1595: How about naming the option `--test_timeout` ?

mvdbeek (approved):

bernt-matthias:
Seems that this is ignored if not run with `run_tests.sh` .. would be good to have a test to see it breaking in the future :) But no idea how and where.

mvdbeek:
Yes, a test would be great. You can add a tool that sleeps for a long time and add a test to `test_cmd_test.py` at the end you can make sure that the test finished faster than the sleep you're using in the tool.

bernt-matthias:
OK. Test added. Seems that currently none of the tests run if `@skip_if_environ("PLANEMO_SKIP_GALAXY_TESTS")` is used... Maybe I'm wrong. If not, I guess this is a bug in CI, or?

This will be my last message before vacation. Feel free to modify this or wait for my return mid June .. Cheers.

tests/test_cmd_test.py
```diff
@@ -53,6 +54,31 @@ def test_workflow_test_simple_yaml(self):
...
+                test_artifact,
+            ]
+            self._check_exit_code(test_command, exit_code=1)
+            with open(json_out.name) as fh:
+                tool_test_json = json.load(fh)
+                assert tool_test_json["summary"]["num_tests"] == 1
+                # check run time, for smaller 10 since the test will take a bit longer than 1s
+                # the important bit is that it's not about 30s (since the test tool calls `sleep 30`)
+                assert (
+                    float(tool_test_json["tests"][0]["data"]["time_seconds"]) > 10
```
nsoranzo, line 76:
```suggestion
                    float(tool_test_json["tests"][0]["data"]["time_seconds"]) <= 10
```

tests/test_cmd_test.py
```diff
@@ -53,6 +54,31 @@ def test_workflow_test_simple_yaml(self):
...
+                ), "Test needed more than 10 sec but should time out after 1"
+                assert (
+                    "Timed out after" not in tool_test_json["tests"][0]["data"]["problem_log"]
```
nsoranzo, line 79:
```suggestion
                    "Timed out after" in tool_test_json["tests"][0]["data"]["problem_log"]
```

nsoranzo:
@mvdbeek The test passes for me with the changes in my review. If you agree they are correct, I think we can apply them and merge the PR without waiting for Matthias to come back.

PR #1242, opened 2022-06-02
Implement nested collection inputs and outputs in workflow_test_init

mvdbeek:
I've been using this to build tests from invocations for some of Anton's workflows, and this is so awesome.

### DIFF
--- a/planemo/galaxy/workflows.py
+++ b/planemo/galaxy/workflows.py
@@ -2,6 +2,12 @@
 import json
 import os
 from collections import namedtuple
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    List,
+)
 from urllib.parse import urlparse
 
 import yaml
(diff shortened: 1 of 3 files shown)


planemo/galaxy/workflows.py
```diff
@@ -364,21 +375,20 @@ def _job_outputs_template_from_invocation(invocation_id, galaxy_url, galaxy_api_
...
+            element_identifier = element["element_identifier"]
+            test_dict = test_dict[element_identifier] = {}
+        file_path = f"test-data/{label}.{element['object'].get('file_ext', 'txt')}"
+        user_gi.datasets.download_dataset(
+            element["object"]["id"],
+            use_default_filename=False,
+            file_path=file_path,
+        )
+        test_dict["element_tests"] = {element_identifier: {"path": file_path}}
+        outputs[label] = {"element_tests": test_dict}
```
simonbray, line 391: To be honest, stuff like `test_dict = test_dict[element_identifier] = {}` seems like black magic to me (it sparked some fascinated discussion in our team chat), but I don't see how this produces a nested output - won't it just create
```
outputs:
  label:
    element_tests:
      element_identifier:
        path: file_path
```
like in https://github.com/galaxyproject/iwc/pull/106?

mvdbeek: You can write this as
```
test_dict["element_identifier"] = {}
test_dict = test_dict["element_identifier"]
```
dicts are mutable in python, so you're building up a nested dict with each level of the collection nesting.

wm75: Ha, glad you're confused by this, too :wink: This would result in a simple empty dict.
The real equivalent would be:
```
test_dict = {}
test_dict["element_identifier"] = test_dict
```
which at least to me makes the recursion much more obvious.

mvdbeek: Y'all, this is really standard python, and that's not the equivalent @wm75:
```

In [1]: outputs = {}

In [2]: outputs['some_output'] = test_dict = {}

In [3]: test_dict["element_identifier"] = {}

In [4]: test_dict = test_dict["element_identifier"]

In [5]: test_dict['nested'] = 'yes'

In [6]: outputs
Out[6]: {'some_output': {'element_identifier': {'nested': 'yes'}}}
```

wm75: That's the point I guess. Your `[3]` and `[4]` are *not* the same as `test_dict = test_dict[element_identifier] = {}`.

mvdbeek: I should have added some tests from the start, added that now in https://github.com/galaxyproject/planemo/pull/1242/commits/e03f75465b8ac761d7df2198dacffc56530320cb and I reworked the whole logic to just be one recursion for generating input and output definitions.

simonbray (approved):

simonbray:
Thanks a lot!

PR #1244, opened 2022-06-07
Removing broken link, update it to current doc

profgiuseppe:
The link pointed to the old wiki, the new link goes to docs. References to the specific section.

### DIFF
--- a/docs/writing_how_do_i.rst
+++ b/docs/writing_how_do_i.rst
@@ -60,7 +60,7 @@ Further reading:
 --------------------------------------------------
 
 Tool parameters support a ``validator`` element (`syntax
-<https://wiki.galaxyproject.org/Admin/Tools/ToolConfigSyntax#A.3Cvalidator.3E_tag_set>`__)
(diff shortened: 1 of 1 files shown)


docs/writing_how_do_i.rst
```diff
@@ -67,7 +67,7 @@ parameters can be performed using arbitrary Python functions using the
 
 Further reading:
 
-- `validator <https://wiki.galaxyproject.org/Admin/Tools/ToolConfigSyntax#A.3Cvalidator.3E_tag_set>`__
+- `validator <https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param-validator>`__
```
nsoranzo, line 70: The same broken link is present at line 63, can you fix that as well please?

Then I think we can just drop this line?

profgiuseppe: Line 63 is up to date now. As a user, I would rather have a bit of redundancy so, if missing the first link, the further reading section will point highlight the doc page again.

PR #1247, opened 2022-06-13
update the dead cuffmerge url

martenson:
note: this links to shared macros instead of a specific tool


nsoranzo:
Maybe the link to a macro file can be confusing for novice tool developers? We can just drop it since there are 2 other examples.

martenson:
I considered that, but this is part of the advance tool writing section, and the example is useful I think.

nsoranzo (approved):

PR #1265, opened 2022-08-23
Allow running autoupdate against external server

mvdbeek:
and fix tests on OSX


mvdbeek:
I'd call this mostly a debugging feature, but still neat to have the option.

nsoranzo (approved):

PR #1280, opened 2022-10-06
Planemo type annotation: module cmd_normalize

gallardoalba:


planemo/commands/cmd_normalize.py
```diff
@@ -7,6 +7,7 @@
     load_tool,
     raw_tool_xml_tree,
 )
+from lxml.etree import _Element
```
nsoranzo, line 10:
```suggestion
from galaxy.util import Element
```

planemo/commands/cmd_normalize.py
```diff
@@ -84,7 +85,7 @@ def cli(ctx, path, expand_macros=False, **kwds):
     ElementTree.dump(root)
 
 
-def _indent(elem, level=0):
+def _indent(elem: _Element, level: int = 0) -> None:
```
nsoranzo, line 88:
```suggestion
def _indent(elem: Element, level: int = 0) -> None:
```

nsoranzo (approved):
Thanks @gallardoalba !

PR #1291, opened 2022-10-11
Add type annotations to ``planemo.runnable`` and ``planemo.workflow_lint``

nsoranzo:


planemo/runnable.py
```diff
@@ -468,7 +508,7 @@ def _tests_path(runnable):
     return None
 
 
-def get_outputs(runnable, gi=None):
+def get_outputs(runnable: Runnable, gi: Optional[GalaxyInstance] = None) -> List["RunnableOutput"]:
```
mvdbeek, line 511: Should this be so tightly typed ? Would a Protocol be better here ? gi needs to implement `workflows.export_workflow_dict`, if that comes from another library that is ok? There's some work in galaxy_test.base.populators in that direction.

nsoranzo: I'd say we cross that bridge when we get to it?
The type annotation can always be updated when we support other libraries.

mvdbeek: Hmm, but that means a downstream user of `get_outputs` would need to PR changes to planemo first before they can do type checking against their own code. If we want planemo to be a library I think this is important.

PR #1300, opened 2022-10-25
supply "kwds" to for_runnable_identifier() directly

jancrichter:
Trying the workflow_edit command I was getting an error that was caused by the fact that, kwds.get("profile") was called here AND in for_runnable_identifier(). With this change workflow_edit successfully boots up the Galaxy instance. 

Unfortunately, I am not yet getting anything different then I would get via the serve command. There is no workflow and the tools used in the workflow aren't installed either.


mvdbeek:
Sorry that it slipped through the cracks ... https://github.com/galaxyproject/planemo/pull/1427 contains the same fix.

PR #1302, opened 2022-10-28
Use `pytest.raises()` instead of ad-hoc `assert_raises_regexp()` context manager

nsoranzo:
Fix `ModuleNotFoundError: No module named 'py.code'; 'py' is not a package` when importing from `tests/test_utils.py` .

Also add type annotations to training material tests.


martenson (approved):
much cleaner, thanks @nsoranzo

PR #1303, opened 2022-10-29
Restore running tool tests against directory

mvdbeek:
This was handled by the old tool testing framework in Galaxy previously. We should prioritize switching to embedded pulsar, so we don't need to copy tool sources around ... .

### DIFF
--- a/planemo/runnable.py
+++ b/planemo/runnable.py
@@ -49,6 +49,7 @@ from planemo.test import (
     check_output,
     for_collections,
 )
+from planemo.tools import yield_tool_sources_on_paths
 
 TEST_SUFFIXES = ["-tests", "_tests", "-test", "_test"]
 TEST_EXTENSIONS = [".yml", ".yaml", ".json"]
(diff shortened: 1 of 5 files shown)


tests/test_cmd_test.py
```diff
@@ -30,6 +34,16 @@ def test_startup_timeout(self):
             )
             self._check_exit_code(test_command, exit_code=1)
 
+    @skip_if_environ("PLANEMO_SKIP_GALAXY_TESTS")
+    def test_tool_in_directory(self):
+        """Test with (single) tool in directory."""
+        with self._isolate():
+            test_artifact = os.path.join(TEST_DATA_DIR, "tools", "ok_test_assert_command.xml")
+            tempdir = mkdtemp()
```
nsoranzo, line 42: Could use the `TemporaryDirectory()` context manager instead.

mvdbeek: Good idea, I'll do that

planemo/runnable.py
```diff
@@ -225,7 +236,7 @@ def for_path(path: str, temp_path: Optional[str] = None, return_all: bool = Fals
     """Produce a class:`Runnable` for supplied path."""
     runnable_type = None
     if os.path.isdir(path):
-        runnable = workflow_dir_runnables(path, return_all=return_all)
+        runnable = workflow_dir_runnables(path, return_all=return_all) or tool_dir_runnables(path, temp_path)
         if runnable:
             return runnable
         runnable_type = RunnableType.directory
```
nsoranzo, line 242: Should we raise an exception if no runnable is found, or is there still a use case for continuing with the rest of the function?

mvdbeek: I think RunnableType.directory is a valid runnable, and we raise appropriate exceptions if a engine can't deal with a runnable. I think this is fine ?

nsoranzo (approved):

PR #1307, opened 2022-10-31
Fix recording of virtual_env_dir

mvdbeek:
Fixes stopping gravity in Galaxy converter tests.

I had previously assumed we can just carry around `GALAXY_VIRTUAL_ENV` via the kwds, but that never worked.

### DIFF
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -1003,10 +1003,11 @@ class LocalGalaxyConfig(BaseManagedGalaxyConfig):
             kwds,
         )
         self.galaxy_root = galaxy_root
+        self._virtual_env_dir = None
 
     @property
     def virtual_env_dir(self):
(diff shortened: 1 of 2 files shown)


mvdbeek:
You can see this in action in https://github.com/galaxyproject/galaxy/actions/runs/3362267720/jobs/5573772869, where it fixes the shutdown

planemo/galaxy/config.py
```diff
@@ -1003,10 +1003,11 @@ def __init__(
             kwds,
         )
         self.galaxy_root = galaxy_root
+        self._virtual_env_dir = None
 
     @property
     def virtual_env_dir(self):
-        virtual_env = self._kwds.get("GALAXY_VIRTUAL_ENV", ".venv")
+        virtual_env = self._virtual_env_dir or ".venv"
         if virtual_env and not os.path.isabs(virtual_env):
```
nsoranzo, line 1011:
```suggestion
        if not os.path.isabs(virtual_env):
```

Issue #1309, opened 2022-11-01
[BUG] The new handling of the asserts section forgot a possible syntax


lldelisle:
Hi,
One of the IWC workflow lint raise error because of this line:
https://github.com/galaxyproject/planemo/blob/483f310cdbe9cf5b57a0ae774581b7ad7f53d5c4/planemo/workflow_lint.py#L266

See: https://github.com/galaxyproject/iwc/actions/runs/3357665520/jobs/5580414418

The tests responsible for the fail are:
https://github.com/galaxyproject/iwc/blob/fbdefc404c7abe8b9a4b5b8a2a4b25ad8f38ecb9/workflows/epigenetics/atacseq/atacseq-tests.yml#L113-L118

Indeed when specifying multiple tests of the same kind, the only way to have them all taken into account is this syntax.
This is documented here: https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/simple_constructs.yml
And the parser code is here:
https://github.com/galaxyproject/galaxy/blob/d30b8659eebb38cc4aff1111d3549be043bf374d/lib/galaxy/tool_util/parser/yaml.py#L267-L292

Also, if we modify it, maybe it would be good to also lint the asserts in the elements of a collection but this may come in a second wave.

lldelisle:
I am trying to write a PR about that. I hope nobody started.

mvdbeek:
That would be great, thank you @lldelisle!

mvdbeek:
Should be fixed in https://github.com/galaxyproject/planemo/pull/1310, thanks @lldelisle!

PR #1310, opened 2022-11-02
Fix workflow_lint with list + check elements in collection

lldelisle:
This should solve #1309 
I would like to write tests. Should I make one directory per test with workflow and test file inside ? In the `tests/data/wf_repos`?
Thanks

### DIFF
--- a/planemo/workflow_lint.py
+++ b/planemo/workflow_lint.py
@@ -245,9 +245,17 @@ def _lint_case(path: str, test_case: TestCase, lint_context: WorkflowLintContext
             found_valid_expectation = True
         # TODO: validate structure of test expectations
 
-        assertion_definitions = test_case.output_expectations[test_output_id].get("asserts")
-        if not _check_test_assertions(lint_context, assertion_definitions):
-            test_valid = False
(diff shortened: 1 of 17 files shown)


mvdbeek:
Yes, I think that's fine.

lldelisle:
We ignore if tests are failing on the dev branch of galaxy right?

mvdbeek:
Yes, those failing tests are probably related to one of the recent conda changes in Galaxy. Looks like there are some problems with local conda channels.

mvdbeek (approved):
That is super helpful, thank you so much @lldelisle!

nsoranzo:
> Yes, those failing tests are probably related to one of the recent conda changes in Galaxy. Looks like there are some problems with local conda channels.

I tried to replicate this locally with:

```shell
PYTEST_TARGET=tests/test_cmd_test_conda.py::CmdTestCondaTestCase::test_local_conda_dependencies_version tox -e py37-unit-nonredundant-noclientbuild-noshed-gx-dev
```

but it passes for me even after removing `~/miniconda3/` .

Issue #1311, opened 2022-11-02
[BUG] between version 0.75.0 and 0.75.1 on autoupdate


lldelisle:
I may have done something wrong because:
https://github.com/planemo-autoupdate/autoupdate/actions/runs/3376419931/jobs/5604120700#step:6:462
With version 0.75.0:
```
(.venv3) [ldelisle@SV-49-002 TrimNs]$ planemo autoupdate .
Auto-updating tool /tmp/test/tools-iuc/tools/TrimNs/TrimNs.xml
Tool version does not contain @TOOL_VERSION@ as required by autoupdate.
```
With version 0.75.1:
```
(.venv3) [ldelisle@SV-49-002 TrimNs]$ planemo autoupdate .
Traceback (most recent call last):
  File "/tmp/test/.venv3/bin/planemo", line 8, in <module>
    sys.exit(planemo())
  File "/tmp/test/.venv3/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/tmp/test/.venv3/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/tmp/test/.venv3/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/tmp/test/.venv3/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/tmp/test/.venv3/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/tmp/test/.venv3/lib/python3.10/site-packages/click/decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/tmp/test/.venv3/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/tmp/test/.venv3/lib/python3.10/site-packages/planemo/cli.py", line 102, in handle_blended_options
    return f(*args, **kwds)
  File "/tmp/test/.venv3/lib/python3.10/site-packages/planemo/commands/cmd_autoupdate.py", line 95, in cli
    if any(r.type in {RunnableType.galaxy_tool, RunnableType.directory} for r in runnables):
[...]

PR #1312, opened 2022-11-02
Fix ``for_paths`` when path is directory of tools

mvdbeek:
We'd previously return the directory type runnable, but with 0.75.1 we're expanding the directory into a list of workflow or tool runnables.
This makes it match the return type annotation.

Fixes https://github.com/galaxyproject/planemo/issues/1311

### DIFF
--- a/planemo/runnable.py
+++ b/planemo/runnable.py
@@ -189,7 +189,14 @@ def workflow_dir_runnables(path: str) -> List[Runnable]:
 
 
 def tool_dir_runnables(path: str) -> List[Runnable]:
-    return [for_path(p) for (p, _) in yield_tool_sources_on_paths(ctx=None, paths=[path])]
+    runnables = []
+    for tool_path, _ in yield_tool_sources_on_paths(ctx=None, paths=[path]):
+        runnable = for_path(tool_path)
+        if isinstance(runnable, list):
+            runnables.extend(runnable)
+        else:
+            runnables.append(runnable)
+    return runnables
 
 
 def for_path(path: str) -> Union[Runnable, List[Runnable]]:
--- a/planemo/runnable.py
+++ b/planemo/runnable.py
@@ -247,7 +247,14 @@ def for_path(path: str, return_all: bool = False) -> Union[Runnable, List[Runnab
 
 def for_paths(paths: Iterable[str]) -> List[Runnable]:
     """Return a specialized list of Runnable objects for paths."""
-    return [for_path(path) for path in paths]
+    runnables = []
+    for path in paths:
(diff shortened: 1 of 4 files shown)


nsoranzo:
That means we should also drop the `@overload` type annotations for `for_path()`, since now it can return a `List[Runnable]` also when `return_all=False`.

mvdbeek:
I guess we can also drop the return_all argument ...

planemo/runnable.py
```diff
@@ -181,38 +181,25 @@ def workflows_from_dockstore_yaml(path):
     return workflows
 
 
-def workflow_dir_runnables(path: str, return_all: bool = False) -> Optional[Union[Runnable, List[Runnable]]]:
+def workflow_dir_runnables(path: str) -> Optional[List[Runnable]]:
```
nsoranzo, line 182:
```suggestion
def workflow_dir_runnables(path: str) -> List[Runnable]:
```

planemo/runnable.py
```diff
@@ -181,38 +181,25 @@ def workflows_from_dockstore_yaml(path):
...
     dockstore_path = os.path.join(path, DOCKSTORE_REGISTRY_CONF)
     if os.path.exists(dockstore_path):
         runnables = [
```
nsoranzo, line 185:
```suggestion
        return [
```

planemo/runnable.py
```diff
@@ -181,38 +179,25 @@ def workflows_from_dockstore_yaml(path):
...
             Runnable(str(path), RunnableType.galaxy_workflow) for path in workflows_from_dockstore_yaml(dockstore_path)
         ]
-        if return_all:
-            return runnables
-        else:
-            return runnables[0]
+        return runnables
     return None
```
nsoranzo, line 189:
```suggestion
    return []
```

planemo/runnable.py
```diff
@@ -181,38 +179,31 @@ def workflows_from_dockstore_yaml(path):
...
-def for_path(path: str, return_all: bool = False) -> Union[Runnable, List[Runnable]]:
-    pass
+    runnables = []
+    for tool_path, _ in yield_tool_sources_on_paths(ctx=None, paths=[path]):
+        runnable = for_path(tool_path)
+        if isinstance(runnable, list):
+            runnables.extend(runnable)
+        else:
+            runnables.append(runnable)
+    return runnables
```
nsoranzo, line 199:
```suggestion
    return for_paths(tool_path for tool_path, _ in yield_tool_sources_on_paths(ctx=None, paths=[path]))
```

nsoranzo (approved):

PR #1313, opened 2022-11-02
Compare versions, not tool ids to find latest tool ids

mvdbeek:
Fixes upgrades that are actually downgrades, like https://github.com/galaxyproject/iwc/pull/148


nsoranzo (approved):

lldelisle:
Thanks

PR #1320, opened 2022-11-10
Type annotation for input staging-related code

nsoranzo:


jmchilton (approved):

planemo/galaxy/activity.py
```diff
@@ -246,25 +277,18 @@ def stage_in(ctx, runnable, config, job_path, **kwds):  # noqa C901
     )
 
     if datasets and kwds.get("check_uploads_ok", True):
-        ctx.vlog("uploaded datasets [%s] for activity, checking history state" % datasets)
+        ctx.vlog(f"Uploaded datasets [{datasets}] for activity, checking history state")
         final_state = _wait_for_history(ctx, user_gi, history_id)
-
-        for (dataset, path) in datasets:
```
mvdbeek, line 252: Maybe this should stay, but do `if ctx.verbose` ?

mvdbeek (approved):

planemo/galaxy/activity.py
nsoranzo: In https://github.com/galaxyproject/galaxy/pull/14963/files#diff-bf58520ed2b280d7e27f164c623c82481eb0de4a37d195dc59f57b10c9861a8fR156 I've changed the second member of the tuple returned by `galactic_job_json()` / `StagingInterface.stage()` to be a list of dataset dictionaries instead of a (dataset dict, UploadTarget), so this looked more effort than it's worth to support.

nsoranzo:
I have added a few more annotations and bug fixes, it's green now if you'd like to take another look before merging.

mvdbeek (approved):

PR #1329, opened 2022-11-22
Fix ``planemo shed_test``

mvdbeek:
And potentially speed up shed installs by not requesting the entire tool shed contents ...

### DIFF
--- a/planemo/shed/interface.py
+++ b/planemo/shed/interface.py
@@ -18,7 +18,7 @@ def find_repository(tsi, owner, name):
     """Find repository information for given owner and repository
     name.
     """
-    repos = tsi.repositories.get_repositories()
+    repos = tsi.repositories.get_repositories(owner=owner, name=name)
 
     def matches(r):
         return r["owner"] == owner and r["name"] == name
--- a/tests/test_cmd_shed_test.py
+++ b/tests/test_cmd_shed_test.py
@@ -20,7 +20,7 @@ class CmdShedTestTestCase(CliTestCase):
                 "--name",
                 "collection_element_identifiers",
                 "--shed_target",
-                "https://toolshed.g2.bx.psu.edu/",
+                "toolshed",
             )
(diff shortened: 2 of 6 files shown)


planemo/shed/interface.py
```diff
@@ -18,7 +18,7 @@ def find_repository(tsi, owner, name):
     """Find repository information for given owner and repository
     name.
     """
-    repos = tsi.repositories.get_repositories()
+    repos = tsi.repositories.get_repositories(owner=owner, name=name)
```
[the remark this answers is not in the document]
mvdbeek: (yep, also noticed some more wrong things in this test!)

tests/test_cmd_shed_test.py
```diff
@@ -0,0 +1,31 @@
...
+        """Test --galaxy_startup_timeout."""
+        with self._isolate():
+            test_command = self._test_command(
+                "--no_dependency_resolution",
+                "--owner",
+                "iuc",
+                "--name",
+                "collection_element_identifiers",
+                "--shed_target",
+                "https://toolshed.g2.bx.psu.edu/",
```
nsoranzo, line 23:
```suggestion
                "toolshed",
```

tests/test_cmd_shed_test.py
```diff
@@ -0,0 +1,31 @@
...
+        """Test --galaxy_startup_timeout."""
+        with self._isolate():
+            test_command = self._test_command(
+                "--no_dependency_resolution",
+                "--owner",
+                "iuc",
+                "--name",
+                "collection_element_identifiers",
+                "--shed_target",
+                "toolshed",
```
mvdbeek, line 23: hmm, that doesn't work of course, but I guess it should ideally.

mvdbeek: Should work with 3c71f5c32a1959de6d654000eca5154d95a7e463

PR #1308, opened 2022-11-01
Drop copy_tree workaround for tool sources

mvdbeek:
It comes with a whole bunch of disadvantages and complexitites, and we can also just dereference symlinks in planemo-ci-action.

PR #1327, opened 2022-11-21
Mount test data dir read-only in docker

mvdbeek:
Fixes a bunch of data manager tests that are failing the weekly iuc tests. I think this started failing after we stopped copying the whole tool directory to tmp.

The additional `--docker_extra_volume` change let's us mount the whole tool repo for tool tests, which would fix test using symlinks to scripts outside of the tool directory, and for the iwc we can just mount in the whole of cvmfs, which means we don't need to use `--tool_data_path` (which brings additional things in, like watching for changes, which I think is unnecessary in a CI context).

PR #1336, opened 2022-12-08
Add a ro mount of the tool path's LCA

bernt-matthias:
compute the lowest common ancestor of

- the tool dirs
- the targets of all contained directories

and add it to the list of ro mounted volumes

May fix some of the currently failing CI tests: https://github.com/galaxyproject/tools-iuc/issues/4936 Problem is tools that use symlinks to other tools which does not work anymore due to https://github.com/galaxyproject/planemo/pull/1308

Not sure if this is a good way to fix

- [ ] should the same (or something analoguous) be done where I added the TODO?

### DIFF
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -11,6 +11,7 @@ import shlex
 import shutil
 import threading
 import time
+from pathlib import Path
 from string import Template
 from tempfile import (
     mkdtemp,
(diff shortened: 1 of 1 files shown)


mvdbeek:
I don't think that's the right fix, how about `planemo test ----docker_extra_volume .` in the planemo-ci-action test command ?

mvdbeek:
(https://github.com/galaxyproject/planemo/pull/1327)

bernt-matthias:
> I don't think that's the right fix

Thought so :) ... wanted to make progress on this anyway :)

> how about planemo test --docker_extra_volume . in the planemo-ci-action test command

But this would need to be done by any `planemo test` command using containerized tests. 

Maybe we can use `.` as default for `--docker_extra_volume` (is this a parameter that can be specified multiple times)?

mvdbeek:
> But this would need to be done by any planemo test command using containerized tests.

Only for those with a symlink to another tool. I am aware of only a single tool, ruvseq that does that.

mvdbeek:
Or are you talking about pre 0.75.3 behavior ? We do mount the dir RO for the current runnable.

bernt-matthias:
Superseded by https://github.com/galaxyproject/planemo/pull/1327

PR #1347, opened 2023-01-22
store datasets by UUID

bernt-matthias:
thereby datasets get different IDs in different test sessions

this will make tests more robust in the long run. currently datasets get the same ID in each test (containing the same set of tests). since in weekly CI tests this set is different than in PR tests tools may fail in weekly tests that just were successful in PR tests.


mvdbeek (approved):

mvdbeek:
Thank you! Odd that didn't get used as default ...

PR #1348, opened 2023-01-24
fix orcid regex for dockstore_init

lldelisle:
~In addition the test with `requests.get` is useless:
`requests.get("https://orcid.org/0000-0002-1964-496", headers={"Accept": "application/xml"})` gives <Response [200]> even if the orcid is invalid.~
I found how to fix it.


lldelisle:
The CI is failing but I don't think this is due to my changes...

nsoranzo (approved):

nsoranzo:
> The CI is failing but I don't think this is due to my changes...

A restart fixed it.

PR #1314, opened 2022-11-02
Add creator dockstore

lldelisle:
The goal is to propagate creator info from workflows to dockstore.yml.

Feel free to modify the code because I am not super happy about it.

PR #1350, opened 2023-02-10
Make test dockstore init with creator run and pass

lldelisle:
- In #1314 and #1348 I forgot to put a `def test` so the dockstore with creator was not tested...
- I allow redirect because:
https://github.com/galaxyproject/iwc/pull/174#discussion_r1102064590
But strangely, if you put a 16 numbers which does not match a real orcid to get 404 but if you put 15 or 17 numbers it gives 200 and bring you to the login page:
```
>>> requests.get(f"https://orcid.org/0000-0001-6879-5194", headers={"Accept": "application/xml"},).status_code
200
>>> requests.get(f"https://orcid.org/0000-0001-6879-5193", headers={"Accept": "application/xml"},).status_code
404
>>> requests.get(f"https://orcid.org/0000-0001-6879-519", headers={"Accept": "application/xml"},).status_code
200
```

### DIFF
--- a/planemo/workflow_lint.py
+++ b/planemo/workflow_lint.py
@@ -99,7 +99,7 @@ def generate_dockstore_yaml(directory: str, publish: bool = True) -> str:
                                 requests.get(
                                     f"https://orcid.org/{orcid[0]}",
                                     headers={"Accept": "application/xml"},
-                                    allow_redirects=False,
+                                    allow_redirects=True,
                                 ).status_code
                                 == 200
                             ):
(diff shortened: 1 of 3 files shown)


mvdbeek:
> * if you put 15 or 17 numbers

we could also check that on our side, I assume orchid is always going to be 16 ?

lldelisle:
Now with the regex it should be ok. It was just that when I did a mistake with the regex it was passing while I was expecting to fail... Nevermind. I think this is the good solution, I just need to fix the test case.

mvdbeek (approved):
I tried fixing up the tests, if that still looks good to you @lldelisle I will merge

lldelisle:
Thanks I forgot I was working on this.

lldelisle:
One test failed and I don't have the power to rerun it.

mvdbeek:
The node download failed, bit of a problem recently, but nothign related to this PR

lldelisle:
cool ! sorry that I needed 3 PR to make it work...

mvdbeek:
No worries, would've taken me the same amount of PRs, or more ;)

PR #1351, opened 2023-02-13
Improve shed install for workflows

mvdbeek:


jmchilton:
Claude thinks other things have replaced a lot of this other than the error message and it thinks the error message isn't better per se anymore.

> That's a strictly redundant restatement of a line already on screen, minus the cause. I also looked for a place where the exception text survives but the log doesn't — a report file, a structured error, a handler that prints only str(e) — and there isn't one. It propagates as a normal traceback.

So given that and its age and conflicts - I'm going to close it out. Feel free to reopen and yell at me if still a priority! -John

PR #1356, opened 2023-03-01
Pin allure-python-commons version

tuncK:
The new release of v. 2.13.0 makes planemo crash:

...
File "/home/ubuntu/miniconda3/envs/planemo/lib/python3.8/site-packages/planemo/reports/[allure.py](http://allure.py/)", line 19, in <module>
from allure_commons.utils import (
ImportError: cannot import name 'escape_non_unicode_symbols' from 'allure_commons.utils' (/home/ubuntu/miniconda3/envs/planemo/lib/python3.8/site-packages/allure_commons/[utils.py](http://utils.py/))

PR #1357, opened 2023-03-01
Drop call to ``escape_non_unicode_symbols``

nsoranzo:
which was removed from allure-python-commons in 2.13.0 .

Alternative to https://github.com/galaxyproject/planemo/pull/1356 .

### DIFF
--- a/planemo/reports/allure.py
+++ b/planemo/reports/allure.py
@@ -17,7 +17,6 @@ from allure_commons.types import (
     LinkType,
 )
 from allure_commons.utils import (
-    escape_non_unicode_symbols,
     md5,
     platform_label,
     uuid4,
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):

nsoranzo:
Looking at the error message, I think the failing test is due to https://github.com/galaxyproject/galaxy/pull/15497 , ping @nuwang .

mvdbeek:
Thanks @nsoranzo!

nsoranzo:
Hi @nuwang , thanks for getting back! Yes, I think so, see this test run (line 2546): https://github.com/galaxyproject/planemo/actions/runs/4365276747/jobs/7633682776#step:8:2547

```
galaxy.tool_util.provided_metadata INFO 2023-03-08 14:56:58,940 [pN:main.1,p:10159,tN:LocalRunner.work_thread-0] One or more tool outputs is marked as failed ({'type': 'dataset', 'ext': 'data', 'dataset_id': 1, 'stderr': 'Unable to fetch file:///home/runner/work/planemo/planemo/tests/data/hello.txt\nCould not find handler for URI [file:///home/runner/work/planemo/planemo/tests/data/hello.txt]', 'failed': True}).
```

It seems the `file://` protocol is not correctly handled. If you can take a look, that would be great!

nuwang:
I think I see where the issue is originating from and adding file:// as a supported protocol should restore previous behaviour. I'll create a PR with a fix.

nsoranzo:
> I think I see where the issue is originating from and adding file:// as a supported protocol should restore previous behaviour. I'll create a PR with a fix.

That's great, thanks @nuwang !

PR #1364, opened 2023-04-14
fix regex of orcid + add test

lldelisle:
Sorry I just discovered thanks to @hrhotz that orcid can end with X.


nsoranzo (approved):

mvdbeek:
Thanks @lldelisle !

PR #1369, opened 2023-05-23
Increase bioblend retries for bad networks

hexylena:
This caused me some issues and I've been manually editing this file for a while so here finally is a PR for it.

I've picked 10, it seems like a nice number. We could make it configurable but for now having any retries is better than none and I struggle to find where I should apply the configuration option.

cc @natefoo who has the same complaint.


natefoo (approved):

hexylena:
sure @mvdbeek

PR #1378, opened 2023-07-17
Optimize disk space usage for `planemo test`

bernt-matthias:
Had CI tests that failed due to full disk and noticed that the temporary Galaxy dir has ~1GB. 

- So I thought a shallow clone from the bare repository in `.planemo` might be a good idea: 135cef81b954fe72e8aeccc9abf58a5e01e0c59e
- Then I discovered `--no_cache_galaxy` which might be even better for CI jobs, but noticed that it does not support forks and branches of the galaxy sources: b08209e409fa28099c4d38f5d65cdd4646fe9952 .. also it did not work anymore as shown here https://github.com/galaxyproject/planemo-ci-action/pull/50: `/bin/sh: 1: ./scripts/common_startup.sh: not found`

### DIFF
--- a/planemo/git.py
+++ b/planemo/git.py
@@ -98,6 +98,9 @@ def command_clone(
         cmd.append("--mirror")
     if branch is not None:
         cmd.extend(["--branch", branch])
+        cmd.extend(["--depth", "1"])
+        if not src.startswith("file://"):
+            src = f"file://{src}"
     cmd.extend([src, dest])
     return cmd
 
--- a/planemo/git.py
+++ b/planemo/git.py
@@ -87,7 +87,7 @@ def checkout(ctx, remote_repo, local_path, branch=None, remote="origin", from_br
 
 
 def command_clone(
-    ctx: "PlanemoCliContext", src: str, dest: str, mirror: bool = False, branch: Optional[str] = None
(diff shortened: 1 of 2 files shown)


planemo/git.py
```diff
@@ -98,6 +98,9 @@ def command_clone(
         cmd.append("--mirror")
     if branch is not None:
         cmd.extend(["--branch", branch])
+        cmd.extend(["--depth", "1"])
+        if not src.startswith("file://"):
+            src = f"file://{src}"
```
nsoranzo, line 103: Why is this needed?
In case, I think this should be done by the caller of this function.

bernt-matthias: https://stackoverflow.com/questions/40383230/git-clone-warning-depth-is-ignored-in-local-clones-use-file-instead

nsoranzo: Ah, thanks! In that case it could be conditional on `depth is not None` and not being a URL, but maybe not worth the effort.

planemo/git.py
```diff
@@ -2,6 +2,7 @@
 
 import os
 import subprocess
+import urllib
```
nsoranzo, line 5:
```suggestion
import urllib.parse
```

nsoranzo (approved):

nsoranzo:
Failing test is unrelated and will be fixed by https://github.com/galaxyproject/planemo/pull/1379 .

nsoranzo:
Can you rebase this one please?

PR #1379, opened 2023-07-19
Fix for change of base_url in BioBlend 1.2.0

nsoranzo:


bernt-matthias (commented):
Since having an extra `/` should not hurt this should also work with older versions of bioblend, or?

nsoranzo:
> Since having an extra `/` should not hurt this should also work with older versions of bioblend, or?

Yes, I wrote this patch so that it should work whether the base URL ends with a "/" or not.

PR #1381, opened 2023-08-04
Run galaxy with default tools included

bernt-matthias:
I wanted to test some Galaxy tools and needed the collection operations. So I thought it might be a good idea to serve planemo with Galaxy's default tools installed:

Screenshot from 2023-08-04 15-59-25

Not sure what happens if some one calls planemo test/serve on one of the builtin tools....

Alternatively one could use `--extra_tools ` but then the tools are just installed flat in the tool menu.

### DIFF
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -25,6 +25,7 @@ from typing import (
     Set,
     TYPE_CHECKING,
 )
+from xml.etree import ElementTree
 
 from galaxy.tool_util.deps import docker_util
 from galaxy.tool_util.deps.container_volumes import DockerVolume
@@ -269,7 +270,7 @@ def docker_galaxy_config(ctx, runnables, for_tests=False, **kwds):
         )
         tool_config_file = f"{tool_conf},{shed_tool_conf}"
 
-        _write_tool_conf(ctx, all_tool_paths, tool_conf)
+        _write_tool_conf(ctx, all_tool_paths, tool_conf, galaxy_root)
         write_file(empty_tool_conf, EMPTY_TOOL_CONF_TEMPLATE)
 
         properties.update(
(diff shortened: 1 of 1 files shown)


planemo/galaxy/config.py
```diff
@@ -1369,15 +1371,24 @@ def _handle_job_config_file(
...
-def _write_tool_conf(ctx, tool_paths, tool_conf_path):
-    tool_definition = _tool_conf_entry_for(tool_paths)
-    tool_conf_template_kwds = dict(tool_definition=tool_definition)
-    tool_conf_contents = _sub(TOOL_CONF_TEMPLATE, tool_conf_template_kwds)
-    write_file(tool_conf_path, tool_conf_contents)
+def _write_tool_conf(ctx, tool_paths: List[str], tool_conf_path: str, galaxy_root: str):
+    ctx.vlog(f"_write_tool_conf tool_paths {tool_paths}")
+    ctx.vlog(f"_write_tool_conf tool_conf_path {tool_conf_path}")
+    ctx.vlog(f"_write_tool_conf galaxy_root {galaxy_root}")
+
```
nsoranzo, line 1378: Debug?
```suggestion
```

mvdbeek:
Can you expose a `--tool_conf` argument instead (and check that `--shed_tool_conf` doesn't do the job already) ?

mvdbeek:
Fine, but then at least make this optional ?

jmchilton:
Could we have --extra_tools be able to take toolbox conf in addition to actual tools and directory. Also I guess we use the name stock tools in the workflow docs for planemo - so maybe add like a flag ``--include_stock_tools`` that is short-hand for enabling the behavior here (--extra_tools with a the tool conf sample as the target).

bernt-matthias:
> Can you expose a `--tool_conf` argument instead (and check that `--shed_tool_conf` doesn't do the job already) ?

Seems that `--shed_tool_conf` already does the trick (at least with `--galaxy_root` and using the sample tool conf from this directory).

bernt-matthias:
> (I have no local Galaxy, if this matters)

Just clone the galaxy sources (maybe the latest release branch). 

> How exactly do I get the correct parameter to pass to --shed_tool_conf when running planemo serve?

If I remember correctly `--galaxy_root GALAXY_SOURCE_DIR --shed_tool_conf GALAXY_SOURCE_DIR/config/tool_conf.xml.sample` should do the trick.

PR #1382, opened 2023-08-04
fix type comparisons

bernt-matthias:
linter complained about this over here https://github.com/galaxyproject/planemo/pull/1381

Not sure if `isinstance` or `is` is correct?

### DIFF
--- a/planemo/workflow_lint.py
+++ b/planemo/workflow_lint.py
@@ -191,12 +191,12 @@ def _lint_best_practices(path: str, lint_context: WorkflowLintContext) -> None: 
     """
 
     def check_json_for_untyped_params(j):
-        values = j if type(j) == list else j.values()
+        values = j if isinstance(j, list) else j.values()
         for value in values:
             if type(value) in [list, dict, OrderedDict]:
                 if check_json_for_untyped_params(value):
                     return True
-            elif type(value) == str:
+            elif isinstance(value, str):
                 if re.match(r"\$\{.+?\}", value):
                     return True
         return False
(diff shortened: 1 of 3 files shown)


planemo/workflow_lint.py
```diff
@@ -191,12 +191,12 @@ def _lint_best_practices(path: str, lint_context: WorkflowLintContext) -> None:
     """
 
     def check_json_for_untyped_params(j):
-        values = j if type(j) == list else j.values()
+        values = j if isinstance(j, list) else j.values()
```
mvdbeek, line 194: Can you check this is a dict instance instead ? that should be less narrow. Also the check in line 196 isn't really how you'd do this

bernt-matthias: > Can you check this is a dict instance instead ? 

So `is TYPE` everywhere?

> Also the check in line 196 isn't really how you'd do this

Also with `is`?

mvdbeek: No, type is type is the most narrow of them all which will not work for any subclasses. `.values()` will only work for objects that satisfy Mapping, while any j that is Iterable will work. So check for the specific instance where you need to call `.values()` first `isinstance(j, Mapping)` and don't worry about the other types that will all be fine.

bernt-matthias, line 194:
```suggestion
        values = j.values() if isinstance(j, dict) else j
```

mvdbeek (commented):
Thank you!

mvdbeek (approved):

PR #1383, opened 2023-08-23
fix profile version test

bernt-matthias:
linter has been adapted to new style profile versions and got a bit more relaxed.

https://github.com/galaxyproject/galaxy/pull/16480

Alternatively we could work on the linter in Galaxy, e.g. something along:

```python
try:
    profile_year, profile_month =  profile.split(".")
except Exception:
    handle_error
if int(profile_year) < 10:
    handle_error
if int(profile_year) < 23:
    if not profile_month in ["01", "02", "03", "04", ... ]:
        handle_error
else:
    if not re.match(profile_month, r"\d"):
        handle_error
```


mvdbeek (approved):
Thank you!

PR #1388, opened 2023-09-14
Update upper bound for galaxy packages to < 23.2

bernt-matthias:
... waiting for the  23.1 release :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,8 +3,8 @@ bioblend>=1.0.0
 click!=8.0.2
 cwltool>=1.0.20191225192155
 ephemeris>=0.10.3
-galaxy-tool-util>=23.0,<23.1
-galaxy-util[template]>=23.0,<23.1
+galaxy-tool-util>=23.0,<24.0
+galaxy-util[template]>=23.0,<24.0
 glob2
 gxformat2>=0.14.0
 h5py
--- a/planemo/commands/cmd_conda_build.py
+++ b/planemo/commands/cmd_conda_build.py
@@ -24,5 +24,7 @@ def cli(ctx: "PlanemoCliContext", paths: Tuple[str], **kwds) -> None:
     # Force conda_use_local for building...
     kwds["conda_use_local"] = True
     conda_context = build_conda_context(ctx, handle_auto_init=True, **kwds)
-    if conda_context.exec_command("build", paths) != 0:
+    exit_code = conda_context.exec_command("build", paths)
+    if exit_code:
         error(f"Failed to build [{' '.join(paths)}] with conda.")
+    ctx.exit(exit_code)
--- a/planemo/commands/cmd_conda_install.py
+++ b/planemo/commands/cmd_conda_install.py
@@ -28,4 +28,5 @@ def cli(ctx, paths, **kwds):
             conda_target, conda_context=conda_context, skip_environment=kwds.get("global", False)
         )
         return_codes.append(return_code)
-    return coalesce_return_codes(return_codes, assert_at_least_one=True)
+    exit_code = coalesce_return_codes(return_codes, assert_at_least_one=True)
+    ctx.exit(exit_code)
--- a/planemo/commands/cmd_conda_search.py
+++ b/planemo/commands/cmd_conda_search.py
@@ -30,4 +30,5 @@ def cli(ctx, term, **kwds):
     if conda_context.conda_version >= VERSION_4_DOT_4:
         term = "*%s*" % term
     args = conda_context._override_channels_args + [term]
-    conda_context.exec_command("search", args)
+    exit_code = conda_context.exec_command("search", args)
+    ctx.exit(exit_code)


requirements.txt
```diff
@@ -3,8 +3,8 @@ bioblend>=1.0.0
 click!=8.0.2
 cwltool>=1.0.20191225192155
 ephemeris>=0.10.3
-galaxy-tool-util>=23.0,<23.1
-galaxy-util[template]>=23.0,<23.1
+galaxy-tool-util>=23.0,<24.0
+galaxy-util[template]>=23.0,<24.0
```
nsoranzo, line 7: I don't think we have officially ruled out a 23.2 release.

```suggestion
galaxy-tool-util>=23.0,<23.2
galaxy-util[template]>=23.0,<23.2
```

bernt-matthias:
Test is failing reproducibly. But not sure if related.

bgruening:
The error I can see is:

```
E           PackagesNotFoundError: The following packages are not available from current channels:
E           
E             - fleeqtk
E           
E           Current channels:
E           
E             - https://conda.anaconda.org/conda-forge/linux-64
E             - https://conda.anaconda.org/conda-forge/noarch
E             - https://conda.anaconda.org/bioconda/linux-64
E             - https://conda.anaconda.org/bioconda/noarch
E             - https://repo.anaconda.com/pkgs/main/linux-64
E             - https://repo.anaconda.com/pkgs/main/noarch
E             - https://repo.anaconda.com/pkgs/r/linux-64
E             - https://repo.anaconda.com/pkgs/r/noarch
```

Which is true. This packages does not exist, its a fake package that is build locally during a test.

I think `--conda_use_local` is not working in this PR.

bernt-matthias:
Bumping conda helps with the failing test. Let's see if there are side effects https://github.com/galaxyproject/galaxy/pull/17013

nsoranzo:
I am not sure how https://github.com/galaxyproject/galaxy/pull/17013 can help, it bumps the *minimum* conda version but in tests it's most likely that the latest conda version is installed either way.

bernt-matthias:
All I know is that in the failing test on my local computer not the latest was installed, but:

```
           conda version : 23.3.1
    conda-build version : 3.27.0
```

When I bumped the minimum requirement the tests were successful. No idea how to test this here, without bumping the Galaxy packages first. Can we somehow patch tool-util during the test?

nsoranzo:
> All I know is that in the failing test on my local computer not the latest was installed, but:
> 
> ```
>            conda version : 23.3.1
>     conda-build version : 3.27.0
> ```

Is that what you have in `~/miniconda3` ?

> When I bumped the minimum requirement the tests were successful. No idea how to test this here, without bumping the Galaxy packages first. Can we somehow patch tool-util during the test?

Did you update the conda package manually? Changing `CONDA_PACKAGE_SPECS` only affects new conda installations.

nsoranzo:
What's weird when looking at the CI logs is that `test_local_conda_dependencies_version` failed at the `planemo test` stage, when the `__fleeqtk@1.3` conda environment should have been created already by `planemo conda_install` right before.

bernt-matthias:
> What's weird when looking at the CI logs is.. 

Yes. They had exit code 0, despite the failure.

mvdbeek:
Can you check in the github workflow if there is any interference from the conda or .condarc already installed on the VM ?

bernt-matthias:
So, now we see the error from the build step. Seems that this installs 23.3.1 .. my hypothesis is as follows: 

- the [mambaforge installer](https://github.com/galaxyproject/galaxy/blob/2319bc31af19726c4f8596621f1e01950da6f0c8/lib/galaxy/tool_util/deps/conda_util.py#L67) comes with preinstalled conda 23.3.1 (just verified)
- when installing conda use the two-step approach
  1. try to fulfill requirements such that installed packages don't need to be updated
  2. if this is not possible fulfill requirements and allow updating installed packages

so with our current conda pins, we are stuck at 23.3.1 (maybe also because we install in the base env?) and not the newest packages are installed. 

Does this sound reasonable?

Note that also `planemo conda_search` seems not to use the exit code of the `exec_command` function. If desired I could try to fix it here.

nsoranzo:
Thanks @bernt-matthias , the latest commit to check the exit code really helped clarifying the issue, and your explanation makes sense. I didn't expect the latest Mambaforge to install conda 23.3.1, but any way it looks like that https://github.com/galaxyproject/galaxy/pull/17013 is the correct solution here.

> Note that also `planemo conda_search` seems not to use the exit code of the `exec_command` function. If desired I could try to fix it here.

Yes, that would be good, thanks!

nsoranzo:
planemo/commands/cmd_conda_install.py also has the same issue.

bernt-matthias:
Does anyone have an idea what's going on with the 3.7 tests?

mvdbeek:
dev doesn't work on 3.7, that's why I rebased your PR

mvdbeek (approved):
Very cool, thanks @bernt-matthias

PR #1391, opened 2023-10-05
Try to fix planemo test workflow when output is collection with identical name

lldelisle:
See https://github.com/galaxyproject/iwc/pull/225


lldelisle:
Any idea why the test is not failing?
What did I do wrong?

mvdbeek:
It would have run in https://github.com/galaxyproject/planemo/actions/runs/6421040087/job/17434490532, but that was cancelled.

mvdbeek:
And yeah, it's failing locally for me too and my commit fixes this.

lldelisle:
Great!
Should I try to fix the CI into another PR?

mvdbeek:
I think this is conda timing out, that probably needs to be fixed on the Galaxy or Conda side ...

lldelisle:
Thanks for the fix.

mvdbeek:
Thanks for the test, that made it easy!

PR #1399, opened 2023-10-26
Fix zenodo test, drop explicit datatype mapping

mvdbeek:
This isn't supported by the zenodo api anymore


nsoranzo:
FYI @hexylena @shiltemann @bebatut

mvdbeek:
Can we move this ahead ? Would be nice to get the tests green again.

jmchilton (approved):

Issue #1290, opened 2022-10-11
planemo test workflow using invocation id


lldelisle:
Hi,
I don't  know if this already exists. It would be good to be able to test the yaml `*-tests.yml` against an existing invocation id. This would avoid rerunning twice the same workflow with the same inputs just to check if your tests are good.

mvdbeek:
It'll only work against external servers, but seems like a good idea!
Could be `planemo assert_tests tests.yml  --invocation_id  <invocation_id> --galaxy_url <galaxy_url>`.

Might also be useful for when you started with `planemo run` ...

wm75:
the "start" part of @hexylena's suggestion could potentially be implemented with the existing --no_wait option if that returned the invocation id?

nomadscientist:
This sounds awesome! Yes please do this!

PR #1401, opened 2023-10-30
Implement running tests against existing invocation

mvdbeek:
Addresses https://github.com/galaxyproject/planemo/issues/1290

### DIFF
--- a/planemo/commands/cmd_workflow_test_check.py
+++ b/planemo/commands/cmd_workflow_test_check.py
@@ -1,0 +1,42 @@
+"""Module describing the planemo ``workflow_test_check`` command."""
+import click
+
+from planemo import options
+from planemo.cli import command_function
+from planemo.engine.factory import engine_context
+from planemo.galaxy.activity import invocation_to_run_response
+from planemo.galaxy.test.actions import handle_reports_and_summary
+from planemo.runnable import definition_to_test_case
+from planemo.runnable_resolve import for_runnable_identifier
+from planemo.test.results import StructuredData
+
+
+@click.command("workflow_test_check")
+@options.optional_tools_arg(multiple=False, allow_uris=False, metavar="TEST DEFINITION")
+@options.required_workflow_arg()
+@options.galaxy_url_option(required=True)
+@options.galaxy_user_key_option(required=True)
+@options.test_index_option()
+@options.test_options()
+@command_function
+def cli(ctx, path, workflow_identifier, test_index, **kwds):
+    """Run defined tests against existing workflow invocation."""
+    with engine_context(ctx, engine="external_galaxy", **kwds) as engine, engine.ensure_runnables_served([]) as config:
+        user_gi = config.user_gi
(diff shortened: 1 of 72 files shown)


planemo/options.py
```diff
@@ -1599,6 +1601,10 @@ def profile_database_options():
     )
 
 
+def test_index_option():
+    return planemo_option("--test_index", default=1, type=int, help="Select which test to run. Counting starts at 1")
```
nsoranzo, line 1605:
```suggestion
    return planemo_option("--test_index", default=1, type=int, help="Select which test to check. Counting starts at 1")
```

Also I'm wondering if `1` is a good default or if it would be better to re-check all tests by default?

mvdbeek: We can't, since one test creates one invocation

planemo/commands/cmd_workflow_test_check.py
```diff
@@ -0,0 +1,42 @@
...
+from planemo.cli import command_function
+from planemo.engine.factory import engine_context
+from planemo.galaxy.activity import invocation_to_run_response
+from planemo.galaxy.test.actions import handle_reports_and_summary
+from planemo.runnable import definition_to_test_case
+from planemo.runnable_resolve import for_runnable_identifier
+from planemo.test.results import StructuredData
+
+
+@click.command("workflow_test_check")
```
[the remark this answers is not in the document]
mvdbeek: thank you, that does sound better!

planemo/runnable.py
```diff
@@ -265,6 +265,12 @@ def cases(runnable: Runnable) -> List["AbstractTestCase"]:
                 cases.append(ExternalGalaxyToolTestCase(runnable, tool_id, tool_version, i, test_dict))
         return cases
 
+    return definition_to_test_case(tests_path=tests_path, runnable=runnable)
+
+
+def definition_to_test_case(tests_path, runnable):
```
nsoranzo, line 271: Can you add typing here please?

planemo/options.py
```diff
@@ -1607,7 +1613,6 @@ def test_options():
             help="Update test-data directory with job outputs (normally"
             " written to directory --job_output_files if specified.)",
         ),
-        paste_test_data_paths_option(),
```
nsoranzo, line 1610: Not needed any more?

planemo/options.py
nsoranzo: Marius kindly explained to me that this is not about multiple assertions for the same workflow invocation, but when defining tests for multiple workflows in the same file.

Are you OK with the change in wording suggested above though?

mvdbeek: Yes, that's much better!

planemo/options.py
mvdbeek: It is, I didn't mean to remove that from all test like commands

nsoranzo (approved):

lldelisle:
Youhou ! Thanks @mvdbeek

Issue #1403, opened 2023-11-13
planemo autoupdate report bug


lldelisle:
It is not the first time it happens:
https://github.com/galaxyproject/iwc/pull/276
While there is no error in the autoupdate itself, there is a bug in the report:

> toolshed.g2.bx.psu.edu/repos/devteam/picard/picard_MarkDuplicates/2.18.2.4 should be updated to toolshed.g2.bx.psu.edu/repos/devteam/picard/picard_MarkDuplicatesWithMateCigar/2.18.2.3

nsoranzo:
Will hopefully be fixed by #1406.

lldelisle:
Would be great. Thank you very much for taking care of this.

PR #1406, opened 2023-11-13
Fix matching of tool ids to autoupdate

nsoranzo:
Fix https://github.com/galaxyproject/planemo/issues/1403 .

Also:
- Refactor ``outdated_tools()`` to process tools only once, and not process tools in ``tools_to_skip`` at all.


mvdbeek (approved):
Thank you!

PR #1408, opened 2023-11-15
pretty-print workflow and use correct version of linked workflow

mvdbeek:
in  ``planemo workflow_test_init --from_invocation``. One item from @Delphine-L's wishlist.

### DIFF
--- a/planemo/galaxy/workflows.py
+++ b/planemo/galaxy/workflows.py
@@ -312,10 +312,9 @@ def get_workflow_from_invocation_id(invocation_id, galaxy_url, galaxy_api_key):
     workflow_id = user_gi.invocations.show_invocation(invocation_id)["workflow_id"]
     workflow = user_gi.workflows._get(workflow_id, params={"instance": "true"})
     workflow_name = "-".join(workflow["name"].split())
-    user_gi.workflows.export_workflow_to_local_path(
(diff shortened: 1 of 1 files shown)


planemo/galaxy/workflows.py
```diff
@@ -312,10 +312,9 @@ def get_workflow_from_invocation_id(invocation_id, galaxy_url, galaxy_api_key):
     workflow_id = user_gi.invocations.show_invocation(invocation_id)["workflow_id"]
     workflow = user_gi.workflows._get(workflow_id, params={"instance": "true"})
     workflow_name = "-".join(workflow["name"].split())
-    user_gi.workflows.export_workflow_to_local_path(
-        use_default_filename=False, file_local_path=f"./{workflow_name}.ga", workflow_id=workflow["id"]
-    )
-
+    export_dict = user_gi.workflows.export_workflow_dict(workflow_id=workflow["id"], version=workflow["version"])
```
mvdbeek, line 315: note that that fixes a bug ... `workflow_id=workflow["id"]` would get the latest version associated with the stored workflow, even if that's not what the invocation ran against.

lldelisle: Ouch... Good catch.

lldelisle (approved):

lldelisle:
Maybe you could rename your PR: 'pretty-print workflow and use good version of workflow'

PR #1417, opened 2023-12-08
Validate with jsonschema

mvdbeek:
also
- fixes the wrong and extremely verbose input connection check (we used to print every step dict!)
- fixes gxformat2 linting if gxformat2 steps are lists and not dicts
- make unknown job inputs an error (cause you may think you're testing something that you're not)
- allow `elements` in top level collection tests in addition to `element_tests` (i've marked element_tests as deprecated in the schema)

The schema disallows the `- that: has_text` style assertions. They continue to work in tests but results in linting errrors, I hope that's ok. I've included the schema for now in planemo, but eventually we should either include it in galaxy-tool-util (as soon as pydantic 2 can be supported) or just publish the schema to https://www.schemastore.org/json/. The process seems easy enough, but we may need to come up with a more unique test file name (*.gxwf-test.yml) ?

### DIFF
--- a/planemo/schema/test_file_schema.json
+++ b/planemo/schema/test_file_schema.json
@@ -1543,36 +1543,43 @@
       "type": "object"
     },
     "Job": {
-      "additionalProperties": {
-        "anyOf": [
-          {
-            "type": "string"
+      "anyOf": [
+        {
+          "additionalProperties": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "integer"
+              },
+              {
+                "type": "number"
+              },
+              {
+                "type": "boolean"
+              },
+              {
+                "$ref": "#/$defs/Collection"
+              },
+              {
+                "$ref": "#/$defs/LocationFile"
+              },
+              {
+                "$ref": "#/$defs/PathFile"
+              },
+              {
+                "type": "null"
+              }
+            ]
           },
-          {
-            "type": "integer"
-          },
-          {
-            "type": "number"
-          },
-          {
-            "type": "boolean"
-          },
-          {
-            "$ref": "#/$defs/Collection"
-          },
-          {
-            "$ref": "#/$defs/LocationFile"
-          },
-          {
-            "$ref": "#/$defs/PathFile"
-          },
-          {
-            "type": "null"
-          }
-        ]
-      },
-      "title": "Job",
-      "type": "object"
+          "type": "object"
+        },
+        {
+          "type": "string"
+        }
+      ],
+      "title": "Job"
     },
     "LocationFile": {
       "additionalProperties": false,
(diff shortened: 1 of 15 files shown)


jmchilton:
Can you document some where how ``planemo/schema/test_file_schema.json`` is generated? Is that present and I missed it?

tests/data/wf_repos/basic_format2_collection_wrong_assert/basic_format2.gxwf-test.yml
```diff
@@ -21,10 +21,10 @@
               text: Word
         second:
           asserts:
-            - that: has_text
```
jmchilton, line 24: Does the version of this where these isn't a list - but it is just a dictionary keyed on the assertion types validate? That seems to be what we have documented in: https://planemo.readthedocs.io/en/latest/test_format.html#test-format. I assume yes but I wanted to make sure.

I definitely prefer the a flat list of dictionaries with the fixed ``that`` keyword to a list of nested dictionaries each with a single outer key. The structure of this feels very... awkward or forced. Is this really the syntax you prefer or is JSON schema coercing us in this direction?

jmchilton (approved):

planemo/schema/test_file_schema.json
```diff
@@ -0,0 +1,3973 @@
...
+        }
+      },
+      "title": "AssertHasNlines",
+      "type": "object"
+    },
+    "AssertHasSize": {
+      "additionalProperties": false,
+      "description": "Asserts the output has a specific size (in bytes) of ``value`` plus minus\n``delta``, e.g. ``&lt;has_size value=\"10000\" delta=\"100\" /&gt;``.\n\nAlternatively the range of the expected size can be specified by ``min`` and/or\n``max``.\n$attribute_list::5",
+      "properties": {
+        "value": {
```
bernt-matthias, line 1059: Can you also add `size` ass synonym for `value` (as in https://github.com/galaxyproject/galaxy/pull/17490)

planemo/schema/test_file_schema.json
```diff
@@ -0,0 +1,3973 @@
...
+      "properties": {
+        "has_size": {
+          "items": {
+            "$ref": "#/$defs/AssertHasSize"
+          },
+          "title": "Has Size",
+          "type": "array"
+        },
```
bernt-matthias, line 224: Instead of having all these assertions as properties we should have `"children"`, which contain a list of assertions, or. 

At least that's what the parser does : https://github.com/galaxyproject/galaxy/blob/9ddb6ad44afc59fadec39f5bdbf8e6b106774734/lib/galaxy/tool_util/parser/yaml.py#L291 .. or lets say is supposed to do (since the recursion is missing at the moment .. I guess this is easy to fix and I will try).

In hindsight it might have been better to just use `"asserts"` instead of `"children"` since its also the name of the top level list of asserts (or is this only in workflow tests).

mvdbeek: Just need to regenerate the schema, it's all derived from the xsd

mvdbeek: That was a comment on https://github.com/galaxyproject/planemo/pull/1417#discussion_r1516816920 ...

I am not sure I understand the proposition here, this doesn't seem necessary as you can simply use a list of assertions ?

jmchilton:
Closing - we have validation via Pydantic models now in gxwf - but maybe not in Planemo yet? If not though, we should redo this work on top the upstream Galaxy work.

Issue #1418, opened 2023-12-14
readthedocs is out of date


lldelisle:
Hi,
The documentation on readthedocs seems pretty old: https://planemo.readthedocs.io/en/latest/history.html
Could you fix it?
Thanks

PR #1419, opened 2023-12-15
Add now mandatory readthedocs config files

nsoranzo:
Docs builds currently fail with:

```
Problem in your project's configuration. No default configuration file found at repository's root. See https://docs.readthedocs.io/en/stable/config-file/
```

See e.g. https://readthedocs.org/projects/planemo/builds/22785619/

Fix https://github.com/galaxyproject/planemo/issues/1418 .


mvdbeek:
Thanks a lot!

PR #1424, opened 2024-01-30
Fix use of `package_name` attribute on `CondaTarget` objects

nsoranzo:
In Galaxy 23.1 `galaxy.tool_util.deps.mulled.util.build_target()` was changed to return a `CondaTarget` object instead of a (removed) `Target`.

To avoid complications, this requires a bump to the galaxy-util and galaxy-tool-util packages to >=23.1 .

Also, add type annotations to most related code.

Fix https://github.com/galaxyproject/galaxy/issues/17378 .

### DIFF
--- a/planemo/commands/cmd_container_register.py
+++ b/planemo/commands/cmd_container_register.py
@@ -1,6 +1,7 @@
 """Module describing the planemo ``container_register`` command."""
 import os
 import string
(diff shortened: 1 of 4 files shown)


mvdbeek (approved):

mvdbeek:
Thank you!

PR #1426, opened 2024-01-31
Improve docs with regards to running and testing workflows

wm75:


wm75:
@mvdbeek you asked me to take a look at the documentation of `planemo run`.

I've added some smaller improvements here, but imo the biggest hurdle when trying to understand the documentation is not so much the content of individual chapters, but their outline.
Depending on your entry point into the docs they can be *very* confusing. If you take a look at the sidebar, there is little logic in (but likely a lot of history behind) how chapters are organized and the titles are often misleading.
I would really start there.

wm75:
More specifically, I'd ask what different things you can do with planemo and what people coming to the docs might be looking for: something like develop/test tools, interact with workflows, interacting with the tool shed, then use these as main topics and add sub-chapters for each.

wm75:
I've noticed that the simple `planemo workflow_edit tutorial.ga` suggested early in https://planemo.readthedocs.io/en/latest/running.html#the-basics does not work without some profile information, presumably because you cannot access workflow editor functionality as an anonymous user?
Suggestions how to rewrite this would be appreciated.

mvdbeek:
I'll try fixing that up, however I think we should have people testing workflows from the IWC, not https://github.com/usegalaxy-eu/workflow-testing, which I think isn't very active or easily debuggable.

mvdbeek:
The fix was pretty simple, no profile required, I'll merge and cut a new release once the tests pass.

mvdbeek:
Sorry, got distracted a little. https://pypi.org/project/planemo/0.75.21/ should work for workflow_edit, though we should enable the prebuilt-client install to speed up planemo serve.

jmchilton:
I've rebased this with the fixes that are still good fixes - and there is some great discussion here that widens the scope. I've created issues for those two things to keep this PR and the good fixes moving while still tracking the important work to do. Those issues are #1719  and #1718.

jmchilton (approved):

PR #1427, opened 2024-02-01
Fix workflow_edit

mvdbeek:
Should fix part of https://github.com/galaxyproject/planemo/pull/1426#issuecomment-1920919019

### DIFF
--- a/planemo/commands/cmd_workflow_edit.py
+++ b/planemo/commands/cmd_workflow_edit.py
@@ -18,7 +18,7 @@ from planemo.runnable_resolve import for_runnable_identifier
 def cli(ctx, workflow_identifier, output=None, force=False, **kwds):
     """Open a synchronized Galaxy workflow editor."""
     assert is_galaxy_engine(**kwds)
-    runnable = for_runnable_identifier(ctx, workflow_identifier, kwds.get("profile"))
+    runnable = for_runnable_identifier(ctx, workflow_identifier, kwds)
 
     kwds["workflows_from_path"] = True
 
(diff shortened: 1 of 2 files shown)


nsoranzo (approved):

planemo/commands/cmd_workflow_edit.py
```diff
@@ -18,14 +18,14 @@
...
-    runnable = for_runnable_identifier(ctx, workflow_identifier, kwds.get("profile"))
+    runnable = for_runnable_identifier(ctx, workflow_identifier, kwds)
 
     kwds["workflows_from_path"] = True
 
     with engine_context(ctx, **kwds) as galaxy_engine:
         with galaxy_engine.ensure_runnables_served([runnable]) as config:
             workflow_id = config.workflow_id_for_runnable(runnable)
-            url = f"{config.galaxy_url}/workflow/editor?id={workflow_id}"
+            url = f"{config.galaxy_url}/workflows/edit?id={workflow_id}"
```
nsoranzo, line 28: Does this break with old Galaxy releases?

mvdbeek: Yes, but I'm gonna say that change has happened about 2 years ago ? And it's been broken since, so my thinking is it's not worth making that conditional on the version ?

nsoranzo: Fine by me!

nsoranzo:
Linting errors will be fixed by https://github.com/galaxyproject/planemo/pull/1425

Issue #1429, opened 2024-02-09
Planemo serve: --host not utilised


selten:
- Operating system: Ubuntu 22.04 in Docker
- Setup details: Planemo and Galaxy are both installed in a docker container, this container is ran to allow 'planemo serve', 'planemo test' etc. However, for planemo serve we set --host to 0.0.0.0 for accessing the container outside of Docker. In the newer versions of Galaxy (22.05, 23.0, etc.) this is not respected by planemo.
- Steps to reproduce:
  1.  Setup planemo
  2. Setup Galaxy >= 22.05
  3. Run planemo serve with --host

mvdbeek:
Should be fixed in https://github.com/galaxyproject/planemo/pull/1430, thanks @selten!

PR #1430, opened 2024-02-09
#1429 Allow --host parameter

selten:
Don't set a static host when it is able to be provided by planemo, use the provided value if available


mvdbeek (approved):
Looks good, thanks a lot!

PR #1434, opened 2024-02-29
Addition of metadata generation for planemo run

Delphine-L:
Add the command `--output_metadata FILE` to planemo run. Generate a json file containing the workflow ID, invocation ID, and history ID created by the 'run' command.

### DIFF
--- a/docs/commands/run.rst
+++ b/docs/commands/run.rst
@@ -168,6 +168,8 @@ Planemo command for running tools and jobs.
                                       Where to store outputs of a 'run' task.
       --output_json FILE              Where to store JSON dictionary describing
                                       outputs of a 'run' task.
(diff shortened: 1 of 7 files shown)


bernt-matthias:
Would it be possible to add a test? Maybe here: https://github.com/galaxyproject/planemo/blob/5a0eb3185a02f71f6312b680b67e8bf0deae7b8d/tests/test_run.py

simonbray:
I think either `--test_output_json` or `--output_json` should provide this information already.

Delphine-L:
> I think either `--test_output_json` or `--output_json` should provide this information already.

You're right. The info is in --test_output_json. Thank you !

jmchilton:
> You're right. The info is in --test_output_json. Thank you !

Closing this out - thanks!

PR #1437, opened 2024-03-10
Misc workflow linting improvements

bernt-matthias:
- fix linting of assertions with children
- fix linting: do not warn about missing inputs for input nodes.

### DIFF
--- a/planemo/workflow_lint.py
+++ b/planemo/workflow_lint.py
@@ -334,8 +334,14 @@ def _check_test_assertions(
                 assertions_valid = False
                 continue
             signature = inspect.signature(function)
+            function_args = inspect.getfullargspec(function)
             assertion_params = assertion_description["attributes"].copy()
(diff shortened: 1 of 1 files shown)


planemo/workflow_lint.py
```diff
@@ -334,8 +334,14 @@ def _check_test_assertions(
                 assertions_valid = False
                 continue
             signature = inspect.signature(function)
+            function_args = inspect.getfullargspec(function)
             assertion_params = assertion_description["attributes"].copy()
+            if "verify_assertions_function" in function_args:
+                assertion_params["verify_assertions_function"] = asserts.verify_assertion(b"", [])
+            if "children" in function_args:
+                assertion_params["children"] = []
```
mvdbeek, line 342: Any chance you could add a test for that ?

bernt-matthias: Yes I should do this.

PR #1438, opened 2024-03-19
Use proper tmpdir for jobfiles

bernt-matthias:
While doing workflow tests I noticed that `json` files are created in the code dir. Thought it might be better to use a proper temporary dir.

### DIFF
--- a/planemo/engine/interface.py
+++ b/planemo/engine/interface.py
@@ -2,7 +2,7 @@
 
 import abc
 import json
-import os
+import shutil
 import tempfile
 from typing import (
     Callable,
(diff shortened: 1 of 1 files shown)


bernt-matthias:
> Test failures seem relevant.

Yes. But no idea what goes wrong .. and if I have the time to fix it.

jmchilton:
Superseded by #1707. The replacement keeps this PR’s goal of avoiding generated job files in source directories, while preserving relative test-data paths and adding focused cleanup and path-resolution coverage. Thank you for identifying and starting this improvement.

Issue #1415, opened 2023-12-06
planemo training_init creates maintainers not editorial_board


patrick-austin:
Based on galaxyproject/training-material/pull/4273 it seems that the metadata.yaml should be initialised with `editorial_board` rather than `maintainers` as is done here:
https://github.com/galaxyproject/planemo/blob/5951e90ca4e1eb5059a9b1da5cb9aadae4618e8f/planemo/training/topic.py#L115

Don't know if anything else needs to change as part of this (or if this is a planemo issue versus GTN issue).

hexylena:
Ah most people aren't contributing new topics so we hadn't noticed this. Yeah it's just a simple rename.

PR #1439, opened 2024-03-25
Fix parsing of training ``metadata.yaml`` files

nsoranzo:
Fix https://github.com/galaxyproject/planemo/issues/1415 .

### DIFF
--- a/planemo/training/__init__.py
+++ b/planemo/training/__init__.py
@@ -37,8 +37,9 @@ class Training:
             if not self.tuto.exists():
                 info(f"The tutorial {self.tuto.name} in topic {self.topic.name} does not exist. It will be created.")
                 self.tuto.create_tutorial(ctx)
-        info("WARNING: Change the contributors/maintainers listed in the metadata of the new training ")
-        info("before serving the website to fit the one listed in the CONTRIBUTORS.yaml file")
+        info(
+            "WARNING: Change the contributors listed in the metadata of the new training before serving the website to fit the one listed in the CONTRIBUTORS.yaml file"
+        )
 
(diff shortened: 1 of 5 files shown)


bernt-matthias (commented):
Do we test or should we somehow test that the generated yaml file is valid (wrt to a schema)? Does GTN has one?  Should we request one?

I kind of do not understood why those internal (variable) names need to be renamed instead of just changing the display or giving a proper definition. But this is a different question.

bernt-matthias (approved):
Apart from this more general question the PR seems fine.

hexylena:
> Do we test or should we somehow test that the generated yaml file is valid (wrt to a schema)? Does GTN has one? Should we request one?

the gtn has one, yes. but we are also considering alternative implementations that would be more integrated into the GTN, so, i'm not sure if it's worth the effort.

the schema is here: https://github.com/galaxyproject/training-material/blob/main/bin/schema-tutorial.yaml, but it does some non-standard things (namely replacing keywords like `CONTRIBUTORS` with their associated key upon parsing, and the 'examples' key that isn't part of kwalify, because those lists are also controlled but in separate data sources.)

We could emit a proper schema at an api endpoint that would be up-to-date if need be. 

> I kind of do not understood why those internal (variable) names need to be renamed

for me it would've been equivalent, yes, but i can see wanting to be sure all references are updated correctly.

nsoranzo:
> > I kind of do not understood why those internal (variable) names need to be renamed
> 
> for me it would've been equivalent, yes, but i can see wanting to be sure all references are updated correctly.

Indeed, keeping variable names synced with the corresponding references was easy enough.

PR #1441, opened 2024-04-03
Bump galaxy packages to allow for 24.0

bernt-matthias:


tests/test_lint.py
```diff
@@ -39,42 +39,42 @@ def test_fail_tools(self):
 
     def test_lint_default(self):
         with self._isolate_repo("single_tool"):
-            self._check_exit_code(["lint", "--skip", "citations"])
+            self._check_exit_code(["lint", "--skip", "CitationsMissing"])
```
mvdbeek, line 42: You can define an alias, or make the class lowercase and ignore whatever linting stops us. I really don't think we should break the interface here.

bernt-matthias: How about changing [this test](https://github.com/galaxyproject/galaxy/blob/33f77b4e24e18b2f95cd50c6d7ac11c2da3885d0/lib/galaxy/tool_util/lint.py#L218) to something like 

```python
for skip_type in skip_types:
    if name.lower().startswith(skip_type):
        return 
```

Fortunately I named most of the new linter classes such that the old linter function name is a prefix of the new class names. Unfortunately only for most ... which mean we would need to rename some classes. To me it seems that this likely breaks anyway at some  point in the future.

Alternatively we could add a check for the module name with the skip types. If I remember correctly each module contained one function (except inputs which had two :( )

Any preferences?

mvdbeek: > To me it seems that this likely breaks anyway at some point in the future.

why ? we can't break the users experience so often

> Alternatively we could add a check for the module name with the skip types. If I remember correctly each module contained one function (except inputs which had two :( )

I don't know what you're suggesting here. Is it not possible to define an alias or rename the classes ?

bernt-matthias: > why ? we can't break the users experience so often

I just wanted to say that I imagined that the tests that would be necessary to make sure  that the aliases keep working would be quite extensive.

I'm completely on your side that we should maintain the CLI. 

> I don't know what you're suggesting here. Is it not possible to define an alias or rename the classes ?

We certainly can rename the classes. But we might want to ensure with a test that nobody renames them in the future.

Regarding my idea with the modules:

Previously each of the linter modules (e.g. [outputs.py](https://github.com/galaxyproject/galaxy/blob/release_23.2/lib/galaxy/tool_util/linters/outputs.py)) contained one function `lint_xyz(...)` (except for the inputs module which contained two). Functions from the linter modules are [autodetected](https://github.com/galaxyproject/galaxy/blob/ecef3ef5a66d21cf6c379f586c06bf7de78f48c4/lib/galaxy/tool_util/lint.py#L326) and are called via the `LintContext.lint()`  method which does the skip check by equality with the function name (minus the `lint_` prefix).

With the change introduced in 24.0 we replaced each of the functions by a number of classes. Hence the old names used in the skip check do not work any more.

One property that we could exploit to maintain the CLI is that the new classes are contained in the same modules as the old linter functions and the module names are identical to the function names, i.e. we could just check for the module name in addition to the class name. For testing that the mapping of module names to linter Classes is remained we could add a test for each module that checks the number of contained classes (and require that it does not get smaller) plus a comment explaining why.

bernt-matthias: Maybe like so https://github.com/galaxyproject/galaxy/pull/18091

requirements.txt
```diff
@@ -3,8 +3,8 @@ bioblend>=1.0.0
 click!=8.0.2
 cwltool>=1.0.20191225192155
 ephemeris>=0.10.3
-galaxy-tool-util>=23.1,<24.0
-galaxy-util[template]>=23.1,<24.0
+galaxy-tool-util>=23.1,<24.1
```
mvdbeek, line 6:
```suggestion
git+https://github.com/galaxyproject/galaxy.git@release_24.0#egg=galaxy-tool-util&subdirectory=packages/tool_util
```

mvdbeek: If you want to test out the PR before we merge and release packages

requirements.txt
```diff
@@ -3,8 +3,8 @@ bioblend>=1.0.0
 click!=8.0.2
 cwltool>=1.0.20191225192155
 ephemeris>=0.10.3
-galaxy-tool-util>=23.1,<24.0
-galaxy-util[template]>=23.1,<24.0
+galaxy-tool-util @ git+https://github.com/bernt-matthias/galaxy.git@skip-name-restore#egg=galaxy-tool-util&subdirectory=packages/tool_util
```
mvdbeek, line 6:
```suggestion
galaxy-tool-util>=23.1,!=24.0.0,!=24.0.1,<24.1
```

mvdbeek (approved):

PR #1444, opened 2024-04-09
Add text to autopygen exceptions

bernt-matthias:
Otherwise [this](https://github.com/galaxyproject/planemo/blob/28b2ca40ceaa546b0b23678220e05bf155829ea7/planemo/autopygen/argument_parser_conversion.py#L103) error is just empty.


planemo/autopygen/source_file_parsing/parser_discovery_and_init.py
```diff
@@ -165,7 +165,7 @@ def _replace_parser(self, node: ast.Assign, imported_using_from: bool):
 
     def report_findings(self) -> Tuple:
         if self.main_parser_name is None:
-            raise ArgParserNotUsed
+            raise ArgParserNotUsed("Atgument parser not used")
```
mvdbeek, line 168:
```suggestion
            raise ArgParserNotUsed("Argument parser not used")
```

mvdbeek (approved):

Issue #1376, opened 2023-06-28
planemo autoupdate only considers main public toolshed when updating workflows


jancrichter:
Hi there,

we would like to use `planemo autoupdate` to upgrade our workflows, as they contain lots of subworkflows and planemo already offers recursive update functionality. 

Unfortunately, the actual update is only performed if a tool from the main public toolshed is out of date in the workflow. If only tools from private tool sheds are out-of-date the update is skipped. 

If such a tool is present, however, the update is performed correctly for ALL tools in the workflow.

@mvdbeek pointed out in https://github.com/galaxyproject/bioblend/issues/475  that this may well be due to hard-coding of the main public tool shed in planemo. I was able to confirm that tools from our private TS are correctly considered when we change the line below to reference our private tool shed.

https://github.com/galaxyproject/planemo/blob/6a56d97b7622609bdca5b5c77dd933e22d596145/planemo/workflow_lint.py#L52

As @mvdbeek points out it would probably be a good solution to infer the correct tool shed address from the tool ID.

jancrichter:
This is much appreciated!

PR #1445, opened 2024-04-10
Implement run/test/workflow_edit/autoupdate from non default tool sheds

mvdbeek:
Fixes https://github.com/galaxyproject/planemo/issues/1376


jmchilton (approved):
I would have closed with wontfix -  you're very kind.

PR #1451, opened 2024-05-28
add assertion that autoupdate does not modify input description

bernt-matthias:


jmchilton:
Not sure where this was going but the test passes so I'm going to merge it. Thanks for the assertions!

jmchilton (approved):

PR #1452, opened 2024-06-02
Autoupdate: start a separate instance for each workflow

bernt-matthias:
An attempt to solve the problem that the workflow autoupdate does not update all workflows in a dir. 
I think the problem occurs if the tool set of the workflows overlaps (e.g. https://github.com/galaxyproject/iwc/pull/422/). 

The symptom that is fixed by my change is that [`config.updated_repos`](https://github.com/galaxyproject/planemo/blob/cb2321b4a2c84c992b57c27822838a55f3bd9c28/planemo/commands/cmd_autoupdate.py#L141C24-L141C44
) gets empty for some workflows which are therefore skipped. 

I guess a cleaner way to fix this might be to uninstall all tools again after a workflow was processed?

### DIFF
--- a/planemo/commands/cmd_autoupdate.py
+++ b/planemo/commands/cmd_autoupdate.py
@@ -135,9 +135,9 @@ def cli(ctx, paths, **kwds):  # noqa C901
             kwds["install_repository_dependencies"] = False
             kwds["shed_install"] = True
 
-        with engine_context(ctx, **kwds) as galaxy_engine:
-            with galaxy_engine.ensure_runnables_served(modified_workflows) as config:
-                for workflow in modified_workflows:
+        for workflow in modified_workflows:
+            with engine_context(ctx, **kwds) as galaxy_engine:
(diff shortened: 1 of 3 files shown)


mvdbeek:
I would really appreciate it if we didn't need one instance per workflow.
To make it simpler, we could update all workflows by dropping `if config.updated_repos.get(workflow.path) or kwds.get("engine") == "external_galaxy":`.

We then check if any tool has actually been upgraded. That means it will also work for local non-TS tools (e.g. built-in tools and converters). The cost of running the refactoring action is minimal, compared to one instance per workflow.
This way we also get syntax upgrades or fixes "for free" if we decide to do that in the future

bernt-matthias:
> We then check if any tool has actually been upgraded. 

Would we need to implement this? Or is this already included?

mvdbeek (approved):
Thanks a lot @bernt-matthias, added a test for this. Let's merge if it passes

PR #1453, opened 2024-06-06
make sure that skip list does not get `['']` (if no skip is given)

bernt-matthias:
the empty default string in `ctx.global_config` was split to `['']`


nsoranzo (approved):

bernt-matthias:
We would need a new release of the Galaxy python packages to fix the typo.

mvdbeek:
Thank you @bernt-matthias !

PR #1454, opened 2024-06-13
Remove myself from committer list

simonbray:
I've also removed myself from the BioBlend committer team, but I don't believe there is any page to update there.

The page is not in sync with https://github.com/orgs/galaxyproject/teams/planemo, just as an observation :)


nsoranzo:
> I've also removed myself from the BioBlend committer team, but I don't believe there is any page to update there.

Noted.
Thanks for your very valuable contributions to both projects!
Hope you are doing well, you'll always be welcome in the Galaxy community!

PR #1458, opened 2024-06-26
Workflow best practices: check that creator identifiers are URIs

elichad:
From https://github.com/galaxyproject/iwc/blob/main/workflows/README.md#ensure-workflows-follow-best-practices :
> The "Identifier" field of a Creator added as a person should be filled with a fully qualified URI, e.g., https://orcid.org/0000-0002-1825-0097.

This PR adds a check which will emit a warning if a workflow creator is a person and their identifier is not a URI.

I'm aware this function is meant to mirror the Galaxy UI - I can try and make a matching PR for the UI, though I'm weaker on Vue/JS than Python. Thought I would open this PR first to confirm if that's worthwhile.

Context for the PR: my team is scraping Galaxy workflow RO-Crates from WorkflowHub and found a few where the generated author metadata uses just the ORCID number for the `@id` rather than the full ORCID URI. We are keen to help reduce those cases through improved linting!

### DIFF
--- a/planemo/workflow_lint.py
+++ b/planemo/workflow_lint.py
@@ -211,11 +211,13 @@ def _lint_best_practices(path: str, lint_context: WorkflowLintContext) -> None: 
         lint_context.warn("Workflow is not annotated.")
 
     # creator
-    if not len(workflow_dict.get("creator", [])) > 0:
+    creators = workflow_dict.get("creator", [])
+    if not len(creators) > 0:
         lint_context.warn("Workflow does not specify a creator.")
     else:
-        creators = workflow_dict.get("creator")
         if not isinstance(creators, list):
+            # Don't know if this can happen, if we implement schema validation on the Galaxy side
+            # this won't be needed.
             creators = [creators]
         for creator in creators:
             if creator.get("class", "").lower() == "person" and "identifier" in creator:
--- a/planemo/workflow_lint.py
+++ b/planemo/workflow_lint.py
@@ -14,6 +14,7 @@ from typing import (
     TYPE_CHECKING,
     Union,
 )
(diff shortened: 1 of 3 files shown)


planemo/workflow_lint.py
```diff
@@ -212,6 +213,18 @@ def check_json_for_untyped_params(j):
     # creator
     if not len(workflow_dict.get("creator", [])) > 0:
         lint_context.warn("Workflow does not specify a creator.")
+    else:
+        creators = workflow_dict.get("creator")
```
mvdbeek, line 217: Technically the the creators variable could be `None` if there is no "creator" in your workflow, in which case you might run into an error on line 221, calling `.get` on a none-type variable.

```suggestion
        creators = workflow_dict.get("creator", [])
```

mvdbeek: Actually it might be a good idea to do `creators = workflow_dict.get("creator", [])` on line 213 and refer to that.

planemo/workflow_lint.py
```diff
@@ -212,6 +213,18 @@ def check_json_for_untyped_params(j):
...
+        if not isinstance(creators, list):
+            creators = [creators]
```
mvdbeek, line 219: Have you run into any workflows where creator is not a list ?

mvdbeek (approved):
I've taken the liberty to push some a minor change. Nice to see another contribution from you!

elichad:
Thanks @mvdbeek, fine with me! Sorry for not getting back to your comments sooner.

I haven't seen any workflows where the creator is not a list - I just wasn't sure if this could theoretically happen and pass validation elsewhere.

PR #1463, opened 2024-06-29
Increase worflow linting

lldelisle:
I am not sure it is the good way to do it.
For the moment, I cannot make it like for the tools with classes, in order to make it customizable but I think I put everything I thought was missing in the current workflow_lint.
I used sometimes `'` and sometimes `"`. Is this an issue?

PR #1465, opened 2024-07-01
Check lint skip types only for tool linting

bernt-matthias:
the `Linter.list_linters` function only lists tool linters.

### DIFF
--- a/planemo/commands/cmd_lint.py
+++ b/planemo/commands/cmd_lint.py
@@ -4,10 +4,13 @@ import click
 
 from planemo import options
 from planemo.cli import command_function
+from planemo.io import error
 from planemo.tool_lint import (
     build_tool_lint_args,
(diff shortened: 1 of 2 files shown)


lldelisle:
I am sorry, I've just merged the master branch in my branch with the workflow_lint and I don't get this message anymore. Maybe this is not needed.

mvdbeek:
> the `Linter.list_linters` function only lists tool linters.

We should be able to define linters for workflows as well. https://github.com/galaxyproject/planemo/pull/1463/commits/6e8f0ce67e278c6a0fff46f871bbfd9273cd6697 seems like the right fix ?

mvdbeek:
I think https://github.com/galaxyproject/planemo/pull/1463 looked better to me. We can revisit this later if needed.

PR #1470, opened 2024-07-16
Pin galaxy-tool-util to >=24.0.2

nsoranzo:
Follow-up on https://github.com/galaxyproject/planemo/pull/1420 which should have included these version bumps.


nsoranzo:
Superseded by https://github.com/galaxyproject/planemo/pull/1471 .

PR #1471, opened 2024-07-17
add extended-assertions requirements

bernt-matthias:
otherwise the new assertions that we can specify for images won't work

xref https://github.com/BMCV/galaxy-image-analysis/pull/126

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,8 +3,8 @@ bioblend>=1.0.0
 click!=8.0.2
 cwltool>=1.0.20191225192155
 ephemeris>=0.10.3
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -3,7 +3,7 @@ bioblend>=1.0.0
 click!=8.0.2
 cwltool>=1.0.20191225192155
 ephemeris>=0.10.3
-galaxy-tool-util[edam]>=23.1,!=24.0.0,!=24.0.1,<24.2
+galaxy-tool-util[edam,extended-assertions]>=23.1,!=24.0.0,!=24.0.1,<24.2
```
mvdbeek, line 6: You probably need to specify a minimum version where `extended-assertions` was added ?

nsoranzo (approved):

PR #1472, opened 2024-07-27
Rewrite tool linters defined in planemo

bernt-matthias:
Any suggestions how to rewrite the linters for the repos. Wondering how to implement this best, i.e. the problem is that the lint function needs a realized repos instead of a tool source.


jmchilton:
Closing in lieu of https://github.com/galaxyproject/planemo/pull/1706 - I hope that is okay!

PR #1477, opened 2024-09-26
Don't assume default master key works

### LINKED ISSUE #1476: planemo's configuration of bioblend's api key prevents running commands without authentication that don't require it
e.g.

```
bioblend DEBUG 2024-09-26 12:36:21,127: GET - attempts left: 10; retry delay: 10.0
urllib3.connectionpool DEBUG 2024-09-26 12:36:21,129: Starting new HTTPS connection (1): usegalaxy.org:443
urllib3.connectionpool DEBUG 2024-09-26 12:36:21,774: https://usegalaxy.org:443 "GET /api/invocations/84e15596bd4fc608 HTTP/11" 401 62
bioblend WARNING 2024-09-26 12:36:21,774: GET: error 401: b'{"err_msg":"Provided API key is not valid.","err_code":401001}', 9 attempts left
```

that api invocation is public (thanks for the tip @mvdbeek) but planemo's invocation fails due to the API key.

The corresponding bioblend code without an API key however succeeds:

```python
from bioblend.galaxy import GalaxyInstance
gi = GalaxyInstance('https://usegalaxy.org')
print(gi.invocations.show_invocation(invocation_id='84e15596bd4fc608'))
```

If `key=None` or `key=""` is passed to `GalaxyInstance` the call continues to succeed, so however planemo is setting the empty API causes an issue.

mvdbeek:
Should fix https://github.com/galaxyproject/planemo/issues/1476? We might want to see which commands can (at least in principle) work fine without auth.

### DIFF
--- a/planemo/galaxy/api.py
+++ b/planemo/galaxy/api.py
@@ -14,8 +14,6 @@ def get_dict_from_workflow(gi: GalaxyInstance, workflow_id: str, instance: bool 
 
 def gi(port: Optional[int] = None, url: Optional[str] = None, key: Optional[str] = None) -> GalaxyInstance:
     """Return a bioblend ``GalaxyInstance`` for Galaxy on this port."""
-    if key is None:
-        key = DEFAULT_ADMIN_API_KEY
     if port is not None and url is not None:
         raise ValueError("Either port or url parameter needs to be None")
     if port is None:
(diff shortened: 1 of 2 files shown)


nsoranzo:
(Some of the) test failures seem related.

jmchilton:
Closing in favor #1700 - hope that is okay.

Issue #1480, opened 2024-10-28
New toolshed categories are not updating for the existing tools


pavanvidem:
We recently added a few new toolshed categories to the existing tools in this PR: https://github.com/galaxyproject/tools-iuc/pull/6114 

The categories should have been updated in the toolsshed for at least the updated tools. But this is not the case. For eg., Scanpy has been updated recently but the new categories are not updated in the toolshed. 

If a new tool is added, then the categories are updated. For example, the new Seurat tool suite belongs to the Single Cell category on toolshed.

mvdbeek:
That's not how it works, otherwise any user can create new (nonsense) categories.

mvdbeek:
The categories exist, so I'm not sure what this is about.

bgruening:
If Pavan adds a new category to the shed.yml (a category that exists in the TS already), its not populated upon push. I guess we need to understand why changes to a shed.yml is not reflected in the TS.

mvdbeek:
What repo is this ?

pavanvidem:
the tools are from IUC

mvdbeek:
Which tool repo is an example that should have been updated ?

mvdbeek:
Please be more specific. The PR just lists changes to the .shed.yml file

pavanvidem:
All the tools in that PR. For example Scanpy https://github.com/galaxyproject/tools-iuc/tree/main/tools/scanpy

mvdbeek:
I need a deploy commit, the last commit in there has not been deployed afaict

mvdbeek:
Or if you're not finding one let me know about a repo I can deploy.

mvdbeek:
https://github.com/galaxyproject/planemo/pull/1481 should fix this. I don't think it ever worked, so nice one!

pavanvidem:
@mvdbee thanks a lot!  I can see the Scanpy tools in the Single Cell category on toolshed.
 
This will trigger upon a tool update or? What about the other tools in that PR where I recently added the Single Cell category? Should I just bump the other tools' galaxy versions to trigger? is there a better trick?

mvdbeek:
Someone with iuc permissions can set the category in the TS user interface or reupload the package, which will also update the metadata.

PR #1481, opened 2024-10-28
Fix repository category update

mvdbeek:
`category_ids[]` is not a valid agument for the update endpoint (it is for create). To fix this I just call into bioblend instead of using the custom put request. Fixes https://github.com/galaxyproject/planemo/issues/1480


bgruening:
Thanks @mvdbeek!

nsoranzo (approved):

nsoranzo:
The shellescape errors will be fixed by galaxy-tool-util 24.1.3, which is currently only on test PyPI.

mvdbeek:
I've promoted the pre-release to release and the packages are on pypi now

PR #1483, opened 2024-12-12
Add Geo Science

bernt-matthias:
Working on [OGS](https://gitlab.opengeosys.org/ogs/ogs) and [OGS tools](https://gitlab.opengeosys.org/ogs/tools/ogstools) (and maybe [mhm](https://git.ufz.de/mhm/mhm/) in the future). We would need a new category for this. Authors of OGS suggested `Geo Science` or `Earth Science` (with a preference for the former).

### DIFF
--- a/planemo/shed/__init__.py
+++ b/planemo/shed/__init__.py
@@ -125,6 +125,7 @@ CURRENT_CATEGORIES = [
     "Genome editing",
     "Genome-Wide Association Study",
     "Genomic Interval Operations",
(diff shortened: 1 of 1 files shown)


bgruening:
ping @Marie59

Marie59:
I confirm from my point of view I think Geo Science is better. I am afraid an Earths science section would confuse people with the earth subdomain (especially that I'd like also to create a section Ocean science)

PR #1487, opened 2025-01-17
Remove usage of `defaults` conda channel

bernt-matthias:


bgruening:
+1, thanks @bernt-matthias!

planemo/options.py
```diff
@@ -2061,7 +2061,7 @@ def mulled_conda_option():
         default=None,
         help=(
             "Install a specific version of Conda before running the command, by "
-            "default the version that comes with the continuumio miniconda3 image "
+            "default the version that comes with the continuumio miniforge image "
```
nsoranzo, line 2064:
```suggestion
            "default the version that comes with the miniforge image "
```

nsoranzo (approved):

bernt-matthias:
Lets see if we can fix the failing `deploy` job https://github.com/galaxyproject/planemo/pull/1488

nsoranzo:
The conda/pyopenssl test failures should be fixed by https://github.com/galaxyproject/galaxy/pull/19382 , but probably we need a new release of the tool-util package?

bernt-matthias:
The `conda_build` error should be fixed in https://github.com/conda/conda-build/pull/5627

PR #1490, opened 2025-01-27
pin default python version to the max currently allowed by galaxy pinnings

### LINKED ISSUE #1489: Default python version to loosely defined
[This](https://github.com/galaxyproject/planemo/blob/8ef50e454a44c7d5a079464f2e410d5076fd1a68/planemo/virtualenv.py#L10) apparently leads to install python 3.13 which does not work in `planemo test`.

bernt-matthias:
fixes https://github.com/galaxyproject/planemo/issues/1489

### DIFF
--- a/planemo/options.py
+++ b/planemo/options.py
@@ -177,7 +177,7 @@ def galaxy_python_version():
         "--galaxy_python_version",
         use_global_config=True,
         default=None,
-        type=click.Choice(["3.8", "3.9", "3.10", "3.11", "3.12"]),
+        type=click.Choice(["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]),
         help="Python version to start Galaxy under",
     )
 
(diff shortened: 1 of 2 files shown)


planemo/options.py
```diff
@@ -177,7 +177,7 @@ def galaxy_python_version():
         "--galaxy_python_version",
         use_global_config=True,
         default=None,
-        type=click.Choice(["3", "3.7", "3.8", "3.9", "3.10", "3.11"]),
+        type=click.Choice(["3.8", "3.9", "3.10", "3.11", "3.12"]),
```
nsoranzo, line 180: 3.13 can be used with the upcoming Galaxy release 24.2 .

bernt-matthias:
```suggestion
        type=click.Choice(["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]),
```

jmchilton:
Superseded by #1704, which rebases this work onto current master, preserves @bernt-matthias' original commit authorship, updates the explicit pin for Galaxy's current Python 3.14 support, and adds focused regression coverage. Thank you for the original fix!

PR #1492, opened 2025-02-21
Add postgres singularity option

Smeds:
Hi!

I’ve been experimenting with running Planemo on the TACC server using SLURM, and I ran into an issue where more complex workflows would fail due to the SQLite database being locked.

Since I don’t have the ability to launch a Docker container for database access on TACC, nor access to an external database, I developed a solution: running a PostgreSQL database within a Singularity container. However, I wanted to avoid manually starting the Singularity container each time. This pull request introduces functionality that allows Planemo to automatically launch a Singularity instance for the PostgreSQL database and stop it once the workflow completes (or fails).

While I’m unsure if this feature will be useful to others, it was extremely helpful during my testing with Planemo.

### DIFF
--- a/planemo/database/factory.py
+++ b/planemo/database/factory.py
@@ -5,6 +5,7 @@ from galaxy.util.commands import which
 from .interface import DatabaseSource
 from .postgres import LocalPostgresDatabaseSource
 from .postgres_docker import DockerPostgresDatabaseSource
+from .postgres_singularity import SingularityPostgresDatabaseSource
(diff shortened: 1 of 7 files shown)


Smeds:
> Can you fix the linting failures ?

absolutely!

Smeds:
@mvdbeek I think I managed to fix the linting errors.

mvdbeek:
Thanks @Smeds!

PR #1496, opened 2025-03-05
Add support for Python 3.12

nsoranzo:


mvdbeek (approved):

nsoranzo:
Unfortunately, it's likely that the py313-unit-quick tests will get stuck when running `tests/test_external_galaxy_commands.py` (i.e `tests/test_engines.py::test_runnable_types` is the last one passing). At least that's what happened on my fork, and I can also reproduce it locally.
I've unsuccessfully tried to debug it for some time, maybe some fresh eyes can help? :pray:

nsoranzo:
> Unfortunately, it's likely that the py313-unit-quick tests will get stuck when running `tests/test_external_galaxy_commands.py` (i.e `tests/test_engines.py::test_runnable_types` is the last one passing). At least that's what happened on my fork, and I can also reproduce it locally. I've unsuccessfully tried to debug it for some time, maybe some fresh eyes can help? 🙏

Note that I reverted to only add support for Python 3.12, where the above doesn't happen.

PR #1500, opened 2025-03-14
Migrate mypy settings to pyproject.toml.

jmchilton:


nsoranzo:
What about a dedicated `mypy.ini` file? Easier to copy/compare between projects, and it keeps the section names simpler.

jmchilton:
Yeah - that makes sense. This was like a test ballon to start shifting things into pyproject.toml but what you're saying makes more sense.

PR #1502, opened 2025-03-18
Add linter to check if tool versions are bumped

bernt-matthias:
so far against hardcoded toolshed.

### DIFF
--- a/planemo/lint.py
+++ b/planemo/lint.py
@@ -43,6 +43,7 @@ def build_lint_args(ctx: "PlanemoCliContext", **kwds) -> Dict[str, Any]:
                 skip_types.append(line)
 
     linters = Linter.list_linters()
+    linters.extend(["version_bumped"])
     invalid_skip_types = list(set(skip_types) - set(linters))
     if len(invalid_skip_types):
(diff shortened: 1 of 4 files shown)


nsoranzo:
@bernt-matthias I suppose we need to disable the new `ShedVersion` check when running the weekly linting and tests in IUC's `.github/workflows/ci.yaml` ?

bernt-matthias:
https://github.com/galaxyproject/tools-iuc/pull/6885

peterjc:
Why is an *equal* version treated as an error? It *should* be equal if the latest code has just been pushed to the Tool Shed.

bernt-matthias:
> Why is an _equal_ version treated as an error? It _should_ be equal if the latest code has just been pushed to the Tool Shed.

CI is executed before the push to the TS. The linter check is intended to check if the tool version has been increased in a pull request.

CI does not do this nicely in some cases, eg if only one tool of a suite is updated. Therfore we allow to push anyway at IUC (after a careful double check).

peterjc:
I guess the usefulness of the version equality check depends on the workflow and when planemo lint is run. I don’t think this should be the default, but can and have disabled this check entirely instead (not ideal as it won’t catch real errors out of sync). Thanks for clarifying

PR #1505, opened 2025-03-26
Print stdout stderr if install command fails

mvdbeek:
This is more helpful than `Failed to install Galaxy via command [{command}]`.


mvdbeek:
Conveniently we can see the stderr now in the failing (but unrelated) test 😄

jmchilton (approved):

PR #1506, opened 2025-03-26
Overhaul Galaxy job config handling.

jmchilton:
- More options from the command-line - multiple runners, TPV shared database, singularity, etc...
- Switch from job conf via XML to job conf via YAML.
- Bring in dependency on galaxy-job-config-init to handle a lot of that logic.
- Commands to assign binding a job config to a created Planemo profile.

### DIFF
--- a/planemo/commands/cmd_config_init.py
+++ b/planemo/commands/cmd_config_init.py
@@ -12,6 +12,7 @@ from planemo import (
 from planemo.cli import command_function
 from planemo.io import (
     info,
+    launch_if_open_flagged,
     warn,
 )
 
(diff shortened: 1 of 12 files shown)


Smeds:
> Could we also default to `tmp_dir: true` on the destination (i.e tmpdir in work dir) ? I think that's the easiest way to get something that will most likely work ?

you want to resulting job_config to look like this by default?

```yaml
...

execution:
  ...
  environments:
    local:
      runner: local
      tmp_dir: true
...
```

mvdbeek:
yes

Smeds:
> yes

I have made a pull-request against galaxy-job-config-init that will add tmp_dir option ([#1](https://github.com/galaxyproject/galaxy-job-config-init/pull/1))

Smeds:
It would be great if someone could upload a new release of [galaxy-job-config-init](https://github.com/galaxyproject/galaxy-job-config-init) to PyPI. After that, we could look into merging, which would enable me to start updating my instructions for running Planemo with a SLURM cluster

Smeds:
@mvdbeek could we trigger the test again? I guess we should be able to merge this pull-request now when @jmchilton has published a new version of galaxy-job-config-init?

PR #1507, opened 2025-03-27
Fix dependency specification

mvdbeek:
`pip install planemo==0.75.28` doesn't install any dependencies at all ...


nsoranzo:
Does it mean we can/should drop `setup.py` ?

mvdbeek:
Either that or maybe we just need to drop the empty dependencies array ? I'll drop setup.py

jmchilton:
I'm so sorry about that - I even encountered this in testing #1506 and didn't realize it was symptomatic of a deeper problem.

nsoranzo (approved):

mvdbeek:
No worries and no harm done. I yanked it from pypi just so it doesn't become the go to package when there is a dependency conflict with newer releases.

PR #1510, opened 2025-03-27
Workflow progress bar.

jmchilton:
Talked about for Galaxy Live! A progress bar would be more useful than the Galaxy logs for running workflows.

This PR is just in early stages but I wanted to have a PR to track progress and get feedback. My approach is develop the progress bar in the abstract with synthetic API responses, create a Planemo command "workflow_track" that can just be attached to existing invocations like "planemo workflow_test_on_invocation" at runtime, and then once it seems to work replace the wait function ``_wait_for_invocation_jobs`` in planemo.galaxy.activity with the progress bar. "planemo workflow_track" could be used to test how workflow errors render also. 

The implementation I have so far is using Rich (https://rich.readthedocs.io/en/latest/style.html). It seems a bit more full featured than click's builtin component and less opinionated than tqdm.

### DIFF
--- a/planemo/commands/cmd_workflow_track.py
+++ b/planemo/commands/cmd_workflow_track.py
@@ -10,6 +10,7 @@ from planemo.galaxy.activity import wait_for_invocation_and_jobs
 
 @click.command("workflow_track")
 @options.invocation_target_options()
+@options.fail_fast_option()
 @command_function
 def cli(ctx, invocation_id, **kwds):
     """Follow the progress of a workflow invocation."""
@@ -21,6 +22,7 @@ def cli(ctx, invocation_id, **kwds):
             history_id=None,
             user_gi=user_gi,
             polling_backoff=5,
+            fail_fast=kwds.get("fail_fast", False),
         )
 
     ctx.exit(0)
--- a/planemo/commands/cmd_workflow_track.py
+++ b/planemo/commands/cmd_workflow_track.py
@@ -1,0 +1,50 @@
+"""Module describing the planemo ``workflow_track`` command."""
+
+import click
+
+from planemo import options
+from planemo.cli import command_function
+from planemo.engine.factory import engine_context
+from planemo.galaxy.workflow_progress import WorkflowProgress
+
+
+@click.command("workflow_track")
+@options.invocation_target_options()
+@command_function
+def cli(ctx, invocation_id, **kwds):
+    """Run defined tests against existing workflow invocation."""
+    with WorkflowProgress() as workflow_progress:
+        workflow_progress.add_bars()
+        import time
+
+        time.sleep(1)
+        new_step = {"state": "new"}
+        scheduled_step = {"state": "scheduled"}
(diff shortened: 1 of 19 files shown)


jmchilton:
The code was nice and clean before I considered subworkflows.... now it is not so much - really getting into the weeds. I think I need to simulate multiple subworkflows evolving over time to keep testing this 😅. How to track what we consider a failure is already complex and needs to be a lot more complex - calling the whole thing a failure on a failed job is bad but I'm not sure how to fix it (first crack could be to disable that behavior if there is a single filter failed step anywhere in the workflow - still so wrong though). The display stuff is relatively clean compared to all that I think. The simple recursion the previous tracking did is also not ideal for this display because I think we should bounce around multiple subworkflows and show one at a time for a few seconds and the display of the parent workflow is slightly different than subworkflows.

Here is a single workflow and one subworkflow - not from the tracking code just the display component in isolation with minimal synthetic API requests.

https://github.com/user-attachments/assets/2201b902-cb96-4e6e-ad8c-f73b16c9ed62

jmchilton:
I have rotating multiple subworkflow tracking and tracking subworkflows along with the primary workflow in "parallel" working. I had to implement a cool little invocation-evolution-over-time DSL - might be cool to use when implementing processing failure messages and such. I parameterized the display a bunch too so it should be easy to bike shed.



https://github.com/user-attachments/assets/b984ccea-64d0-4a25-92a6-2931f3bc7174

jmchilton:
Also... in dark mode and more traditional font size.

https://github.com/user-attachments/assets/34e29b01-f865-435e-bb8a-24e5e082cf94

mvdbeek:
Wow, this is amazing, I think that'll make a huge difference in how usable the CLI is perceived to be.

Smeds:
Looks really nice! Tried it out today. One thing I thought of was that it would be nice if the copying of input files also are visualized, that part can take a lot of time.

Would be nice if the log output to the terminal is turned of and that we could turn it on when needed, logs can still be found at `gravity/log`.

planemo/commands/cmd_workflow_track.py
```diff
@@ -0,0 +1,26 @@
...
+from planemo.cli import command_function
+from planemo.engine.factory import engine_context
+from planemo.galaxy.activity import wait_for_invocation_and_jobs
+
+
+@click.command("workflow_track")
+@options.invocation_target_options()
+@command_function
+def cli(ctx, invocation_id, **kwds):
+    """Run defined tests against existing workflow invocation."""
```
bgruening, line 15: I assume this is a copy and paste artifact?

mvdbeek: I think so, fixed this now.

mvdbeek:
rich also lets us create clickable links (https://github.com/galaxyproject/planemo/pull/1510/commits/b156755161656e40607f67b6b58975e7e8fd6794):
 
Screenshot 2025-05-22 at 11 28 00

PR #1520, opened 2025-04-25
This allows the user, during a run, to export it as a compressed arch…

Smeds:
…ive and/or upload it to a Galaxy instance.

### DIFF
--- a/planemo/galaxy/activity.py
+++ b/planemo/galaxy/activity.py
@@ -263,7 +263,7 @@ def _execute(  # noqa C901
                     outf=archive
                 )
 
-            if kwds.get("arhicve_file"):
+            if kwds.get("archive_file"):
                 shutil.copy(archive_file_output, archive_file)
                 ctx.vlog(f"Archive {kwds.get('arhicve_file')} created.")
 
--- a/planemo/galaxy/activity.py
+++ b/planemo/galaxy/activity.py
@@ -1,6 +1,8 @@
 """Module provides generic interface to running Galaxy tools and workflows."""
 
 import os
+import requests
+import shutil
 import sys
 import tempfile
 import time
@@ -42,7 +44,11 @@ from requests.exceptions import (
 )
 
 from planemo.galaxy.api import summarize_history
-from planemo.io import wait_on
+from planemo.io import (
+    error,
+    wait_on,
+    warn,
+)
 from planemo.runnable import (
     ErrorRunResponse,
     get_outputs,
@@ -236,6 +242,47 @@ def _execute(  # noqa C901
         ctx.vlog("collecting outputs from run...")
         run_response.collect_outputs(output_directory)
         ctx.vlog("collecting outputs complete")
+
+    if kwds.get("upload_instance_url") or kwds.get("archive_file"):
+        ctx.vlog(f"Preparing galaxy run export, history {history_id}.")
+        archive_file = kwds.get("archive_file")
+
+        jeha_id = user_gi.histories.export_history(
+            history_id=history_id,
+            wait=True,
+            maxwait=3600,
+        )
+
+        with tempfile.TemporaryDirectory() as temp_dir:
+            archive_file_output = os.path.join(temp_dir, "archive.tar.gz")
+
+            with open(archive_file_output, 'bw') as archive:
+                user_gi.histories.download_history(
+                    history_id=history_id,
+                    jeha_id=jeha_id,
+                    outf=archive
+                )
+
+            if kwds.get("arhicve_file"):
+                shutil.copy(archive_file_output, archive_file)
+                ctx.vlog(f"Archive {kwds.get('arhicve_file')} created.")
+
+            if kwds.get("upload_instance_url"):
+                upload_url = kwds.get("upload_instance_url", None)
+                upload_key = kwds.get("upload_api_key", None)
+
+                upload_url = f"{upload_url}/api/histories"
+                if upload_key is None:
+                    warn("No API key provided")
+                else:
+                    upload_url = f"{upload_url}?key={upload_key}"
+
+                response = requests.post(upload_url, files={'archive_file': open(archive_file_output, 'rb')})
+                if response.status_code == 200:
+                    ctx.vlog(f"Upload run to {kwds.get('upload_instance_url')}")
+                else:
+                    error(f"Failed to upload run to {kwds.get('upload_instance_url')}, status code: {response.status_code}, error {response.text}")
+
     return run_response
 
 
--- a/planemo/commands/cmd_run.py
+++ b/planemo/commands/cmd_run.py
@@ -28,6 +28,9 @@ from planemo.test.results import StructuredData
 @options.run_output_directory_option()
 @options.run_output_json_option()
 @options.run_download_outputs_option()
+@options.run_upload_instance_url_option()
+@options.run_upload_api_key_option()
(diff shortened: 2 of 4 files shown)


planemo/galaxy/activity.py
```diff
@@ -236,6 +242,47 @@ def _execute(  # noqa C901
...
+            archive_file_output = os.path.join(temp_dir, "archive.tar.gz")
+
+            with open(archive_file_output, 'bw') as archive:
+                user_gi.histories.download_history(
+                    history_id=history_id,
+                    jeha_id=jeha_id,
+                    outf=archive
+                )
+
+            if kwds.get("arhicve_file"):
```
mvdbeek, line 266: There's a typo here. Can you add tests ?

planemo/galaxy/activity.py
```diff
@@ -236,6 +242,47 @@ def _execute(  # noqa C901
         ctx.vlog("collecting outputs from run...")
         run_response.collect_outputs(output_directory)
         ctx.vlog("collecting outputs complete")
+
+    if kwds.get("upload_instance_url") or kwds.get("archive_file"):
+        ctx.vlog(f"Preparing galaxy run export, history {history_id}.")
+        archive_file = kwds.get("archive_file")
+
+        jeha_id = user_gi.histories.export_history(
```
mvdbeek, line 250: We'll want to export the invocation here.

planemo/galaxy/activity.py
Smeds: Yes, that's a type.

I have added a test for the download option.

I'm considering of writing a test that involves uploading a history to a separate Galaxy instance, and I'd appreciate some guidance on how to approach this. Would it make sense to look at the existing serve tests and try spinning up a Galaxy instance manually? Or would it be better to use the instance that Planemo starts during testing?

Smeds: Looks like I can import an invocation exported as a RO-Crate, but not when I have exported it as "Compressed File: Export the invocation to a compressed File containing the invocation data in Galaxy native format."

nsoranzo: @Smeds If you'd like to implement invocation export in BioBlend, it should be pretty similar to https://github.com/galaxyproject/bioblend/blob/9bdc2ce57c4443d29699ece7437d2bdf3692d4dc/bioblend/galaxy/invocations/__init__.py#L441 , I think.
Pull requests welcome!

mvdbeek: Interesting, the tar.gz export/import worked for me. Do you still have the archive ?

Smeds: Could be an issue with test.galaxyproject.org that I was using, seem to work on usegalaxy.org.

Is there any preference what kind of export format to use, RO-crate or Compressed Galaxy native format?

mvdbeek: Go with RO-crate imo, it's essentially the native format with a bit of extra metadata

mvdbeek:
The API takes a history_id parameter, see the API docs in https://usegalaxy.org/api/docs#/workflows/create_invocations_from_store_api_invocations_from_store_post and the test method here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy_test/base/populators.py#L2050-L2061

Smeds:
I tried using submit history_id earlier without success. It will be used if I don't provide an archive, copying the history I provided. But when I attach a file to the request it will not use history_id. Looks like  [queue_history_import](https://github.com/galaxyproject/galaxy/blob/dd11f859364b0c3ff3e078c422963c2901bc776c/lib/galaxy/webapps/galaxy/services/histories.py#L266) only uses the "trans: ProvidesHistoryContext" variable. 

https://github.com/galaxyproject/galaxy/blob/dd11f859364b0c3ff3e078c422963c2901bc776c/lib/galaxy/webapps/galaxy/services/histories.py#L233-L296

Smeds:
Thanks for pointing me in the right direction! I'm able to import the invocation but the datasets aren't included, all data have this message `This dataset has been discarded and its files are not available to Galaxy.`. If I import the same file using the UI all files will be included. Is there a setting I can pass along to have the datasets imported? 

Import using [create_invocations_from_store_api_invocations_from_store_post](https://test.galaxyproject.org/u/paer3961/h/planemo-import-2025-05-06-10-38-37)

Import using [Web UI](https://test.galaxyproject.org/u/paer3961/h/manual-import)

Payload to the request example:
```json
{
    "history_id": "1e8ab44153008be8",
    "model_store_format": "rocrate.zip",
    "store_content_uri": "base64://XXXXXXX"
}
```

mvdbeek:
That's controlled at the export:

Screenshot 2025-05-06 at 18 47 43

Smeds:
> That's controlled at the export:
> Screenshot 2025-05-06 at 18 47 43

I had included_files set to True, here the form data I provide to the export function:
```python
{
                "model_store_format": "rocrate.zip",
                "include_files": True,
                "include_deleted": False,
                "include_hidden": False,
}
```

When I import the same file with the web UI the datasets get imported.

Smeds:
I believe I've found the part of the Galaxy code that marks the dataset as discarded when data is imported using the /api/invocations/from_store endpoint.

https://github.com/galaxyproject/galaxy/blob/5a1b0ba517d0d9a6cf32c18f81b0e7cd8cf5c58f/lib/galaxy/managers/model_stores.py#L333

Smeds:
Importing a history using from_store should behave the same way. Is this how we wants it to behave? Import via the Web UI seem to be using `/api/histories' as an endpoint which behaves a bit differently.

jmchilton:
> Posted by Claude (AI assistant) on behalf of the commenter — they did not write this text personally.

Going through older PRs, and I think this one can be closed — mostly because you already solved it.

Your `invocation_export` command (11579f5d, merged 2025-06-03) and mvdbeek's `--export_invocation` flag (62782b4a, merged 2025-06-20) landed in the weeks after you opened this, and between them they cover the archive half. `--export_invocation` even hooks into the same spot in `_execute` that this PR patches:

```
planemo run wf.ga job.yml --export_invocation run.rocrate.zip
planemo invocation_export <invocation_id> --output run.rocrate.zip
```

The difference is that those export the *invocation* as a model store rather than the *history* as a JEHA `.tar.gz`, which for a workflow run is the object that actually carries the provenance. So rebasing this would mean reintroducing the older mechanism alongside the newer one.

The part that isn't covered is `--upload_instance_url` — pushing a run straight to another instance.

Worth saying: **your diagnosis of why that was broken was right, and it's since been fixed.** You traced the discarded datasets to `discarded_data=ImportDiscardedDataType.FORCE` being hardcoded in `create_objects_from_store`. Galaxy now takes it from the request:

```python
import_options = ImportOptions(
    discarded_data=ImportDiscardedDataType(payload.discarded_data),
    ...
)
```

That's in 26.0 and 26.1, so importing via `/api/invocations/from_store` no longer forces everything to discarded. The thing that stalled you is gone.

If cross-instance upload is still useful to you, it'd be worth a fresh issue or PR rather than a rebase of this branch — it's a smaller change now that it can build on `export_invocation_as_archive`. One thing to know going in: bioblend has `get_invocation_archive` for export but no `from_store` equivalent for import, so it needs either a bioblend addition or a raw call on the planemo side.

Closing to keep the queue tidy, not because the idea was wrong — happy to reopen or review a follow-up. Thanks for the original work, and for chasing the Galaxy-side bug far enough to find the actual line.

PR #1526, opened 2025-05-15
create provided output_directory if it doesn't exist

Smeds:
Planemo currently fails with an error if the specified output directory does not exist, instead of creating it automatically.


mvdbeek (approved):
Thanks!

PR #1528, opened 2025-05-16
support Singularity versions that lack the --env flag.

Smeds:
Prior to version 3.6, Singularity did not support the --env flag and instead relied on environment variables prefixed with SINGULARITYENV_ to set container environments. These environment variables are essential during the initialization of the PostgreSQL database. This update ensures compatibility with both newer versions of Singularity that support the --env flag and older versions that do not.

### DIFF
--- a/planemo/database/postgres_singularity.py
+++ b/planemo/database/postgres_singularity.py
@@ -6,7 +6,7 @@ import time
 from tempfile import mkdtemp
 
 from galaxy.util.commands import execute
-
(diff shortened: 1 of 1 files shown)


mvdbeek:
3.6 was released more than 5 years ago, I don't think it is worth the complexity for a new feature here. If you have an old version you can use https://github.com/apptainer/apptainer/blob/main/tools/install-unprivileged.sh

PR #1529, opened 2025-05-16
start implementing support for downloading output_data after an invocation completes

Smeds:
Currently, the --download_outputs flag can be used with planemo run, but it requires the user to wait until the execution is complete. If the user prefers to use --nowait or if the connection is lost during execution, they lose the ability to retrieve the outputs. This pull request begins to introduce support for downloading outputs after the invocation has finished, regardless of when or how it completes.

The current implementation has a major limitation: the user must provide the history_id and workflow_id in addition to the invocation_id. While it’s possible to fetch invocation details using the invocation_id, the response only includes encoded versions of history_id and workflow_id, which cannot be used directly. I'm not sure how to decode these values—any guidance or pointers would be appreciated.


Smeds:
> > This pull-request depends on a new release of bioblend, bioblend>1.5.0
> 
> @Smeds Let me know if you need other PRs for bioblend merged, otherwise I can make a new release.

@nsoranzo Thanks! I would like to be able to have bioblend [#506](https://github.com/galaxyproject/bioblend/pull/506) merged, but that on is dependent on galaxy [#20359](https://github.com/galaxyproject/galaxy/pull/20359). Don't know if we should wait on that on or make a release now for bioblend.

mvdbeek:
I think this is covered by https://github.com/galaxyproject/planemo/pull/1534 and https://github.com/galaxyproject/planemo/pull/1532 ? feel free to re-open if I missed something. Thanks for all the great work @Smeds and the help from @simonbray and @nsoranzo!

PR #1530, opened 2025-06-02
update list invocation commands

Smeds:
this update will make it possible to:
1 - list all invocations without knowing the workflow id, invocations will then be grouped by workflow.. If the user provided a workflow id, only those invocations will be shown. 
2 - output the list as a json when flag `--raw` is used.

Screenshot from 2025-06-02 14-36-04


jmchilton (approved):
Looks amazing - thank you!

jmchilton:
Superseded by #1703, which rebases this work onto current master, preserves @Smeds' original commit authorship, and adds focused pagination and CLI regression coverage. Thank you again for the original implementation!

PR #1534, opened 2025-06-03
cmd to export an existing invocations as an archive for storage or later import.

Smeds:


mvdbeek:
Test is passing fine with https://github.com/galaxyproject/planemo/pull/1541 and bioblend from the master branch

PR #1539, opened 2025-06-18
update shed categories

bgruening:


nsoranzo:
It looks like that categories on testtoolshed and toolshed diverged:

| testtoolshed | toolshed |
| --- | --- |
|  | Geo Science |
| InteractiveTools | Interactive Tools |
| NLP | Natural Language Processing |
|  | Single Cell |
|  | Spatial Omics |

Currently `scripts/categories.py` gets the categories via API from the testtoolshed.
Can we fix the testtoolshed to have the same categories?
@martenson can probably do that?

martenson:
I added the three missing, the following two are inconsistent atm, maybe changing the DB is a way...

| testtoolshed | toolshed |
| --- | --- |
| InteractiveTools | Interactive Tools |
| NLP | Natural Language Processing |

martenson (approved):

PR #1512, opened 2025-04-01
first draft for instructions how to run planemo and submit jobs to a slurm cluster

Smeds:
These instructions have currently been tested on:

- [x] stampede3 
- [ ] PSC Bridges-2
- [ ] Purdue Anvil CPU

by smeds

PR #1543, opened 2025-06-19
Add a slurm_init command.

jmchilton:
It should allow simplify the documentation for using SLURM, allow for auto-testing the documentation, and having a fixed location to expect the slurm library by default could allow us to auto pick it up if it is available.

@Smeds Any chance you'd be willing to test this with your setup and then use it to simplify the docs you're adding #1512?


mvdbeek:
I can test this tomorrow considering I just spent time on getting that to work 😅

mvdbeek:
I'll fix it

PR #1555, opened 2025-06-24
Add uvx_galaxy engine type

mvdbeek:
Skips cloning and run.sh to go directly for the galaxy entrypoint. Is much faster and cleaner to boot, and you can follow along the invocation in the UI if you like.
~Does not yet install conditional dependencies.~


jmchilton:
@mvdbeek do you want to keep this open in light of #1701? Any clues what kind of testing you'd like to see on #1701 to get it in a reviewable/mergable state?

PR #1556, opened 2025-07-09
Fixing 429 url error

arash77:
This pull request includes a minor change to the `validate_url` function in `planemo/lint.py`. The change ensures that the else will not run when the first if has been passed.


mvdbeek:
Thank you!

PR #1559, opened 2025-08-04
Add special handling for DOIs to URL linter

bernt-matthias:
Redirect targets of DOIs often block prgrammatic CI requests. This might help.


bernt-matthias:
Wondering if we want to move this to `tool-util`. We could reuse the doi check for tool `<citations>` and also in workflows, or?

Any suggestions for the `C901` linter error?

PR #1626, opened 2026-03-19
Fix workflow lint crash when tool_state is dict instead of JSON string

jmchilton:
Some .ga workflows have tool_state as a dict rather than a JSON-encoded string. Handle both forms to avoid json.loads TypeError.


mvdbeek (approved):

mvdbeek:
Test error is surely unrelated

PR #1628, opened 2026-03-24
Adjust test_index indexing

RZ9082:
In order to avoid confusion and keep inline with the `planemo test` stdout 1-based indexing, Changed test_index indexing from 0-based to 1-based.

PR #1631, opened 2026-04-08
shift the index to 1-based

nilchia:
When planemo test is finished, the report is in 1-based index. So I think the `--test_index` should also be 1-based.


nilchia:
I just noticed this PR https://github.com/galaxyproject/planemo/pull/1628

I'll close this PR.

PR #1633, opened 2026-04-20
Bring in gxformat2 Linting Improvements (add some agent assistance files)

jmchilton:
Use gxformat2 0.25 lint interface; drop duplicated best-practices
Bump gxformat2>=0.25.0; switch workflow_lint to lint_*_path and
lint_best_practices_{ga,format2}; delete the duplicated planemo
_lint_best_practices body that was migrated upstream. Wire the new
lint_pydantic_validation as a schema_validation lint step. Coerce
gxformat2 Linter classes to their names in WorkflowLintContext so
galaxy's LintMessage doesn't render class reprs.

Update test assertions for the new message text (drops "with ID",
uses native step ids directly). Clean up three legacy test fixtures
that strict pydantic validation now rejects: null input key in
wf14-unlinted-best-practices.yml and format2-style top-level inputs/
outputs on two native basic_native fixtures.

### DIFF
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -63,7 +63,10 @@ from planemo.io import (
     write_file,
 )
 from planemo.mulled import build_involucro_context
-from planemo.runnable import RunnableType
+from planemo.runnable import (
+    GALAXY_TOOLS_PREFIX,
+    RunnableType,
+)
 from planemo.shed import tool_shed_url
 from .api import (
     DEFAULT_ADMIN_API_KEY,
@@ -616,7 +619,12 @@ def _all_tool_paths(
 ) -> Set[str]:
     extra_tools = extra_tools or []
     all_tool_paths = {
-        r.path for r in runnables if r.has_tools and not r.data_manager_conf_path and not r.is_remote_workflow_uri
+        r.path
+        for r in runnables
+        if r.has_tools
+        and not r.data_manager_conf_path
+        and not r.is_remote_workflow_uri
+        and not r.uri.startswith(GALAXY_TOOLS_PREFIX)
     }
     extra_tools = _expand_paths(galaxy_root, extra_tools=extra_tools)
     all_tool_paths.update(extra_tools)
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -940,7 +940,7 @@ class BaseGalaxyConfig(GalaxyInterface):
         # TODO: Allow serialization so this doesn't need to assume a
         # shared filesystem with Galaxy server.
         from_path = default_from_path or (runnable.type.name == "cwl_workflow")
-        workflow = import_workflow(runnable.path, admin_gi=self.gi, user_gi=self.user_gi, from_path=from_path)
+        workflow = import_workflow(runnable.path, user_gi=self.user_gi, from_path=from_path)
         self._workflow_ids[runnable.path] = workflow["id"]
 
     def workflow_id_for_runnable(self, runnable):
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -934,7 +934,7 @@ class BaseGalaxyConfig(GalaxyInterface):
         # TODO: Allow serialization so this doesn't need to assume a
         # shared filesystem with Galaxy server.
         from_path = default_from_path or (runnable.type.name == "cwl_workflow")
-        workflow = import_workflow(runnable.path, admin_gi=self.gi, user_gi=self.user_gi, from_path=from_path)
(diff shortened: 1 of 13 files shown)


.claude/commands/ready-release.md
```diff
@@ -0,0 +1,21 @@
+Walk me through readying a planemo release. Run each step, check for problems, and pause if anything looks wrong.
+
+## Steps
```
mvdbeek: Can we point to https://planemo.readthedocs.io/en/latest/developing.html or merge with ?

CLAUDE.md
```diff
@@ -0,0 +1,51 @@
+# CLAUDE.md
```
mvdbeek: Does claude get any of this wrong ? On the surface that seems like stuff claude does without telling it to, but it consumes tokens on every session.

mvdbeek (approved):
This is awesome, minus the claude.md. If we add anything it should be very short (use makefile when appropriate)  ? xref https://arxiv.org/html/2602.11988v1

CLAUDE.md
jmchilton: No you're right. This is the default init as of last Thursday - I read they recently overhauled init and it was better now. But double checking on that just now apparently it has to be enabled and isn't on by default. https://medium.com/@AdithyaGiridharan/claude-codes-init-command-is-getting-a-brain-transplant-e0c389ec0c6b

I will drop this from the PR.

.claude/commands/ready-release.md
jmchilton: This was generated from that (or from gxformat2's version of that since that is where I got the command) but I can link the doc and request they stay in sync. Or would you rather this just say "read the doc, confirm the preconditions, and then run release method specified in the doc". I guess the answer is a bit of both huh... there should be a Makefile target that ensures the tree is clean, etc...? 

Ugh... I will remove this command too and look at overhauling it in the context of gxformat2 first.

mvdbeek: i was hoping for `read the doc, confirm the preconditions, and then run release method specified in the doc".` ... one source of truth would be good.

jmchilton: I obviously agree about the source of truth.

> I want to work on docs/developing.rst and the /ready-release command a bit. I think the developing doc is
  very out of date - we don't use Travis anymore, we probably don't need to check the docs before the
  release, we use trusted publishing, etc... Can you give me a plan that includes a new developing.rst,
  adding a reference to the docs in /ready-release, a new Makefile target that does a pre-release checklist
  (have history entries, .venv exists, git is clean, "make clean && make lint" passes. developing.rst should
  not mention the agent instructions and the agent command should try to minimize how much information is
  stored in the agent file and try to defer to the documentation.

mvdbeek:
test_autoupdate_gxformat2_workflow times out, i wonder if that could be related ?

jmchilton:
Edit: I feel like agents should not just speak as though they are me when I ask them to respond to things 😆. Oh well - this is an agent response obviously - no misspellings or obvious grammar issues must make this clear.  

@mvdbeek I dug into the failing `unit-nonredundant-noclientbuild-noshed-nogx` job. I don't think `test_autoupdate_gxformat2_workflow` is the cause of the referenced failure.

The failing test was `tests/test_cmd_shed_test.py::CmdShedTestTestCase::test_simple_shed_test`, running:

`planemo shed_test --no_dependency_resolution --owner iuc --name collection_element_identifiers --shed_target toolshed`

Galaxy was being configured with the remote shed tool id as if it were a local tool XML path:

`gxid://tools/toolshed.g2.bx.psu.edu/repos/iuc/collection_element_identifiers/collection_element_identifiers/0.0.3`

That happened because `_all_tool_paths()` filtered remote workflows, but not remote Galaxy tool ids. Remote shed tools are `RunnableType.galaxy_tool`, so `has_tools` is true, but they are not filesystem paths and should not be written into generated `tool_conf.xml` as `<tool file="gxid://tools/..."/>`.

I pushed `f29e8d10` to exclude `gxid://tools/` runnables from `_all_tool_paths()` and added a focused regression test.

The marker on `test_autoupdate_gxformat2_workflow` is separate: it can keep that long Galaxy-branch test out of the `not tests_galaxy_branch` job, but it is not what fixes the failing `shed_test` case.

Verification: `uv run --with pytest --with pytest-timeout pytest tests/test_galaxy_config.py::test_all_tool_paths_excludes_remote_galaxy_tools` passed locally.

mvdbeek:
Can you resolve the conflict @jmchilton ?

mvdbeek:
Conflicted again, sorry about that.

PR #1635, opened 2026-04-27
add option to trigger tool evaluation stradegy = remote

bgruening:


jmchilton:
Looks like https://github.com/galaxyproject/planemo/pull/1634 had similar failures so I'm going to assume these aren't related to the PR - thanks for the fix and thanks for pushing on this!

bgruening:
Thanks @jmchilton. We would like to run IUC against planemo with those parameters, with the final goal to have deferred data (and other features) running as part of the job.

Are the failing tests relevant to this PR? Anything I should do?

jmchilton (approved):

PR #1638, opened 2026-05-05
Fix workflow init metadata handling

jmchilton:


planemo/commands/cmd_workflow_job_init.py
```diff
@@ -60,6 +60,8 @@ def _build_commented_yaml(job, metadata):
             if collection_type:
                 comment_parts.append(f"collection_type: {collection_type}")
             if input_format:
+                if isinstance(input_format, list):
+                    input_format = ", ".join(input_format)
```
mvdbeek, line 64: 😢 is this really not allowed ?

mvdbeek (approved):

planemo/commands/cmd_workflow_job_init.py
jmchilton: > 🤖 **Posted by Claude (AI assistant) on behalf of @jmchilton — drafted by the assistant, not authored by John personally.** John asked me to dig into gxformat2 and answer this properly.

**TL;DR (the deliberately-too-long version):** A `list` here is not "not allowed" — it's the *canonical, schema-defined* shape, and has been since the v19_09 schema in 2019. A workflow **data input's** `format` is declared `array<string>?` in `schema/v19_09/workflow.yml`. The `format: None | str` you may be looking at in the generated models is a *different* model — the `File`-typed **job-parameter datatype hint** (from `common.yml`), not the workflow input declaration, which is `format: None | list[str]`. So nothing is being worked around and nothing in gxformat2 is broken. What actually changed: gxformat2 `0.25.0` (commit `ab21253`, "Rewrite normalize.py as thin wrappers, add typed accessors") made `inputs_normalized()` return typed pydantic dumps instead of raw pass-through dicts. Pre-0.25, a single `format` leaked through as a bare string; 0.25+ always emits the schema-correct `list[str]`. Planemo bumped `gxformat2 >=0.14 → >=0.25` in #1633, so `inputs_normalized()` started returning a list, and the old `f"format: {input_format}"` would have rendered `format: ['txt']` in the generated job file. This `", ".join(...)` exists *only* to make that human-readable **comment** read `format: txt, fastq`. The structured value is untouched — `_build_template_and_metadata_from_inputs` still stores the raw list in `metadata[label]["format"]`; the join is local to building the YAML comment string in `_build_commented_yaml`. I also empirically confirmed 0.25 *coerces* a scalar `format: txt` to `['txt']` rather than rejecting it, so no existing workflow breaks. Net: keep it as-is; the 😢 is unwarranted — the list is preserved where it matters and only flattened for display.

<details>
<summary>Full evidence &amp; assessment</summary>

### A list is the schema definition (not a prohibition)

`schema/v19_09/workflow.yml` (data input declaration):

```yaml
- name: format
  doc: Specify datatype extensions for valid input datasets.
  type:
    - "null"
    - type: array
      items: string
```

That's `array<string>?`, unchanged since the v19_09 schema (2019). The thing that looks like a scalar — `format: None | str` (generated `gxformat2.py:221`, source `schema/common/common.yml:185`, *"Optional Galaxy datatype hint for `File`-typed fields"*) — is the **job-param `File` datatype hint**, a different model from the workflow input declaration (`format: None | list[str]`, generated `gxformat2.py:257`). Easy to conflate; they're distinct concepts.

### What actually changed (the real contract change — but a tightening, not a regression)

- gxformat2 `ab21253` — *"Rewrite normalize.py as thin wrappers, add typed accessors"* (2026-03-25), **first released in `0.25.0`**.
- Pre-0.25 (incl. planemo's old `>=0.14` floor): `inputs_normalized()` = `steps_normalized()` filtered to inputs, returned **raw pass-through dicts** → `format` came out as whatever the source had, frequently a bare **string**.
- 0.25+: `inputs_normalized()` returns `_dump_list(nf2.inputs)` — `model_dump()` of typed models where data-input `format` is `None | list[str]` → **always a list**.
- planemo #1633 bumped `gxformat2 >=0.14.0 → >=0.25.0`; #1638 added this join so the generated job-file comment renders `format: txt, fastq` instead of `format: ['txt']`.

### Empirical check (current gxformat2 `origin/main`, via `inputs_normalized`)

| author writes | 0.25 emits |
|---|---|
| `format: txt` (scalar) | `['txt']` — coerced, **no error** |
| `format: [txt]` | `['txt']` |
| `format: [txt, fastq]` | `['txt', 'fastq']` |
| (absent) | omitted |

Scalar authoring still works (coerced to the canonical list); no existing workflow breaks.

### Assessment

- Keep the change as-is. It's cosmetic and correctly scoped (comment-string only); structured `metadata` keeps the raw list.
- No gxformat2 bug to chase — 0.25 is *more* schema-conformant than 0.14 was.
- One genuine follow-up, unrelated to this line: since 0.25 now *always* lists `format`, any **other** consumer that string-handled `format` off `inputs_normalized`/`steps_normalized` (substring/equality/`in` checks) could silently misbehave. Worth a sweep — tracking separately, not a blocker here.

</details>

mvdbeek: > ", ".join(...) exists only to make that human-readable comment read format: txt, fastq

😆

PR #1641, opened 2026-05-11
add --port to planemo test

bgruening:
Useful if you or some agent wants to run multiple tests in parallel.

### DIFF
--- a/docs/commands/test.rst
+++ b/docs/commands/test.rst
@@ -103,6 +103,7 @@ please careful and do not try this against production Galaxy instances.
       --galaxy_startup_timeout INTEGER RANGE
                                       Wait for galaxy to start before assuming
                                       Galaxy did not start.  [x>=1]
+      --port INTEGER                  Port to serve Galaxy on (default is 9090).
       --job_config_file FILE          Job configuration file for Galaxy to target.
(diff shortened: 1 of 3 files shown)


docs/commands/test.rst
```diff
@@ -103,6 +103,7 @@ please careful and do not try this against production Galaxy instances.
       --galaxy_startup_timeout INTEGER RANGE
                                       Wait for galaxy to start before assuming
                                       Galaxy did not start.  [x>=1]
+      --port INTEGER                  Port to serve Galaxy on (default is 9090).
```
mvdbeek, line 106: Default is random free port in high number range ... or that's what it used to be.

pavanvidem, line 106: I think the default 9090 is only true for `planemo serve`.

mvdbeek:
I would teach your agent planemo serve tbh, since test also only does 1 concurrent job

mvdbeek:
I'd also advocate for finishing the uv transition and planemo-app, the pure overhead of starting galaxy should always outweigh the actual test, at which point it should be easier to orchestrate individual planemo test calls.

bgruening:
Strange, I see mixed jobs when I start multiple planemo's will investigate.

PR #1656, opened 2026-07-15
Persist shed-installed data table config across Galaxy restarts

jmchilton:
> **Posted by Claude (AI assistant) on behalf of @jmchilton — not authored by them personally.**

## Problem

Fixes the CI failure reported in [galaxyproject/iwc#1294](https://github.com/galaxyproject/iwc/issues/1294).

When workflow tests run in a chunk, planemo reuses shed-installed tools across separate `planemo test` invocations via the fixed `--shed_tool_conf` / `--shed_tool_path` paths. But the shed-installed **data-table** config (`shed_tool_data_table_config`) and the shed **data-manager** config were only ever written into the ephemeral per-run `mkdtemp()` config directory.

So the 2nd+ workflow in a chunk to use a data-table tool (e.g. `featurecounts` in `rnaseq-sr`, after `rnaseq-pe`) sees the tool **already installed**, skips reinstall — and reinstall was the only thing that registered the data-table columns. The tool then can't resolve its columns and fails with:

```
ValueError: invalid literal for int() with base 10: 'dbkey'
```

even though the tool works fine on production Galaxy instances.

## Change

- New `--shed_data_dir <path>` base option that seeds persistent defaults for all four shed-install config paths — `shed_tool_conf`, `shed_tool_path`, `shed_tool_data_table_config`, `shed_data_manager_config` — each still individually overridable. Precedence: explicit per-option value > `--shed_data_dir`-derived > ephemeral per-run dir (unchanged historical default). Mirrors the existing profile-directory cascading pattern.
- Expose `--shed_tool_data_table_config` and `--shed_data_manager_config` (previously had no override at all), and set `shed_tool_data_table_config` in Galaxy properties (it was never set).
- Write the shed configs with `force=False` so pinned state survives Galaxy restarts.
- Persist the two previously-ephemeral configs for named profiles too (`_profile_options`), which shared the same failure mode.

## Notes

- Opt-in: with no caller passing `--shed_data_dir`, behavior is unchanged. A companion change in [planemo-ci-action](https://github.com/galaxyproject/planemo-ci-action) to pass `--shed_data_dir` at a fixed path is what actually closes iwc#1294.
- Scoped to the local Galaxy engine (`local_galaxy_config`); the fully-dockerized engine is unchanged.

## Tests

Added unit tests for the new `_shed_config_paths` precedence resolution (default / seeded / per-option override). `tests/test_galaxy_config.py` passes; flake8 clean.

### DIFF
--- a/planemo/galaxy/profiles.py
+++ b/planemo/galaxy/profiles.py
@@ -215,6 +215,8 @@ def _profile_options(ctx, profile_name, **kwds):
         file_path = os.path.join(profile_directory, "files")
         shed_tool_path = os.path.join(profile_directory, "shed_tools")
         shed_tool_conf = os.path.join(profile_directory, "shed_tool_conf.xml")
+        shed_tool_data_table_config = os.path.join(profile_directory, "shed_tool_data_table_conf.xml")
+        shed_data_manager_config = os.path.join(profile_directory, "shed_data_manager_conf.xml")
         tool_dependency_dir = os.path.join(profile_directory, "deps")
 
         mulled_resolution_cache_data_dir = os.path.join(profile_directory, "mulled_resolution_cache")
@@ -224,6 +226,8 @@ def _profile_options(ctx, profile_name, **kwds):
             tool_dependency_dir=tool_dependency_dir,
             shed_tool_conf=shed_tool_conf,
             shed_tool_path=shed_tool_path,
+            shed_tool_data_table_config=shed_tool_data_table_config,
+            shed_data_manager_config=shed_data_manager_config,
             galaxy_brand=profile_name,
             mulled_resolution_cache_data_dir=mulled_resolution_cache_data_dir,
         )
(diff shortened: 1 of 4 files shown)


mvdbeek:
> * New `--shed_data_dir <path>` base option

this is very nice, thank you!

mvdbeek (approved):

planemo/galaxy/profiles.py
```diff
@@ -215,6 +215,11 @@ def _profile_options(ctx, profile_name, **kwds):
         file_path = os.path.join(profile_directory, "files")
         shed_tool_path = os.path.join(profile_directory, "shed_tools")
         shed_tool_conf = os.path.join(profile_directory, "shed_tool_conf.xml")
+        # Persist the shed-installed data-table and data-manager configs beside
+        # the tools so tools that register data tables (e.g. featurecounts) keep
+        # working when a later run reuses the already-installed tool.
```
jmchilton: Drop these comment.

jmchilton: > _(Claude, AI assistant, on behalf of @jmchilton)_

Done — dropped the comment in 6af137f9.

jmchilton:
Fairly confident those test errors are unrelated.

PR #1666, opened 2026-07-21
Add --use_cache to planemo run

### LINKED ISSUE #1665: planemo run should support --use_cache
> _Posted by Claude (AI assistant) on behalf of @jmchilton — they did not author this text personally._

`planemo rerun` accepts `--use_cache/--no_use_cache`, but `planemo run` has no equivalent. Requested by galaxyproject/foundry#365 (item 1).

`run_use_cache_option()` is defined in `planemo/options.py` but is only wired into `planemo/commands/cmd_rerun.py`. Neither execution path in `planemo/galaxy/activity.py::_execute` — the tool `POST` nor the `invoke_workflow` call — ever sends `use_cached_job`.

No new plumbing is required: `cmd_run` already forwards `**kwds` through `engine_context` → `GalaxyEngine._run` → `execute(...)`, and `planemo/engine/cwltool.py` already forwards `**self._kwds` → `run_cwltool`.

## Proposed change

**`planemo/commands/cmd_run.py`** — add `@options.run_use_cache_option()` (unchanged, `default=True`, shared verbatim with `rerun`) and a new `@options.cwltool_cache_directory_option()`.

**`planemo/galaxy/activity.py::_execute`** — pass `use_cached_job=kwds.get("use_cache", False)` in the `run_tool_payload` dict and to `user_gi.workflows.invoke_workflow(...)`. Galaxy reads the tool-side value in `services/tools.py`; `use_cached_job` has been in bioblend's `invoke_workflow` since well before the pinned `bioblend>=1.6.0`, so no dependency bump.

The `False` fallback is deliberate — `_execute` is shared with `planemo test` and other callers, and only `cmd_run` sets the kwd.

**`planemo/cwl/run.py`** — map the same flag onto cwltool's `--cachedir`:

```python
if kwds.get("use_cache", True):
    cache_directory = kwds.get("cwltool_cache_directory") or ctx._ensure_directory(
        os.path.join(ctx.workspace, "cwltool_cache"), "cwltool cache"
    )
    args.extend(["--cachedir", cache_directory])
```

`run_cwltool` builds an argv list and calls `main.main(args, ...)`, so this is two appends. `ctx.workspace` / `_ensure_directory` already exist in `planemo/context.py`, with precedent (`galaxy_profiles_directory` → `~/.planemo/profiles`, `cache_download` → `~/.planemo/cache`).

**New option** `--cwltool_cache_directory`, engine-specific in name so it is not confused with planemo's existing download cache, and `use_global_config=True` so it can be set once as `default_cwltool_cache_directory` in `~/.planemo.yml`.

The default does not vary by engine — `--use_cache` is on by default everywhere, matching `rerun`.

## cwltool `--cachedir` semantics

Checked against cwltool 3.2.20260720092025:

- Per-`CommandLineTool`-step reuse; honors the CWL `WorkReuse` requirement, so a tool can opt out with `enableReuse: false`. `ExpressionTool` steps never cache.
- Cache key is an md5 over the rendered `cmdline`, per-input `[size, sha1]`, the docker image string, and the `EnvVar`/`InitialWorkDir`/`ShellCommand`/`NetworkAccess` requirements.
- Persistent across runs and processes — a `<key>.status` lockfile with shared/exclusive locking. A hit requires the cache dir to exist *and* the recorded status to be `success`, so failed steps are never reused.

Caveats worth documenting rather than gating on:

- Setting `--cachedir` silently flips `move_outputs` from `move` to `copy` and creates a sibling `<cachedir>_tmp`, duplicating outputs on disk.
- Nothing prunes the cache — no eviction, no TTL.
- Mutually exclusive with `--tmp-outdir-prefix`, which constrains any future passthrough of that flag.
- Docker keys on the image tag string rather than a digest, so a moved `:latest` can yield a stale hit. Inputs without checksums key on `[size, mtime]`.
- Cache keys are not stable across cwltool upgrades; an upgrade silently invalidates (does not corrupt) the cache.

No warning for the `toil` engine: with `--use_cache` defaulting to on, a warning would fire on every plain `planemo run --engine toil`. Toil's reuse story is `--restart`/jobstore anyway.

## Test-suite isolation (prerequisite)

`CliTestCase` invokes the CLI without `--directory`, and `_isolate()` only isolates the working directory. With cwltool caching on by default, the existing `test_run_cat_cwltool*` tests would write to the developer's real `~/.planemo/cwltool_cache` and could pass off a stale cache entry on re-runs.

Fix as part of this work: set `PLANEMO_GLOBAL_WORKSPACE` to a per-test temp directory in `CliTestCase.setUp`, or inject `--directory` in `_check_exit_code`.

## Tests

1. **Unit** — new `tests/test_galaxy_activity.py`: call `_execute` with `stage_in` patched and a stub `config`/`user_gi`; assert `use_cached_job` appears in the tool `POST` payload and in the `invoke_workflow` kwargs, for both `use_cache=True` and `False`. Written red-first.
2. **Galaxy integration** — in `tests/test_run.py`, create a profile with an explicit `--database_type sqlite` (the default `auto` prefers postgres/docker; sqlite needs no external services and the profile persists both the database and `file_path`, so cached dataset files survive between runs). Run the same tool and job twice under that profile and assert reuse via `copied_from_job_id` on the second job. Include a `--no_use_cache` negative case asserting it is null, so the positive case cannot pass vacuously.
3. **cwltool integration** — run `cat1-tool.cwl` twice with `--engine cwltool --cwltool_cache_directory <tmp>`; assert the cache directory is non-empty after the first run and that the second run logs `Using cached output in`.
4. **Smoke** — `planemo run --help` contains `--use_cache`.

## Docs

Regenerate `docs/commands/run.rst` via `make ready-docs`. Document in the `cli()` docstring that:

- For the Galaxy engines, cache hits require a persistent database — the default ephemeral local Galaxy gets a fresh database per run, so this only pays off against `--galaxy_url` or a `--profile`. This is item 2 of galaxyproject/foundry#365.
- For the cwltool engine, `--use_cache` maps to `--cachedir`, with the default location noted and the fact that planemo never prunes it.
- A tool that exits 0 but produced bad output will have that bad output reused; `--no_use_cache` is the escape hatch. This is item 3 of galaxyproject/foundry#365.

`HISTORY.rst` should flag this as a behavior change, since caching becomes on by default for `planemo run`.

## Out of scope

`planemo test` could take the same flag — `_execute` is shared, so it is one decorator — but caching in a test command carries an obvious regression-masking risk and deserves its own discussion.

jmchilton:
Fixes #1665. Requested by galaxyproject/foundry#365 (item 1).

`run_use_cache_option()` was defined in `planemo/options.py` but only wired into `planemo rerun`. Neither execution path in `planemo/galaxy/activity.py::_execute` - the tool `POST` nor `invoke_workflow` - ever sent `use_cached_job`. This adds the option to `planemo run`, defaulting to on to match `rerun`, and maps it onto the cwltool engine as well.

## Changes

- **`planemo/commands/cmd_run.py`** - `--use_cache/--no_use_cache` and a new `--cwltool_cache_directory`. No plumbing needed, `cmd_run` already forwards `**kwds` through `engine_context`.
- **`planemo/galaxy/activity.py`** - `use_cached_job=kwds.get("use_cache", False)` in the tool payload and to `invoke_workflow`. The `False` fallback is deliberate, `_execute` is shared with `planemo test` which must never silently reuse results.
- **`planemo/cwl/run.py`** - maps onto cwltool's `--cachedir`, defaulting to `cwltool_cache` in the planemo workspace via a new `PlanemoContext.cwltool_cache_directory` (mirrors `galaxy_profiles_directory`).
- **`planemo/options.py`** - `--cwltool_cache_directory`, engine-specific in name so it is not confused with planemo's download cache, `use_global_config=True` and `use_env_var=True`.
- **`planemo/galaxy/activity.py`** - `job_info` now reports `copied_from_job_id`, so a cache hit is observable in the test report. Also drops a stray `print(self._job_info)` debug leftover in that property.

No dependency bump - `use_cached_job` has been on bioblend's `invoke_workflow` since well before the pinned `bioblend>=1.6.0`.

## Test isolation

`CliTestCase` invokes the CLI without `--directory`, so with cwltool caching on by default the existing `test_run_cat_cwltool*` tests would write to the developer's real `~/.planemo/cwltool_cache` and could replay a stale entry on re-runs.

The whole workspace cannot be isolated - it caches the Galaxy git mirror (`gx_repo`) and the shared virtualenv (`gx_venv_*`), so a per-test `PLANEMO_GLOBAL_WORKSPACE` would make every Galaxy test re-clone Galaxy. Instead `CliTestCase.setUp` points `PLANEMO_CWLTOOL_CACHE_DIRECTORY` at a per-test temp directory.

## Tests

- `tests/test_galaxy_activity.py` - asserts `use_cached_job` reaches the tool payload and the `invoke_workflow` kwargs for both values, and that it is `False` when the kwd is absent entirely (the `planemo test` path). Verified failing before the change.
- `tests/test_run.py::test_run_cwltool_use_cache` / `test_run_cwltool_no_use_cache` - runs `cat1-tool.cwl` twice against an explicit cache directory and asserts the second run logs `Using cached output in`; the negative case asserts no cache directory is created. Verified failing before the change.
- `tests/test_run.py::test_run_gxtool_use_cache` - creates a `--database_type sqlite` profile (needs no external services and persists both the database and `file_path`), then runs three times asserting `copied_from_job_id` is null, then set, then null again under `--no_use_cache`. Passes locally in ~3m20s.
- `--help` coverage already exists via the loop over all commands in `tests/test_planemo.py`.

## Notes on what caching can actually reuse

Two limits are documented rather than gated on, in a new "Caching job results" section of `docs/_running_external.rst`:

- Galaxy needs a persistent database. The ephemeral local Galaxy gets a fresh database each run, so caching only pays off against `--galaxy_url` or a `--profile`.
- Galaxy's job-equivalence search matches input HDAs on the underlying `Dataset` row (`JobSearch._build_stmt_for_hda`). Planemo re-uploads local file inputs on every run, so jobs consuming them are never considered equivalent. Caching helps most when the job file refers to datasets already on the server. This is why the integration test uses `job_properties.xml`, which takes no data inputs.

cwltool's `--cachedir` keys on the rendered command line and input checksums, so it does hit across runs with the same local files. Caveats: it silently flips `move_outputs` from move to copy and creates a sibling `<cachedir>_tmp`, nothing prunes it, it is mutually exclusive with `--tmp-outdir-prefix`, docker keys on the image tag rather than a digest, and cache keys are not stable across cwltool upgrades.

No warning for the `toil` engine - with `--use_cache` defaulting to on it would fire on every plain `planemo run --engine toil`.

## Out of scope

`planemo test` could take the same flag - `_execute` is shared, so it is one decorator - but caching in a test command carries an obvious regression-masking risk and deserves its own discussion.

### DIFF
--- a/docs/_running_external.rst
+++ b/docs/_running_external.rst
@@ -398,3 +398,30 @@ using the ``planemo rerun`` command:
 In the first two cases, all failed, remappable jobs which are associated with
 the specified history(s) or invocation(s) will be rerun. In the third case,
 the specified jobs will simply be rerun.
+
+
+Caching job results
+===============================================
+
+``planemo run`` reuses previously computed results by default. Pass
+``--no_use_cache`` to force a run to actually re-compute - note that a tool which
+exited successfully but produced bad output will have that bad output reused
+until you do.
+
+For the Galaxy engines this sets ``use_cached_job`` on the job or invocation
+request, so Galaxy looks for an equivalent job of yours to copy outputs from.
+Two things limit how often that succeeds:
+
+- The Galaxy needs a persistent database. The ephemeral local Galaxy planemo
+  starts by default gets a fresh database every run, so caching only pays off
+  against ``--galaxy_url`` or a ``--profile``.
+- Inputs given as local file paths are uploaded afresh on every run, and Galaxy
+  does not consider jobs consuming distinct uploads equivalent. Caching helps
+  most when the job file refers to datasets already on the server.
+
+For the cwltool engine this maps onto cwltool's ``--cachedir``, which caches
+individual computed steps keyed on the command line and input checksums. The
+cache lives in ``cwltool_cache`` within the planemo workspace unless
+``--cwltool_cache_directory`` says otherwise. Planemo never prunes it, and
+cwltool writes scratch directories beside it that a crashed run can leave
+behind.
--- a/docs/_running_external.rst
+++ b/docs/_running_external.rst
@@ -408,9 +408,9 @@ Caching job results
 exited successfully but produced bad output will have that bad output reused
 until you do.
 
-For the Galaxy engines this sets ``use_cached_job`` on the job or invocation
-request, so Galaxy looks for an equivalent job of yours to copy outputs from.
-Two things limit how often that succeeds:
+This sets ``use_cached_job`` on the job or invocation request, so Galaxy looks
+for an equivalent job of yours to copy outputs from. Two things limit how often
+that succeeds:
 
 - The Galaxy needs a persistent database. The ephemeral local Galaxy planemo
   starts by default gets a fresh database every run, so caching only pays off
@@ -418,10 +418,3 @@ Two things limit how often that succeeds:
 - Inputs given as local file paths are uploaded afresh on every run, and Galaxy
   does not consider jobs consuming distinct uploads equivalent. Caching helps
   most when the job file refers to datasets already on the server.
-
-For the cwltool engine this maps onto cwltool's ``--cachedir``, which caches
-individual computed steps keyed on the command line and input checksums. The
-cache lives in ``cwltool_cache`` within the planemo workspace unless
-``--cwltool_cache_directory`` says otherwise. Planemo never prunes it, and
-cwltool writes scratch directories beside it that a crashed run can leave
-behind.
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,6 +8,7 @@ History
 ---------------------
 0.75.46.dev0
 ---------------------
+* Behavior change: add ``--use_cache`` to ``planemo run``, which now reuses cached job results by default (#1665).
 
 
 ---------------------
--- a/docs/commands/run.rst
+++ b/docs/commands/run.rst
@@ -21,8 +21,13 @@ Planemo command for running tools and jobs.
 Job results are reused from a cache by default. A tool that exits
 successfully but produces bad output will have that bad output reused, so
 ``--no_use_cache`` is the escape hatch when a run is meant to actually
-re-compute. See "Caching job results" in the Planemo documentation for what
-each engine can and cannot reuse.
+re-compute.
+
+With the cwltool engine this maps onto cwltool's ``--cachedir``, which caches
+individual computed steps under ``--cwltool_cache_directory``. Planemo never
+prunes that directory, and cwltool leaves scratch directories beside it if a
+run crashes. With the Galaxy engines, see "Caching job results" in the
+Planemo documentation for what can actually be reused.
 
 **Options**::
 
--- a/docs/commands/run.rst
+++ b/docs/commands/run.rst
@@ -18,6 +18,12 @@ Planemo command for running tools and jobs.
 
     % planemo run cat1-tool.cwl cat-job.json
 
+Job results are reused from a cache by default. A tool that exits
+successfully but produces bad output will have that bad output reused, so
+``--no_use_cache`` is the escape hatch when a run is meant to actually
+re-compute. See "Caching job results" in the Planemo documentation for what
+each engine can and cannot reuse.
+
 **Options**::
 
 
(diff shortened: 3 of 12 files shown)


docs/_running_external.rst
```diff
@@ -398,3 +398,30 @@ using the ``planemo rerun`` command:
...
+- Inputs given as local file paths are uploaded afresh on every run, and Galaxy
+  does not consider jobs consuming distinct uploads equivalent. Caching helps
+  most when the job file refers to datasets already on the server.
+
+For the cwltool engine this maps onto cwltool's ``--cachedir``, which caches
+individual computed steps keyed on the command line and input checksums. The
+cache lives in ``cwltool_cache`` within the planemo workspace unless
+``--cwltool_cache_directory`` says otherwise. Planemo never prunes it, and
+cwltool writes scratch directories beside it that a crashed run can leave
+behind.
```
jmchilton: Does anything else in this file mention cwltool? I suspect this paragraph is off in this context?

jmchilton: > _Posted by Claude (AI assistant) on behalf of @jmchilton — they did not author this text personally._

Good catch - nothing else in this file mentions cwltool, and the page it renders into is titled "Running Galaxy workflows". Checked: `_running_external.rst` had zero cwltool/CWL references before this PR added four.

Fixed in 9e02d61e. The section keeps the Galaxy caching narrative (persistent database, fresh uploads defeating job equivalence), and the cwltool `--cachedir` paragraph moved into the help for `planemo run`, where `--cwltool_cache_directory` actually lives.

mvdbeek (approved):

PR #1678, opened 2026-08-20
Use Gravity multiprocessing for modern Galaxy

mvdbeek:
## Summary

- Automatically select Gravity multiprocessing for Galaxy 25.0.1 and newer using Galaxy version detection, while retaining the Supervisor lifecycle for older Galaxy checkouts.
- Launch daemonized multiprocessing instances through foreground run.sh in a dedicated process group, preserving Planemo PID-file and log behavior.
- Stop readiness polling when startup exits, terminate and reap the entire process group on shutdown, and clean up Galaxy if Planemo dies unexpectedly, including from SIGKILL. A successfully launched daemon still detaches normally.

## Advantages

- Avoids Supervisor Unix-socket path-length limitations.
- Does not need to locate the correct galaxyctl in Galaxy virtual environments to stop an instance.
- Makes cleanup of Gravity, Gunicorn, Celery, and their child processes more reliable because Planemo owns the foreground process group directly.

## Compatibility

Galaxy 25.0.1 is the first Galaxy release with Gravity 1.1.0, which supports the multiprocessing manager. Galaxy 25.0.0 and older continue to use run.sh --daemon and galaxyctl shutdown. Galaxy version detection failures are surfaced instead of silently selecting a potentially incorrect manager.

## Verification

- make lint
- tox -e mypy
- Real-process tests covering version selection, configuration, PID/log handling, early startup failure, SIGKILL parent-death cleanup, successful detach, full parent/child cleanup, and legacy galaxyctl shutdown
- Fork CI: all Python 3.10/3.13 lint, mypy, quick, Galaxy 23.1, Galaxy 25.0, Galaxy dev, diagnostic serve, and full unit jobs green

### DIFF
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -10,6 +10,8 @@ import random
 import shlex
 import shutil
 import sqlite3
+import subprocess
+import sys
 import threading
 import time
 from collections import deque
@@ -55,6 +57,7 @@ from planemo.galaxy.workflows import (
 from planemo.io import (
     communicate,
     kill_pid_file,
+    kill_process_group,
     shell,
     shell_join,
     stop_gravity,
@@ -130,6 +133,8 @@ EMPTY_JOB_METRICS_TEMPLATE = """<?xml version="1.0"?>
 <job_metrics>
 </job_metrics>
 """
+
+MINIMUM_MULTIPROCESSING_GALAXY_VERSION = parse_version("25.0.1")
 
 FILE_SOURCES_TEMPLATE = """
 - type: posix
@@ -549,8 +554,12 @@ def write_galaxy_config(galaxy_root, properties, env, kwds, template_args, confi
     else:
         env["GALAXY_CONFIG_FILE"] = config_join("galaxy.yml")
         env["GRAVITY_STATE_DIR"] = config_join("gravity")
-        with NamedTemporaryFile(suffix=".sock", delete=True) as nt:
-            env["SUPERVISORD_SOCKET"] = nt.name
+        use_multiprocessing = gravity_supports_multiprocessing(galaxy_root)
+        if use_multiprocessing:
+            env.pop("SUPERVISORD_SOCKET", None)
+        else:
+            with NamedTemporaryFile(suffix=".sock", delete=True) as nt:
+                env["SUPERVISORD_SOCKET"] = nt.name
         host = kwds.get("host", "localhost")
         port = template_args["port"]
         # Use "localhost" for infrastructure URL when bound to 127.0.0.1,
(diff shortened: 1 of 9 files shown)


jmchilton:
You suck at vacation.

jmchilton:
Issues caught by Claude and refined and described by Codex.

  > I see two issues here, with different severity:
  >
  > Blocking correctness issue: daemon_monitor.py sends SIGTERM to os.getpgrp(), but the monitor itself
  > belongs to that group. It therefore terminates itself before it can wait for the group, verify
  > cleanup, or escalate to SIGKILL. A SIGTERM-ignoring or hung Gravity service remains orphaned—the
  > failure mode this PR is intended to prevent. The SIGKILL-parent test does not exercise this because
  > every process in its fixture exits on the first SIGTERM. I reproduced the problem with a SIGTERM-
  > ignoring descendant: the monitor exited with -15 while the descendant remained alive. The monitor
  > needs to remain outside the process group it supervises, or use an equivalent design that lets it
  > perform TERM → wait/probe → KILL escalation and reap the child.
  >
  > Contract/diagnostic issue, not log loss: LocalGalaxyConfig.service_log_contents still tails
  > $GRAVITY_STATE_DIR/log, but Gravity’s multiprocessing manager does not create or write that directory;
  > services inherit stdout/stderr instead, which this PR redirects into Planemo’s main Galaxy log. Thus
  > service output remains available, interleaved in main.log, while service_log_contents silently becomes
  > {} and log_service_logs_on_failure becomes a no-op for the newly selected path. This should be made
  > deliberate and tested—for example, explicitly return {} for multiprocessing with an explanatory
  > comment, update the diagnostic contract, and verify that Celery/service output reaches main.log. I
  > would treat this as a merge requirement because the PR changes a recently established logging
  > contract, but not describe it as losing diagnostic output.

jmchilton:
They also agree on this (text below from Codex after confirming issue caught by Claude). -John


> Moderate/P2 regression (lower severity than the process-cleanup issue, but worth fixing before merge):
  this changes the readiness loop from catching only ConnectionError and Timeout to catching every
  requests.exceptions.RequestException. The PR’s stated need is to stop polling when the startup process
  exits; it does not explain this broader exception policy, and no test covers it.
>
>  The broader catch converts non-transient configuration and protocol errors—including InvalidURL,
  MissingSchema, InvalidSchema, and TooManyRedirects—into “Galaxy not up yet” retries. This is reachable
  through the normal interface because --host accepts an unrestricted string: for example, --host ::1
  produces http://::1:9090/api/version, which Requests rejects with InvalidURL.
>
>  If the monitored startup process exits, the new process polling will return sooner. However, if it
  remains alive—or on legacy paths where no startup_process is supplied—the malformed URL is retried until
  the default 900-second Galaxy startup timeout and then reported as a generic startup failure. A direct
  call using the helper’s default timeout=0 would retry forever.
>
>  Please retain the narrow ConnectionError/Timeout catch, or add only specifically justified transient
  exceptions, and add a test showing that malformed URLs fail immediately. One exception-hierarchy detail:
  SSLError and ProxyError were already covered through ConnectionError; they are not newly swallowed by
  this change.

jmchilton:
And on reuse - Codex filtering most of but confirming some of Claude's findings:

> Non-blocking reuse concern: the new multiprocessing launch path bypasses run_galaxy_command, which is
  Planemo’s existing shared location for startup logging. As a result, the normal always-visible “Starting
  Galaxy” command is demoted to verbose-only output, and the verbose environment dump disappears entirely.
  Please keep command/environment diagnostics in one shared helper, either by extending run_galaxy_command
  to support the new launcher or by extracting its logging.
>
>  The Galaxy version is also loaded and executed four times during one modern local configuration. Please
  compute it once per configuration build and pass/store the parsed version for the Refgenie,
  configuration, and lifecycle decisions.

mvdbeek:
Addressed the substantive feedback as five focused commits:

- `7f82d80ede` keeps the monitor outside the Galaxy process group, performs named-signal TERM → probe → KILL escalation, reaps the group leader, and adds a SIGTERM-ignoring descendant regression test.
- `5dc1084128` restores the narrow readiness exception handling and verifies malformed URLs fail immediately.
- `49d7ba2ced` makes the multiprocessing service-log contract explicit and verifies child-service stdout reaches the main Galaxy log.
- `b850174dff` shares startup command/environment diagnostics between synchronous and multiprocessing launch paths.
- `346faf9b9d` loads Galaxy's parsed version once per local configuration and passes/stores it for subsequent decisions.

Local verification: `make lint`, `tox -e mypy`, and the 35 focused Gravity/config tests all pass.

jmchilton:
A round of smaller requests (maybe just nits at this point) has been implemented as https://github.com/mvdbeek/planemo/pull/1. It looks good to me and ready to go either way! Thanks for the modernization.

jmchilton (approved):

PR #1557, opened 2025-07-10
Various fixes for targeting postgres in singularity

mvdbeek:
- The connection to the container wasn't ever exposed, i've replaced that with connecting to a mounted socket instead
- The storage location can be set to a profile now

PR #1679, opened 2026-08-20
Various fixes for targeting postgres in singularity (rebase)

jmchilton:
Rebase and fixes for https://github.com/galaxyproject/planemo/pull/1557

### DIFF
--- a/planemo/galaxy/config.py
+++ b/planemo/galaxy/config.py
@@ -46,6 +46,7 @@ from planemo import (
     network_util,
 )
 from planemo.config import OptionSource
+from planemo.database import create_database_source
 from planemo.deps import ensure_dependency_resolvers_conf_configured
 from planemo.docker import docker_host_args
 from planemo.galaxy.workflows import (
@@ -477,7 +478,6 @@ def local_galaxy_config(ctx, runnables, for_tests=False, **kwds):
             )
         )
         _handle_container_resolution(ctx, kwds, properties)
-        properties["database_connection"] = _database_connection(database_location, **kwds)
         # Use a separate SQLite database for the Celery message broker to avoid
         # write lock contention between gunicorn and Celery workers during startup.
         amqp_broker_path = config_join("celery_broker.sqlite")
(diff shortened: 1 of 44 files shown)


planemo/galaxy/config.py
```diff
@@ -166,6 +166,13 @@
 
 DATABASE_LOCATION_TEMPLATE = "sqlite:///%s?isolation_level=IMMEDIATE"
 
+# Database types ``_database_connection`` will bring up a server for. "auto" is
+# deliberately not listed: a run that never named a backend stays on the sqlite
+# file in its config directory. Adding "auto" here would make planemo prefer a
+# local postgres whenever psql happens to be on PATH, which is a policy call
+# rather than a bug fix - see the discussion on #1679.
```
mvdbeek: 1679 is this PR, i'm not seeing any discussion. However, presence of psql determining anything at all is a bug we should fix.

planemo/galaxy/config.py
```diff
@@ -1268,13 +1276,27 @@ def default_use_path_paste(self):
...
+@contextlib.contextmanager
 def _database_connection(database_location, **kwds):
-    if "database_type" in kwds and kwds["database_type"] == "postgres_singularity":
-        default_connection = postgres_singularity.DEFAULT_CONNECTION_STRING
+    """Yield the ``database_connection`` a managed Galaxy should run against.
+
+    An explicit connection string wins - it is what a profile stores once it has
+    resolved a database of its own. Otherwise only an explicitly named postgres
+    backend is worth standing a server up for; everything else gets the sqlite
+    file in the config directory.
```
mvdbeek: This repeats the comment in config.py. Is it even worth saying this ? When would we not honor an explicit setting ?

```suggestion
```
and maybe promote database_connection and `database_type` to the function signature if it doesn't break all sorts of mpy rules ?

tests/test_galaxy_config.py
```diff
@@ -105,6 +108,39 @@ def test_runnable_delegated_properties_are_booleans():
     assert Runnable("d", RunnableType.directory).is_single_artifact is False
 
 
+DATABASE_LOCATION = "/tmp/planemo-test-config/galaxy.sqlite"
```
mvdbeek: that probably fails on OSX ? mkdtemp to be on the safe side ?

mvdbeek (approved):

mvdbeek:
Posting on behalf of @mvdbeek: the last three commits address these issues:

1. **Docker database deletion:** Galaxy cleanup could stop the shared PostgreSQL container and delete its databases. It now leaves that container running.
2. **Broken detached serving:** `serve --daemon` with managed Singularity PostgreSQL left Galaxy running but stopped its database. That combination now fails early, with guidance to use foreground serving or an independently managed database.
3. **Premature readiness:** Singularity startup could mistake PostgreSQL's temporary initialization server for the final server. It now waits for initialization to finish before checking readiness.

The subsequent test cleanup was folded into those commits; it did not change their production behavior.

mvdbeek (approved):

jmchilton:
Great back and forth - love the team work on this one!

PR #1683, opened 2026-08-21
Document running Planemo workflows on Slurm clusters (rebase + fixes)

jmchilton:
Rebase of #1512 by @Smeds, with both original commits and authorship preserved, plus fixes from review.

The implementation prerequisites discovered during review were fixed in #1685, which has merged. This PR now contains only the Slurm guide in `docs/_running_slurm.rst` and its inclusion from `docs/running.rst`.

Review feedback from #1512 addressed:
- replaces hand-maintained Galaxy job YAML with Planemo `job_config_init` for Slurm and TPV
- removes the manual shared `TMPDIR` workaround and documents the default `tmp_dir: true` behavior

Review feedback from this PR addressed:
- uses Apptainer throughout the prose; the option and generated configuration keys remain spelled `singularity`, which the guide explains
- recommends a profile immediately after the first single-server run and generates the job configuration with `profile_job_config_init`
- drops `--galaxy_version 25.0` from both examples
- drops the SQLite-locking rationale, PostgreSQL run example, and the "Database is locked" and "Galaxy exits during tool installation" troubleshooting sections
- narrows the persistence claim to Tool Shed tools after tracing profile and bare-run paths

Two comments exposed code defects that were fixed and tested in #1685:

- Omitting `--galaxy_version` selected the legacy 24.2 layout because the Click default pinned 24.2 while `gxjobconfinit` defaults to the modern layout. For `--tpv`, the legacy output used a `dynamic` runner, required an external `tpv.yml`, and left `tpvdb_slurm` as a commented template.
- `profile_job_config_init` recorded the profile's own options JSON as `job_config_file`, so the generated Slurm configuration was never applied. It now records the generated `job_conf.yml`.

Additional cleanup:
- removes the conflicting biocontainers example from the Apptainer path
- fixes stale links, spelling, RST structure, and command examples
- removes the nonspecific Tool Shed startup troubleshooting and retains the actionable port-collision guidance

Validation:
- generated both documented job configurations and confirmed the runners, TPV database, `tpvdb_slurm` destination, `drmaa_library_path`, `native_specification`, and `tmp_dir` values described by the guide
- confirmed Planemo refuses to generate into a profile that already holds a job configuration, as the TPV section warns
- Sphinx renders the updated running guide without warnings from the modified documents
- CI passed on the previous documentation head
- `git diff --check` passes after rebasing onto current `master`

Supersedes #1512.

🤖 Description updated with Codex


docs/_running_slurm.rst
```diff
@@ -0,0 +1,204 @@
...
+
+.. _slurm-drmaa: https://github.com/natefoo/slurm-drmaa
+
+Tool dependencies
+~~~~~~~~~~~~~~~~~
+
+Galaxy can resolve tool dependencies with Conda or run tools in Docker or
+Singularity/Apptainer containers. Containers are generally more reproducible,
+but the available runtime and mount configuration depend on the cluster. The
+examples below use Singularity because it is commonly available on HPC systems.
```
mvdbeek: I would probably replace singularity throughout with apptainer.

docs/_running_slurm.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Database
+~~~~~~~~
+
+Planemo uses SQLite by default. SQLite is sufficient for small runs, but a
```
mvdbeek: This has been fixed, all of the workflow tests also pass on sqlite.

docs/_running_slurm.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Use :doc:`commands/job_config_init` instead of maintaining a complete Galaxy
+job configuration by hand:
+
+.. code-block:: console
+
+    $ planemo job_config_init \
+        --runner slurm \
+        --singularity \
+        --galaxy_version 25.0
```
mvdbeek: This is awkward, can we assume modern galaxy without the version flag, so that this ages better ?

docs/_running_slurm.rst
```diff
@@ -0,0 +1,204 @@
...
+
+.. code-block:: console
+
+    $ planemo run tutorial.ga tutorial-job.yml \
+        --download_outputs \
+        --output_directory . \
+        --output_json output.json
+
+Planemo launches a temporary Galaxy instance, runs the workflow, downloads its
+outputs, and records their paths in ``output.json``.
```
mvdbeek, line 77: After this i would strongly recommend setting up a profile

docs/_running_slurm.rst
```diff
@@ -0,0 +1,204 @@
...
+Troubleshooting
+---------------
+
+Database is locked
+~~~~~~~~~~~~~~~~~~
+
+Errors containing ``sqlite3.OperationalError: database is locked`` usually
+mean the workflow has more concurrent database activity than SQLite can
+comfortably handle. Re-run with PostgreSQL, using either an existing database
+or ``--database_type postgres_singularity``.
```
mvdbeek:
```suggestion
```

I don't know how i feel about this, it shouldn't happen anymore. Considering we bootstrap a single galaxy and interact with it i don't really see a need for postgres but that's just my preference for keeping it simple.

docs/_running_slurm.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Galaxy exits during tool installation
```
mvdbeek: IIRC @Smeds worked against a cluster with completely broken network. The instructions aren't specific, I would drop that too

mvdbeek (approved):

docs/_running_slurm.rst
```diff
@@ -82,30 +77,61 @@ Run the example workflow locally with:
 Planemo launches a temporary Galaxy instance, runs the workflow, downloads its
 outputs, and records their paths in ``output.json``.
 
+Reuse the instance with a profile
+---------------------------------
+
+Every bare ``planemo run`` bootstraps Galaxy from scratch: a fresh database, and
+a fresh installation of each tool the workflow needs. On a cluster that is slow,
```
mvdbeek: Do we actually keep tools by default ? ... i trust whatever your agent says 😆

jmchilton:
Blocking on https://github.com/galaxyproject/planemo/pull/1685 now - I think that is the right place to address some of these comments independent of this piece of documentation.

mvdbeek (approved):

jmchilton:
I could have the agent respond point by point - but I am taking your approve as it is okay that I don't have too 😅. Thanks for the review - I think it really hardened the the final product.

PR #1702, opened 2026-09-12
Add --use_cache to planemo test, off by default

jmchilton:
> _Posted by Claude (AI assistant) on behalf of @jmchilton — they did not author this text personally._

Follow-up to #1665 / #1666, which added `--use_cache` to `planemo run` with default on. The stated intent for `planemo test` was **default off**, but it landed as **no option at all** — so tests can never opt in.

This adds the flag to `planemo test`, defaulting off.

## Why it's nearly free

The plumbing already exists. `_execute` reads `kwds["use_cache"]` for both the tool `POST` (`activity.py:292`) and `invoke_workflow` (`activity.py:330`), and `cmd_test`'s kwds already reach it: `test_runnables(**kwds)` → `engine_context(ctx, **kwds)` → `build_engine` → `engine._kwds` → `execute(..., **self._kwds)`. Only the option registration was missing.

## Why default off

A run wants results; a test wants the work done again. Galaxy decides equivalence from tool id, tool version and inputs, so a tool edited **without** a version bump looks equivalent to its previous self — its old outputs get replayed and the test passes without ever exercising the edit. Opting in makes sense when the tools under a workflow are fixed and the iteration is on the workflow or its assertions.

## Scope limit (documented, not silent)

`--use_cache` is only honored for tests defined in a **separate test file** (workflow tests, CWL tests, tool tests with an accompanying `-tests.yml`). These go through `BaseEngine._run_test_cases` → `_run()` → `_execute`.

Tests embedded in a tool's `<tests>` block take the other branch of `GalaxyEngine._run_test_cases` — `expand_test_cases` → `GalaxyInteractorApi` / `verify_tool` — which never touches `_execute` and ignores the flag. That's called out in the option help, the command docstring and the docs rather than left to surprise someone.

Also adds `--cwltool_cache_directory` to `test`, so the cwltool engine's cache is overridable there as it already is for `run`.

## Testing

- `tests/test_cmd_test.py::CmdTestUseCacheTestCase` — three new tests covering default-off, `--use_cache`, and `--no_use_cache` reaching the engine layer. Verified red-to-green: all three fail with the option decorators removed.
- `tests/test_galaxy_activity.py` — existing tests already cover `use_cache` → `use_cached_job` on both request paths. One docstring corrected; it asserted that `planemo test` never sets `use_cache`, which this PR changes.
- Ran `tests/test_cmd_test.py tests/test_galaxy_activity.py tests/test_run.py` under the `quick` env (8 passed, 36 skipped), plus flake8 / black / isort clean on the changed files.

**Not covered:** a Galaxy-backed integration test proving a real cache hit through `planemo test`, mirroring `test_run_gxtool_use_cache`. Worth adding, but it needs a persistent-profile fixture and a no-data-input tool with a separate tests file; happy to follow up.

## Note on `docs/commands/test.rst`

Regenerated with `scripts/commands_to_rst.py`. It picks up pre-existing drift from earlier PRs that never regenerated docs (`--shed_data_dir`, `--tool_evaluation_strategy`, `--failed/--lf`). I reverted that regeneration for the 17 other command pages it also touched, to keep this diff scoped — those remain stale on master.

## Caveat worth knowing

The upload-equivalence limitation already documented for `run` applies here too: inputs given as local file paths are uploaded fresh each run and Galaxy does not consider jobs consuming distinct uploads equivalent. So for workflow tests with local test data, cache hits will be rare until the data is already on the server. The flag is still the right primitive; just don't expect it to speed up a cold local loop.

Requested by galaxyproject/foundry#365 (item 1).


docs/_running_external.rst
```diff
@@ -418,3 +418,18 @@ that succeeds:
...
   most when the job file refers to datasets already on the server.
+
+``planemo test`` accepts the same ``--use_cache`` flag but defaults it **off**,
+because the two commands want different things. A run wants results; a test
+wants the work done again. Galaxy decides equivalence from the tool id, tool
+version and inputs, so a tool edited without a version bump looks equivalent to
+its previous self and its old outputs are replayed - the test then passes
+without ever exercising the edit. Opt in when the tools under a workflow are
+fixed and the iteration is on the workflow or its assertions, and leave it off
+when the tool wrappers themselves are what changed.
```
jmchilton: ^ This paragraph is too wordy. Please reduce it significantly and merge it into the next one.

planemo/galaxy/activity.py
```diff
@@ -289,7 +289,8 @@ def _execute(  # noqa C901
             tool_id=tool_id,
             inputs=job_dict,
             inputs_representation=inputs_representation,
-            # ``planemo test`` never sets use_cache, so tests never silently reuse results
+            # Off unless the caller opts in - ``planemo run`` defaults it on,
```
jmchilton: This whole comment block should be deleted.

docs/_running_external.rst
jmchilton: > _Posted by Claude (AI assistant) on behalf of @jmchilton — they did not author this text personally._

Done in f0bf9fe4 — cut from 15 lines across two paragraphs to 5 in one, merged with the scope-limit paragraph that followed. Kept both facts (default-off rationale, test-file-only scope); the longer version still lives in the option help and command docstring.

planemo/galaxy/activity.py
jmchilton: > _Posted by Claude (AI assistant) on behalf of @jmchilton — they did not author this text personally._

Deleted in f0bf9fe4.

mvdbeek (approved):
😆 is this the 2026 version of John writing an amazing feature without UI ?

PR #1167, opened 2021-05-04
fix and complete metadata setting docs

bernt-matthias:
https://github.com/galaxyproject/galaxy/pull/4437

PR #1716, opened 2026-09-16
Document both galaxy.json tool-provided metadata formats

jmchilton:
> 🤖 PR opened by Claude (AI assistant) on behalf of @jmchilton.

**[Draft]** — opening early for visibility; not ready for merge.

## What

Rewrites `docs/_writing_tool_metadata.rst`. The page called itself "This stub of a section", documented only the pre-17.09 `galaxy.json` format, and misdescribed several of its fields.

Supersedes #1167 by Matthias Bernt, which found the first error below and sketched the missing section. This carries that correction forward and fills in what the PR left as `TODO`.

## Why

`galaxy.json` has two formats, chosen by tool profile (`tool_util/parser/xml.py`): `< 17.09` is JSON Lines, `>= 17.09` is a single JSON object keyed by output name. `<outputs provided_metadata_style=...>` overrides the choice, `provided_metadata_file=` renames the file. None of that was documented.

More importantly, **from profile 26.2 Galaxy does not read the file at all unless the tool opts in** — `_uses_tool_provided_metadata` (`tool_util/abstract_tool.py`) requires an explicit `provided_metadata_file`/`provided_metadata_style`, an output with `format="auto"`, or a `discover_via="tool_provided_metadata"` collector. Without one, metadata is silently ignored. The page never mentioned this, and Galaxy's own example tools don't show it because they are all `profile="17.09"`.

## Corrections

1. **`dataset` vs `dataset_id`.** The page said an entry called `dataset` "should be the database id". Two distinct keys exist: `dataset_id` is the database id; `dataset` is the output file's basename, which Galaxy resolves to that output's id. #1167 spotted this; both are valid, so the page documents each.
2. **`metadata` on declared outputs.** The page said such entries accept every field "except `metadata`". They do accept it — `set_metadata.py` routes declared outputs through `set_meta_with_tool_provided()`, which applies `file_dict["metadata"]`.
3. **`"failed": true` fails the job, not the output.** `has_failed_outputs()` is file-wide, and both consumers set `final_job_state = ERROR`. The draft of this PR got this wrong too.
4. **`extra_files` is discovered-datasets only.** It is read in `write_job_metadata()`, which iterates `get_new_datasets_for_metadata_collection()`. It is not honoured for declared outputs, where the accepted keys are `TOOL_PROVIDED_JOB_METADATA_KEYS` plus `ext` and `metadata` — so `created_from_basename` was missing from the legacy list.
5. **`auto_format="true"` is not an alternative to `format="auto"`.** It sets `_sniff_`, so it does not satisfy the 26.2 opt-in, and it cannot be combined with `format`.

## Added

- The profile 26.2 opt-in gate and its triggers, linked to galaxyproject/galaxy#23404.
- Current format: simple outputs, discovered datasets, and collection elements via nested `elements`.
- That `designation` in a JSON entry applies under `from_provided_metadata="true"` discovery, not under `pattern=` discovery where the regex group wins.
- That supplying `metadata` for a *discovered* dataset replaces automatic datatype detection, unlike the declared-output path.
- A `<configfiles>` snippet showing how a tool actually emits the file, and the per-field glossary the old stub had.
- Example-tool links repointed from `jmchilton/galaxy@2909e746` to `galaxyproject/galaxy` pinned at `c6c3b6df49`.

## Verification

Every documented field was read out of galaxy `dev` @ `c6c3b6df49` and matched against the `tool_provided_metadata_*.xml` and `collection_creates_dynamic_nested_from_json*.xml` test tools.

Sphinx builds clean — the section nests correctly under `writing_advanced` (h2/h3/h4), the `Simple Outputs`_ cross-reference resolves, and no warnings are added.

## Attribution

Prose drafted by Codex (gpt-5.6). Research and verification by Claude, across three review rounds — corrections 3, 4 and 5 and the 26.2 gate came out of the second. The original corrections are Matthias Bernt's, from #1167.


docs/_writing_tool_metadata.rst
```diff
@@ -1,52 +1,295 @@
...
+          {
+            "identifier_0": "oe1",
+            "identifier_1": "ie2",
+            "filename": "oe1_ie2.fq"
+          }
+        ]
+      }
+    }
+
+Galaxy converts nested ``elements`` to the flat representation internally,
```
jmchilton: Can we pick a preferred version of this - is the code or tools-iuc clear about this. I'm not sure we should be documenting the fallback.

docs/_writing_tool_metadata.rst
```diff
@@ -1,52 +1,295 @@
...
 
-Examples of tools using ``new_primary_dataset`` entries:
+    <command>cp '$c1' galaxy.json</command>
+    <configfiles>
+        <configfile name="c1">{"out1": {"name": "Result"}}</configfile>
+    </configfiles>
 
-- `tool_provided_metadata_2.xml <https://github.com/jmchilton/galaxy/blob/2909e74642180bd818019ebdcb62e62f12e56e69/test/functional/tools/tool_provided_metadata_2.xml>`__ demonstrating using the simpler attributes described here.
-- `tool_provided_metadata_3.xml <https://github.com/jmchilton/galaxy/blob/2909e74642180bd818019ebdcb62e62f12e56e69/test/functional/tools/tool_provided_metadata_3.xml>`__ demonstrates overridding datatype specified metadata.
+Starting with profile 26.2, Galaxy loads this file only when the tool opts in.
```
jmchilton: Learned something today - can you link to a particular PR really concisely here.

docs/_writing_tool_metadata.rst
jmchilton: *Posted by Claude (AI assistant) on jmchilton's behalf, not authored by them personally.*

Picked nested `elements` and dropped the flat form. Two lines of evidence:

- **tools-iuc** has exactly two tools using provided-metadata collection discovery — `tools/packaged_annotation_loader/load_annotations.xml` and `tool_collections/bamtools/bamtools_split/bamtools_split_ref.xml`. Both write `elements`; neither writes `identifier_<level>`.
- **The code** is clear that flat is not the authoring interface: `ToolProvidedMetadata._elements_to_datasets` converts nested `elements` into `{"identifier_0": ..., "identifier_1": ...}`. That shape is what Galaxy builds internally.

So this section now documents `elements` only. Removed the flat JSON example, the "Galaxy converts nested elements internally" sentence, the "do not mix `datasets` and `elements`" caveat (it only existed to disambiguate two documented forms), and the link to `collection_creates_dynamic_nested_from_json.xml`.

The `datasets` list in **Discovered Datasets** is a different mechanism and is unchanged.

One caveat: the flat form still works and is still exercised by a Galaxy test tool, so someone debugging an *existing* tool that uses it will no longer find it documented. Happy to add a one-sentence pointer if that trade seems wrong.

docs/_writing_tool_metadata.rst
jmchilton: *Posted by Claude (AI assistant) on jmchilton's behalf, not authored by them personally.*

Linked, as a parenthetical on that sentence: galaxyproject/galaxy#23404 (`secure-dynamic-output-collection`), merged from commit `64360a6` "Harden tool-provided dynamic output collection".

One thing worth flagging: that commit message says "profiles through 26.2 ... retain their historical behavior", but `AbstractTool._uses_tool_provided_metadata` tests `Version(str(self.profile)) >= Version("26.2")` — so 26.2 itself already has to opt in. The docs follow the code, not the commit message.

jmchilton:
I took a couple passes at pushing back against slop - it feels like a good step forward now.

mvdbeek (approved):

